PR #67, opened 2026-04-20
Fix Min/Max Temperature Aggregation for Forecast in Weather Widget

bfbones:
## Summary

The min/max temperature values on dashboard were taken from the 12:00 entry from OWM. In that case and endpoint, this provides only a snapshot of 12:00 so min/max values are same most of the time for that snapshot.
To get min/max values correctly from this free API endpoint we have to aggregate one day and get the min/max out of the series of values per day.
To get a matching icon we could still use the icon for the 12:00 entry (or maybe use some AI in future to forecast a day by the series of weather information of one day.. :D)

## Changes

- changed the block which fetches and aggregates forecast data from OWM API in server/routes/weather.js

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.20.33] - 2026-04-20
+
+### Fixed
+- Min/Max Temperature Aggregation for Forecast in Weather Widget
+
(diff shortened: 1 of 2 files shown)


ulsklyc:
Thanks for the contribution @bfbones! The fix is correct — the OWM free-tier `/forecast` endpoint's per-entry `temp_min`/`temp_max` fields only reflect the 3-hour window, not the full day. Aggregating across all daily entries is exactly the right approach.

Released in **v0.20.33** 🎉

PR #91, opened 2026-04-27
Adds a more family-oriented member management flow and profile customization.

rafaelfoster:
## Summary

  Adds a more family-oriented member management flow and profile customization.

  ## Changes

  - Separated family role from system access role.
  - Added family roles such as Dad, Mom, Parent, Child, Grandparent, Relative, and Family member.
  - Replaced Admin/Member selection with a System Admin checkbox.
  - Added editing for existing members, including name, username, family role, system admin access, color, and profile picture.
  - Added profile picture upload/update for the current user.
  - Added read-only `GET /api/v1/family/members` endpoint for listing family members.
  - Updated dashboard/family member displays to support profile pictures.
  - Added OpenAPI documentation and translations for supported locales.
  - Fixed profile picture preview to work with the existing CSP.
  - Fixed member edit save flow so it no longer prompts to discard changes after a successful save.

  ## Validation

  - Syntax checks passed.
  - Locale JSON validation passed.
  - Dashboard and task tests passed.
  - Docker build and health check passed.


ulsklyc:
Thank you, @rafaelfoster! 🎉

This is a great addition to Oikos. Family roles, profile pictures, and the member-edit flow all feel right at home in a family planner — and the full i18n coverage across all 16 locales is much appreciated.

A few things I cleaned up before merging (none of them were blockers):

- `server/routes/family.js` — verified that `requireAuth` is applied globally on `/api/v1` in `server/index.js`, so the endpoint was already protected; no change needed.
- All 26 tests passed without modification.

The changes have been released as **[v0.27.0](https://github.com/ulsklyc/oikos/releases/tag/v0.27.0)**.

Thanks again for the contribution!

PR #110, opened 2026-04-30
Consistent upload option in the calendar, and fiz images rendering 

rafaelfoster:


ulsklyc:
Thanks @rafaelfoster for this contribution! 🎉

This is a really solid improvement — the attachment rendering fix addresses a real bug (raw base64 stored without the `data:` prefix), the drag-and-drop dropzone is a nice UX upgrade consistent with the Documents module, and moving the theme-init script to an external file tightens up the CSP.

I made two small adjustments before merging:
- **Error message fix**: The "file too large" validation error was swallowed by the `saveEvent` catch block (which expected an `ApiError` shape); now falls back to `err.message` so the correct toast is shown.
- **CSS token consistency**: Changed hardcoded `--module-calendar` to `--module-accent` in the new dropzone styles, aligning with the design system convention used in `documents.css`.

Released as **[v0.37.0](https://github.com/ulsklyc/oikos/releases/tag/v0.37.0)**. Thanks again!

PR #140, opened 2026-05-09
Update pl.json

czacheur:
corrected double quotation mark at the end of the line to typo quotation and quotation mark.


ulsklyc:
Thanks for the fix! The correct closing quotation mark makes the Polish UI look properly typeset.

PR #141, opened 2026-05-11
Feature split expenses

rafaelfoster:
Summary

  Adds a native Oikos Split Expense module inside Budget, replacing the need for Splitwise-style external expense sharing. The feature integrates with existing auth, SQLite migrations, CSRF/ session security, locale system, contacts, birthdays, family users, and the vanilla SPA UI.

  Key Changes

  - Added Budget > Split tab for shared expenses.
  - Added expense groups with create, edit, archive, delete-empty-group, and member management.
  - Added shared expenses with payer, participants, equal/exact/percentage/shares split methods, validation, and ledger-backed balances.
  - Added settlements, activity feed, recurring expense support, and debt simplification.
  - Added restricted Split guest accounts:
      - guests can log in only to the Split module
      - guests only see their invited group
      - non-Split APIs are blocked server-side
  - Added guest/contact integration:
      - create guests from Split
      - add family members or contact-only people to groups
      - contact-only people are converted to restricted Split guests
      - guest contacts and birthdays sync into existing Oikos data
  - Added receipt/proof architecture via existing family documents linkage.
  - Added multi-currency fields and historical exchange snapshot support.
  - Added localized Split module strings across supported locales.
  - Added docs and focused split expense tests.

  Financial Safety

  - Uses integer minor currency units and Decimal-string parsing.
  - Avoids floats for money.
  - Balances are derived from immutable ledger entries.
  - Expense split rows now always store currency, fixing the expense_splits.currency NOT NULL failure.
  - Debt simplification is deterministic.

  UX Improvements

  - Budget FAB opens a Split expense when Split tab is active.
  - Budget prev/next controls are hidden on Split and Loans.
  - Split expense modal defaults Paid by to the logged-in user.
  - Removed unused Category field and quick actions.
  - Save button is disabled until split math validates.
  - Recent expenses show Paid by: {name}.
  - Create Guest birth date follows Settings > General date format and auto-inserts separators.


ulsklyc:
### Code review

Found 8 issues:

1. **Migration renumbering breaks append-only constraint** (CLAUDE.md: "Migrations append-only. Never modify or reorder existing entries.") — The PR inserts v30 (birthday reminders) before the existing v30 (CardDAV), bumping it to v31. Any database that already ran the original v30 (CardDAV) will silently skip the newly inserted v30 because the migration runner has already recorded `version: 30`. This silently leaves the birthday reminder columns absent.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/db.js#L1078-L1086

2. **`birthdayReminderAt` silently drops `offsetMin` parameter** — The function body was simplified to `return \`${next}T12:00:00Z\``, discarding the offset calculation. All custom reminder offsets are silently ignored after this change.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/services/birthdays.js#L59-L61

3. **`bcrypt.hashSync` blocks the Node.js event loop** — `userFromContact()` calls `bcrypt.hashSync(…, 12)` synchronously. At cost 12 this takes ~200 ms and stalls all concurrent requests. Use `await bcrypt.hash()` with the hash pre-computed before the transaction (as done elsewhere in the same file).

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/routes/split-expenses.js#L4211

4. **UTC date bug in expense date input default** (CLAUDE.md: "never `new Date().toISOString().slice(0,10)`, which gives UTC and can be one day off") — Default date for the expense form uses `.toISOString().slice(0,10)`. Replace with local-time arithmetic (`d.getFullYear()`, `d.getMonth()+1`, `d.getDate()`).

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/routes/split-expenses.js#L4344

5. **`GET /split-expenses/meta` route missing `try/catch`** (CLAUDE.md: "Every route handler in try/catch.") — This is the only route in the file without a try/catch; `defaultCurrency()` executes a DB query that can throw.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/routes/split-expenses.js#L4354-L4356

6. **`--module-accent` not set on `.split-page`** (CLAUDE.md: "Each module CSS file sets `--module-accent` on its page root class.") — The FAB and active nav item won't reflect the split-expenses color.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/public/styles/split-expenses.css#L1-L10

7. **Guest added to a second group is silently blocked** — `split_expense_guest_users.user_id` is a PRIMARY KEY, so `INSERT OR IGNORE` on a second group silently does nothing. The guest becomes a member of the new group in `expense_group_members` but `requireGroupAccess` still reads the original group ID, blocking all access with no error returned to the caller.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/routes/split-expenses.js#L4515

8. **`server/openapi.js` not updated** (CLAUDE.md: "API schema, security model, all route definitions" → `server/openapi.js`) — 26 new routes were added but the OpenAPI spec was not updated.

https://github.com/ulsklyc/oikos/blob/316e7a594f316ef1800403b48b82128315bb49d9/server/openapi.js

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ulsklyc:
Thank you for this contribution! The Split Expenses module is a great fit for Oikos — well-architected with the immutable ledger pattern, integer arithmetic for financial safety, and full multi-currency support.

Before merging, I applied the following corrections:

- **Migration renumbering fixed**: the PR inserted a v30 "Advanced reminder options" entry before the existing CardDAV v30, which violated the append-only constraint. The erroneous entry was removed and migration order restored to match `main` (CardDAV = v30, birthday reminders = v31, new split-expenses = v39/v40).
- **`birthdayReminderAt` offset restored**: the `offsetMin` calculation had been dropped, silently discarding user-configured reminder offsets. Restored the original offset-aware implementation.
- **`bcrypt.hashSync` replaced**: `userFromContact()` was calling `bcrypt.hashSync()` synchronously inside a DB transaction, blocking the event loop for ~200 ms per call. Replaced with `await bcrypt.hash()` and made the route handler async.
- **UTC date bug fixed**: replaced `new Date().toISOString().slice(0,10)` in the expense date input with local-time arithmetic per project convention.
- **`GET /split-expenses/meta` try/catch added**: the only route handler in the file without error handling.
- **`--module-accent` added to `.split-page`**: required for FAB and active nav item to reflect the split-expenses color.
- **Guest multi-group bug fixed**: `INSERT OR IGNORE` silently allowed a guest to be added to a second group while their `splitGuestGroupId` still pointed to the original group, blocking all access. Returns 409 instead.
- **OpenAPI spec updated**: 26 new routes and the `SplitExpenses` tag added to `server/openapi.js`.
- **Polish translations added**: `pl.json` was added to `main` after this PR was opened and was missing split-expenses keys.

The changes are live in [v0.51.0](https://github.com/ulsklyc/oikos/releases/tag/v0.51.0).

PR #204, opened 2026-06-03
refactor: compress text assets with gzip and defer the Lucide script

jesposito:
## What

Two small load-performance fixes for cold page loads:

1. **Add `compression()` (gzip) middleware.** The frontend source is served unminified and was going out **uncompressed** — roughly 1.6 MB of highly compressible text (`router.js` 76.5 KB, `calendar.js` 83 KB, `settings.js` 133 KB, `lucide.min.js` 358 KB, `dashboard.css` 70 KB, …), plus JSON API responses. With gzip, `router.js` drops **76.5 KB → 19.5 KB (~75%)**; CSS/JSON see similar wins.
2. **`defer` the Lucide `<script>`.** It's a 358 KB library loaded synchronously in `<head>`, fully render-blocking on every page. `defer` makes it non-blocking. It still executes **before** the deferred `api.js`/`router.js` ES modules (they appear later in document order and modules are deferred too), so the global `lucide` is defined before any `lucide.createIcons()` call at render time.

## Why a dependency

`compression` is a **backend-only** dependency (the standard Express middleware). My reading is that the CONTRIBUTING "no external dependencies" rule targets the **frontend** (it carves out Lucide specifically). Happy to swap for a small built-in `zlib` middleware instead if you'd prefer zero new deps — just say the word.

## Testing

- Smoke: started the server, confirmed `Content-Encoding: gzip` + `Vary: Accept-Encoding` on `/router.js` (76574 → 19479 bytes), HTML still 200, server boots clean.
- `npm run test:api` (5/5) and `npm run test:dashboard` (11/11) pass.

## Notes

For deployments behind a reverse proxy that already gzips, this is harmless (the proxy won't double-compress an already-encoded response). It guarantees compression for users who run Oikos without a compressing proxy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Compress CSS/JS/JSON responses with gzip and defer the icon library so it no longer blocks first paint.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "bcrypt": "^6.0.0",
         "better-sqlite3": "^12.10.0",
+        "compression": "^1.8.1",
         "dotenv": "^17.4.2",
         "express": "^5.2.1",
         "express-rate-limit": "^8.5.2",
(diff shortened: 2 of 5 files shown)


ulsklyc:
Thanks! The gzip compression middleware and deferred Lucide script are merged into `main`. Minor cleanup applied: removed the multi-line explanation comment (the code is self-evident) and dropped the redundant comment above the `defer` attribute in the HTML.

PR #205, opened 2026-06-03
fix(weather): honor app language and city IDs, key cache per query

jesposito:
## Problems

Running Oikos with the UI in English, the weather widget still showed German descriptions (e.g. *"Bedeckt"*). Digging in surfaced three related issues in `server/routes/weather.js`:

1. **Language was hardcoded to German.** `const lang = process.env.OPENWEATHER_LANG || 'de'` — the OpenWeatherMap `lang` never followed the app locale, so non-German users got German descriptions unless they knew to set an env var.
2. **The in-memory cache was keyed only by timestamp** (`{ data, ts }`), ignoring city/units/lang. If those ever vary (they now can, see below), the 30-minute cache serves the wrong city/language/units to everyone.
3. **Numeric `OPENWEATHER_CITY` values 404.** The URL always used `q=`, but OWM treats a numeric value as a name and returns `404 city not found`. Users who pasted a **city ID** to disambiguate (e.g. Wellington NZ vs Wellington UK) ended up with an empty widget.

## Changes

- **Locale-aware language.** The dashboard now sends `?lang=<active locale>` (via the existing `getLocale()`), and the route resolves `lang` as `req.query.lang || OPENWEATHER_LANG || 'en'`. Same pattern allows optional `?city=`/`?units=` overrides. (The ultimate fallback is `'en'` to match the app's own English i18n fallback — easy to change back to `'de'` if you'd prefer.)
- **Per-parameter cache.** Cache is now a `Map` keyed by `city|units|lang`, bounded to 50 entries (oldest evicted).
- **City IDs supported.** A numeric `OPENWEATHER_CITY` is sent as `id=`; otherwise `q=`. Names can still be disambiguated with a country code, e.g. `Wellington,NZ`.

## Testing

- Boot smoke: server starts clean; `GET /api/v1/weather?lang=en` returns 401 (auth-gated) rather than 500 — route logic intact.
- `npm run test:dashboard` (11/11) and `npm run test:frontend-audit` (57/57) pass.
- Verified live against OpenWeatherMap: `q=2179537` → `404 city not found`; `q=Wellington,NZ` → Wellington, **NZ**; `?lang=en` returns English descriptions.

## Note on locale codes

`getLocale()` returns the app's locale code (e.g. `en`, `de`, `fr`). Most line up with OWM's `lang` codes; a few differ (OWM uses `pt_br`, `zh_cn`, etc.) and OWM falls back to English for unknown values. A small locale→OWM map could be added later if you want exact coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Make the weather widget follow the app language, support city IDs, and key its cache per city/units/language.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/public/pages/dashboard.js
+++ b/public/pages/dashboard.js
@@ -1274,7 +1274,7 @@ export async function render(container, { user }) {
   try {
     const [dashRes, weatherRes, prefsRes] = await Promise.all([
       api.get('/dashboard'),
-      api.get('/weather').catch(() => ({ data: null })),
+      api.get(`/weather?lang=${encodeURIComponent(getLocale())}`).catch(() => ({ data: null })),
       api.get('/preferences').catch(() => ({ data: {} })),
     ]);
     data         = dashRes;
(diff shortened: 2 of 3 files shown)


ulsklyc:
Thanks for this fix! The changes have been rebased onto `main` (resolved a CHANGELOG conflict with recent releases) and merged manually. The weather widget now follows the app language, supports numeric city IDs, and has a per-city/units/language cache.

Closing this PR as the changes are live on `main`.

PR #206, opened 2026-06-03
refactor(ui): debounce document and split-expense search inputs

jesposito:
## What

Two search inputs ran their full handler on **every keystroke**:

- **Documents** (`documents.js`): each character set `state.query` and called `renderDocuments()`, rebuilding the entire document list synchronously.
- **Split-expenses** (`split-expenses.js`): each character fired **two sequential API calls** (`loadGroups()` then `loadGroupData()`) plus a full `renderAll()`. Typing "groceries" = ~9 keystrokes × multiple requests.

## Fix

Wrap both in the same `setTimeout`/`clearTimeout` debounce that the **Contacts** search already uses (`contacts.js:107`) — 200 ms for documents, 250 ms for split-expenses (it hits the network). The input value is captured at event time so the deferred handler reads the right string. Behaviour is identical, just coalesced.

## Testing

- `node --check` on both files.
- `npm run test:frontend-audit` and `npm run test:split-expenses` pass.

Small, self-contained; part of a series of render-perf fixes. The larger one (scoping the ~51 unscoped `lucide.createIcons()` calls) will be a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Debounce the documents and split-expenses search inputs.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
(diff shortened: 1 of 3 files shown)


ulsklyc:
Thanks! The debounce changes have been rebased onto `main` and merged. Documents search now waits 200 ms and split-expenses group search 250 ms before querying — noticeably smoother on slower connections.

PR #210, opened 2026-06-03
fix(ui): remove backdrop-filter that made modal scrolling janky

jesposito:
## Problem

The event editor modal scrolls with severe lag and visible hangs (reported on a real install; worst on the tallest modal because it's the one that actually scrolls).

## Root cause

`.modal-panel` has `backdrop-filter: var(--blur-sm)` (glass.css `@supports` block) **and** contains the `.modal-panel__body` scroll container (`overflow-y: auto`). A `backdrop-filter` on an ancestor of a scrolling element forces the browser off the fast async-scroll/composited path and re-rasterizes the blurred region every frame — the classic "laggy/hanging modal scroll," brutal on mobile GPUs.

This codebase already documents the same class of bug for sticky toolbars (see the "backdrop-filter on position:sticky inside overflow:auto" comments in `housekeeping.css` and `shopping.css`) and fixes it the same way: go opaque.

## Fix

Remove the `backdrop-filter` override on `.modal-panel`. It falls back to the opaque `background-color: var(--color-surface)` already defined in `layout.css`, so no new tokens. The page-dimming blur stays on the **non-scrolling** `.modal-overlay`, so the glass aesthetic is preserved — the panel just sits opaque on top of the still-blurred backdrop. One rule removed.

## Notes
- Visual: the modal panel itself no longer blurs what's directly behind it; in practice it sits on the blurred dark overlay so the difference is barely perceptible, and scroll becomes smooth.
- No JS change; `test:frontend-audit` and `test:calendar` still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fix janky, laggy scrolling in tall modals (e.g. the event editor) caused by a backdrop-filter on the scroll container.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/public/styles/glass.css
+++ b/public/styles/glass.css
@@ -95,11 +95,15 @@
     -webkit-backdrop-filter: var(--blur-xs) saturate(120%);
   }
 
-  .modal-panel {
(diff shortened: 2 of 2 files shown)


ulsklyc:
Great catch! The `backdrop-filter` removal (with a short comment explaining why it must stay off) has been rebased and merged into `main`. Scroll jank in tall modals like the event editor is fixed.

PR #211, opened 2026-06-03
refactor(ui): scope icon rendering and index calendar day lookups

jesposito:
Two frontend render-performance fixes.

## Fix 1 — scope `lucide.createIcons()` calls
Unscoped `lucide.createIcons()` (no args) re-scans the whole document and rebuilds every `[data-lucide]` SVG. Called after rendering a list or per keystroke (autocomplete, ingredient rows, icon picker), it reprocessed all page icons and caused jank.

All ~38 remaining unscoped sites across 15 files now pass `{ el: <container> }` so only the just-rendered subtree is processed — matching the scoped pattern already used elsewhere in the codebase. Each call is scoped to the nearest element confirmed to wrap the rendered icons (rebuilt list/panel/grid/modal/page root). The shell-wide `router.js` `updateNav()` is scoped to `#app` (the app-shell root that holds both nav regions).

## Fix 2 — calendar day-bucketing (`public/pages/calendar.js`)
The month grid renders 42 cells; each called `eventsOnDay()`/`tasksOnDay()`, which each `.filter()` the entire `state.events`/`state.tasks` and built a `new Date` per event per cell — O(cells x events) with repeated date parsing every render.

Now `buildDayIndex()` (called once at the top of `renderView`) builds a `Map<isoDate, [...]>` for events and one for tasks, parsing each item's date once and bucketing multi-day events across their range (clamped to the loaded `[rangeFrom, rangeTo]` window). Per-cell lookups read O(1) from the Map. The helpers fall back to the original filter when no index is active, so behavior is preserved exactly (item ordering, all-day vs timed, and week/day/agenda views which share the same helpers).

## Tests
All pass, none modified:
- `test:calendar` — 29 passed
- `test:frontend-audit` — 57 passed
- `test:dashboard` — 11 passed
- `test:tasks` — 18 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Scope icon re-rendering to changed subtrees and index calendar day lookups for smoother large lists and month views.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/public/components/modal.js
+++ b/public/components/modal.js
@@ -290,7 +290,7 @@ export function openModal({ title, content, onSave, onDelete, onClose, size = 'm
   activeOverlay._onCloseCallback = onClose;
 
   // Lucide-Icons rendern
-  if (window.lucide) window.lucide.createIcons();
+  if (window.lucide) window.lucide.createIcons({ el: activeOverlay });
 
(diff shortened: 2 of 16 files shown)


ulsklyc:
Merged! Both parts — the scoped `createIcons({ el })` calls across all pages and the O(events+cells) calendar day-bucket index — are now on `main`. The calendar month view in particular should render noticeably faster on large datasets.

PR #212, opened 2026-06-03
refactor(db): batch split-expense queries, sort dashboard tasks in SQL

jesposito:
## What

Two safe, well-tested query optimizations on hot read paths.

### 1. Kill the split-expenses N+1
`serializeExpense()` ran **2 queries per expense** (splits + attachments). Mapped over `GET /groups/:id/expenses` (limit up to 100) and `GET /dashboard`, that was up to **~201 queries** per request.

New `serializeExpenseList()` batch-loads splits and attachments for all N expenses in **2 queries total** using `WHERE expense_id IN (...)` (placeholder count built per request), groups in JS, and attaches per row. The list endpoints (group expenses, dashboard recent, search) now run **3 queries instead of 2N+1**.

- Single-expense paths (`POST`/`PUT` returning one expense) keep the per-row path unchanged.
- Output shape is **byte-identical** to before (same fields, same key order).

### 2. Dashboard top-tasks sort in SQL
`urgentTasks` previously `SELECT`ed **all** non-done tasks and sorted in JS to take 5. Pushed ordering into SQL with `LIMIT 5`: overdue-first, then a `COALESCE`'d due datetime sort key (`NULLS LAST`), then priority rank. This is a faithful translation of the previous JS comparator.

## Why

Removes O(N) query fan-out on the most-hit list endpoints and stops loading the full task table into JS just to slice 5.

## Tests

- `npm run test:split-expenses` ✅
- `npm run test:dashboard` ✅
- `npm run test:db` ✅
- `node --check` on both changed files ✅
- Verified the new SQL top-5 ordering exactly matches the old JS comparator output for representative data (overdue, future-dated, undated, mixed priorities).

No test files were modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Batch split-expense detail loads to remove an N+1 query and sort dashboard tasks in SQL.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/server/routes/dashboard.js
+++ b/server/routes/dashboard.js
@@ -60,39 +60,33 @@ router.get('/', (req, res) => {
     result.upcomingEvents = [];
   }
 
-  // Offene Aufgaben: in JS sortiert damit due_time korrekt gegen lokale Zeit geprüft wird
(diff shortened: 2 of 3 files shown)


ulsklyc:
The N+1 batch fix and SQL-based task sorting are merged into `main`. Split-expense list endpoints now do 2 queries regardless of how many rows are returned, and dashboard tasks are sorted directly in SQLite.

PR #209, opened 2026-06-03
refactor(db): add indexes for assignment and recurring-event lookups

jesposito:
## What

Adds **migration 43** with four indexes on hot query paths. On a SQLCipher-encrypted DB every page read costs decryption, so unindexed scans hurt more than on plain SQLite.

| Index | Why |
|-------|-----|
| `idx_event_assignments_user` | PK is `(event_id, user_id)`; filtering events **by assignee** had no `user_id`-leading index. |
| `idx_task_assignments_user` | Same for `task_assignments` (PK `(task_id, user_id)`). |
| `idx_budget_loan_payments_entry` | Budget month list `LEFT JOIN budget_loan_payments ON budget_entry_id`, which was unindexed (only `loan_id`/`paid_date` were) -> a scan per budget row. |
| `idx_calendar_recurring` | Partial index `(start_datetime) WHERE recurrence_rule IS NOT NULL` for the calendar GET that selects all recurring events to expand. |

I deliberately **did not** add a `family_document_access(document_id)` index — its PK `(document_id, user_id)` already leads with `document_id`.

## Testing
- Fresh DB reaches **schema v43**, `Migration 43 applied` logged; all four indexes confirmed in `sqlite_master`.
- `test:db` (36), `test:dashboard` (11), `test:calendar` (29), `test:split-expenses` pass.

Additive, idempotent (`IF NOT EXISTS`), follows the "append new migrations at the end" rule.

PR #213, opened 2026-06-03
refactor(search): back global search with an FTS5 index

jesposito:
## What

Global search previously ran `LIKE '%q%'` — a **leading-wildcard, unindexable** pattern — across 5 tables (tasks, calendar_events, notes, contacts, shopping_items) on every request, i.e. **5 full table scans** per keystroke.

This replaces it with SQLite **FTS5**.

### Migration 43 (new, highest existing was 42)
- Creates one FTS5 virtual table `search_index(entity UNINDEXED, entity_id UNINDEXED, title, body, tokenize='unicode61')` indexing the searchable text of all five entities.
- `entity` + `entity_id` are stored UNINDEXED so each match joins back to its source row for the exact output fields and owner filtering.
- `AFTER INSERT/UPDATE/DELETE` triggers on all 5 source tables keep the index in sync (15 triggers).
- Backfills the index from existing rows in the same migration.

### Search rewrite
- Pure logic extracted to `server/services/search.js` (takes the db handle, no `db.js` side effects → directly testable under `node:sqlite`).
- Route runs FTS5 `MATCH` with safe query escaping: each token is sanitized, wrapped as a quoted prefix phrase (`"token"*`), and AND-joined. Empty/garbage queries short-circuit to empty results.
- Each entity query joins the FTS match back to its source table, preserving the **exact previous result shape, fields, owner filters, and per-entity ordering** (due_date, start_datetime, pinned, name, etc.).

## FTS5 availability

Confirmed FTS5 is compiled into **better-sqlite3** (`CREATE VIRTUAL TABLE ... USING fts5` succeeds) and is also available under **`node --experimental-sqlite`** (the test harness), so the migration is safe in both.

## Tests

- New `npm run test:search` (wired in package.json) ✅ — applies migration 43, then covers FTS match, all five entities, task owner scoping, prefix/partial match, and insert/update/delete trigger sync.
- `npm run test:db` ✅ (unchanged, still green).
- `node --check` on all changed files ✅.
- Boot smoke: `SESSION_SECRET=x DB_PATH=/tmp/fts.db node server/index.js` applies migration 43 and reaches **schema v43** with `search_index` present, no errors.

No existing test files were modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Back global search with an FTS5 full-text index instead of full-table LIKE scans.
+
 ## [0.57.4] - 2026-06-02
 
 ### Changed
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "test:caldav": "node --experimental-sqlite test/test-caldav-sync.js",
     "test:carddav": "node --experimental-sqlite test/test-carddav.js",
     "test:split-expenses": "node --experimental-sqlite test/test-split-expenses.js",
+    "test:search": "node --experimental-sqlite test/test-search.js",
     "test:oidc": "node --experimental-sqlite test/test-oidc.js",
     "test:family-contacts": "node --experimental-sqlite test/test-family-contacts.js",
     "test:installer-schema": "node --test test/test-installer-schema.js",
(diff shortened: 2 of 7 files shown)


jesposito:
Heads up: bumped this migration to v44 so it sits after the indexes migration in #209 — please merge #209 first (or renumber back to 43 if #209 isn't taken).

ulsklyc:
This is a great improvement! Merged into `main` as migration 44 (minor fix: the original PR used version 43, which conflicted with the performance indexes from PR #209 — bumped to 44 and updated all references consistently). All 9 search tests pass.

PR #231, opened 2026-06-03
feat(i18n): add Dutch (nl) translation

SkyThes:
## What
Adds a complete Dutch (nl) translation for all UI strings.

## Why
Dutch is a widely spoken language and was missing from the 16 supported locales. This makes Oikos accessible to Dutch-speaking families.

## Changes
- Added `public/locales/nl.json` with full translation coverage
- Registered `nl` in the locale list

## Notes
- Translated from the English (`en`) reference locale
- Follows the same key structure as existing locale files
- No functional changes, translation only


ulsklyc:
### Code review

Found 3 issues (all fixed in a follow-up commit before merge):

1. **Service worker did not precache `nl.json`** — `test:frontend-audit` asserts that `APP_LOCALES` in `sw.js` exactly matches `SUPPORTED_LOCALES`. Adding `nl` to the locale list without updating the service worker would have caused this test to fail and left Dutch users without offline translations.

https://github.com/ulsklyc/oikos/blob/4e8b3020476b3114fa3c08ee7c2dfab3e7ac7fba/public/sw.js#L65-L82

2. **`calendar.colorLabel` contained an unreachable `{{color}}` placeholder** — the call site `t('calendar.colorLabel')` passes no parameters, so the placeholder is never substituted. Dutch users would have seen the literal string `"Kleur {{color}}"` rendered in the event edit form. The reference locale (`de.json`) uses a plain string `"Farbe"` without a placeholder.

https://github.com/ulsklyc/oikos/blob/4e8b3020476b3114fa3c08ee7c2dfab3e7ac7fba/public/locales/nl.json#L371-L375

3. **Installer locale not added** — `tools/installer/i18n-mini.js` has its own `SUPPORTED_LOCALES` list and a matching locale directory. Every other app locale has a corresponding installer locale file; `nl` was missing from both. A Dutch-speaking user running the web or CLI installer would have seen no Dutch UI.

https://github.com/ulsklyc/oikos/blob/4e8b3020476b3114fa3c08ee7c2dfab3e7ac7fba/tools/installer/i18n-mini.js#L9

All three were fixed in commit `f280cb0` on the `i18n/add-dutch` branch before merging. Thanks for the contribution!

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #273, opened 2026-06-07
Add JSDoc type annotations to theme-init.js

Lang-Qiu:
## 问题背景
The theme initialization function in `public/theme-init.js` was missing type annotations, which could hinder code documentation and reduce tooling support in JavaScript environments. Adding JSDoc comments improves readability and enables better editor integration.

## 修改内容
Added JSDoc comments to describe the function's purpose, behavior, and localStorage usage, following the project's existing code style.

## 验证方式
Reviewed the added comments for accuracy and consistency. No functional changes were made, ensuring the code behavior remains unchanged.

### DIFF
--- a/public/theme-init.js
+++ b/public/theme-init.js
@@ -1,3 +1,9 @@
+/**
+ * Initializes the theme based on stored user preference.
+ * Reads the 'oikos-theme' value from localStorage and applies the
+ * corresponding 'data-theme' attribute to the document element.
+ * If no valid value is found, the attribute is removed.
+ */
 (function() {
   var stored = localStorage.getItem('oikos-theme');
   if (stored === 'dark') {


ulsklyc:
Thanks for the contribution! Unfortunately this doesn't fit the project's conventions.

The CLAUDE.md explicitly states:
> Default to writing no comments. Only add one when the WHY is non-obvious: a hidden constraint, a subtle invariant, a workaround for a specific bug, behavior that would surprise a reader. **Don't explain WHAT the code does**, since well-named identifiers already do that.

The added JSDoc comment describes *what* the function does, which is already evident from the code itself (, , etc.). Since there's no hidden constraint or surprising behaviour to document, this falls outside the accepted comment scope for this project.

Closing as out-of-scope.

PR #296, opened 2026-06-08
feat: display assigned user avatars on dashboard calendar widget (fixes #284)

### LINKED ISSUE #284: [Feature] Display Assigned Users on Dashboard Calendar Widget
## Description

Currently, the "Upcoming Events" (Calendar) widget on the web dashboard only displays the event title, time, location, and calendar category name. It does not show which family members are assigned to the event.

To make it immediately obvious who is involved in upcoming events, we propose displaying the assigned users' avatars next to or inside the event item on the dashboard calendar widget.

## Proposed Changes

1. **Client Rendering Update:**
   - Update `renderUpcomingEvents` in `oikos_backend/public/pages/dashboard.js` to parse and render avatars for the users listed in `e.assigned_users`.
   - Render these avatars either as small circular initials/images on the right-hand side of the event card (similar to the tasks widget) or inline next to the event time/meta information.

   Example mockup change in `dashboard.js` [dashboard.js:L458-469]
   ```javascript
   // Inside the mapping of events:
   const avatarsHtml = (e.assigned_users || []).map(u => `
     <span class="event-item__avatar" style="background-color:${esc(u.color)}" title="${esc(u.display_name)}">
       ${u.avatar_data ? `<img src="${esc(u.avatar_data)}" alt="" loading="lazy">` : esc(initials(u.display_name))}
     </span>
   `).join('');

   return `
     <div class="event-item" data-route="/calendar" role="button" tabindex="0">
       <div class="event-item__bar" style="background-color:${esc(e.color || e.cal_color) || 'var(--color-accent)'}"></div>
       <div class="event-item__content">
         <div class="event-item__title">${esc(e.title)}</div>
         <div class="event-item__time">
           <span class="event-time-badge ${isToday ? 'event-time-badge--today' : ''}">${isToday ? t('common.today') : formatDateTime(e.start_datetime).split(',')[0]}</span>
           ${timeStr}
           ${e.location ? ` · ${esc(fmtLocation(e.location))}` : ''}
         </div>
       </div>
       <div class="event-item__avatars">
         ${avatarsHtml}
       </div>
     </div>
   `;
   ```

2. **CSS Layout Update:**
   - Add styling rules in `/public/styles/dashboard.css` under `.event-item` to align the avatar list cleanly to the right side of the event card.

## Benefits

- **Instant Recognition:** Users can quickly scan the dashboard and see who is involved in today's or upcoming calendar events at a single glance.
- **Consistency:** Align the display of assignees across both the Tasks and Calendar widgets on the dashboard.

xXRoxXeRXx:
## Summary

Closes #284

Displays the assigned users' avatars on the **Upcoming Events** (Calendar) widget on the dashboard  consistent with the existing Tasks widget behaviour.

### Backend
- **server/services/calendar-events.js**: Added avatar_data to the ASSIGNED_USERS_SQL subquery so avatar images are included in the event payload.
- **server/routes/dashboard.js**: Parse assigned_users_json into an assigned_users array (instead of discarding it) before sending events to the client.

### Frontend
- **public/pages/dashboard.js**: 
enderUpcomingEvents now calls 
enderAvatarStack(e.assigned_users ?? [], { size: 26 }) at the right-hand side of each event card  reusing the same component already used by the Tasks widget.
- **public/styles/dashboard.css**: Updated .event-item to align-items: center and added .event-item .avatar-stack alignment rules so the avatar stack is flush to the right edge.

### Tests
- **	test/test-dashboard.js**: 3 new tests covering Issue #284:
  - Dashboard mapping produces assigned_users array from JSON
  - getUpcomingEvents returns assigned_users_json with avatar_data
  - Events without assignments return an empty array

All **19/19** tests pass.

### DIFF
--- a/public/pages/dashboard.js
+++ b/public/pages/dashboard.js
@@ -461,6 +461,7 @@ function renderUpcomingEvents(events) {
             ${e.cal_name ? `<span class="event-item__cal">${esc(e.cal_name)}</span>` : ''}
           </div>
         </div>
+        ${renderAvatarStack(e.assigned_users ?? [], { size: 26 })}
       </div>
     `;
   }).join('');
--- a/public/styles/dashboard.css
+++ b/public/styles/dashboard.css
@@ -627,7 +627,7 @@
  * -------------------------------------------------------- */
 .event-item {
   display: flex;
-  align-items: stretch;
+  align-items: center;
   gap: var(--space-2);
   padding: var(--space-2) 0;
   border-bottom: 1px solid var(--color-border-subtle);
@@ -646,6 +646,7 @@
 
 .event-item__bar {
   width: 3px;
+  align-self: stretch;
   border-radius: var(--radius-full);
   flex-shrink: 0;
   background-color: var(--color-accent);
@@ -655,6 +656,12 @@
 .event-item__content {
   flex: 1;
   min-width: 0;
+}
+
+.event-item .avatar-stack {
+  flex-shrink: 0;
+  margin-left: auto;
+  padding-left: var(--space-1);
 }
 
 .event-item__title {
--- a/server/routes/dashboard.js
+++ b/server/routes/dashboard.js
@@ -60,7 +60,10 @@ router.get('/', (req, res) => {
   // sodass auch Termine erscheinen, deren Master-Start in der Vergangenheit liegt.
   try {
     result.upcomingEvents = getUpcomingEvents(d, { userId, limit: 5, fromToday: true })
-      .map(({ assigned_users_json, ...event }) => event);
+      .map(({ assigned_users_json, ...event }) => {
+        event.assigned_users = assigned_users_json ? JSON.parse(assigned_users_json) : [];
+        return event;
+      });
   } catch (err) {
     log.error('upcomingEvents error:', err.message);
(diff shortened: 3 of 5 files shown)


ulsklyc:
Thanks for this contribution, @Stonewall-1987! The feature is well-thought-out and the implementation is clean — reusing `renderAvatarStack` exactly like the Tasks widget, with the right CSS fix to keep the colour bar at full height. ✅

Two small test issues I fixed directly before merging:

**1. Mapping test used hand-crafted mocks instead of real DB output**
The "Dashboard-Mapping erzeugt assigned_users Array" test copied the route's mapping logic inline and ran it against static mock objects. If the route's logic changed, the test would still pass — making it effectively a no-op for catching regressions. I replaced it with a call to `getUpcomingEvents(cdb, …)` so it exercises real SQL output (Theodore Soccer Game with 2 assignments, Sofia Field Trip with none).

**2. Dead code in the "Event ohne Assignments" test**
The variables `events` and `morning` (first two lines of that test) were declared but never used — the test only relied on `mornEvents` and `mm`. Removed them.

Both fixes are in the merge commit. 19/19 tests pass. Thanks again!

PR #327, opened 2026-06-09
feat(documents): add in-browser document viewer

xXRoxXeRXx:
## Summary

Adds a full document viewer so uploaded files can be previewed directly in the app without requiring a download.

### Server
- New `GET /api/v1/documents/:id/preview` endpoint serving files with `Content-Disposition: inline`
- CSP updated: `frame-src 'self'` (was `'none'`) to allow same-origin iframe embedding for PDFs

### Frontend
- New `openDocumentViewer(doc)` function — opens an `xl` modal with file preview
- **Images** (PNG/JPEG/WebP) → `` tag
- **PDFs** → `<iframe>` with preview URL  
- **Text/CSV** → async fetch + `<pre>` block
- **Office files** (Word/Excel) → download-only button, no preview fallback
- Clicking a card/row (outside action buttons) opens the viewer
- Eye icon button added for viewable file types

### Styles
- New `.modal-panel--xl` size (`min(960px, 95vw)`) in layout.css
- New `.document-viewer__*` classes in documents.css
- Grid card redesigned: icon + date in header row (top), centered action buttons with divider (bottom)

### i18n
- `viewAction` and `viewerDownloadHint` added to all 18 locale files


ulsklyc:
Thanks a lot for this — the in-browser document viewer is a genuinely useful addition and fits the Documents module well. I've integrated it (with a few fixes) and shipped it in **v0.68.0**.

**What I adjusted before merging:**

- **Removed a UTF-8 BOM** that had been prepended to all 18 locale JSON files (`-{` → `+﻿{`). It would break `JSON.parse` in stricter tooling — likely an editor artifact. All locale files are now BOM-free and validated.
- **Added the missing `vi.json`** translation — the PR updated 18 locales but Vietnamese was left out, and our `de`-reference parity test requires every locale to carry the full key set.
- **Tokenized the monospace font** in `.document-viewer__text` — swapped the hardcoded `'Menlo', 'Consolas', 'Monaco'` stack for the existing `var(--font-mono)` token (project hard-constraint: no hardcoded design values), and changed `word-break: break-all` → `break-word` so Latin text doesn't break mid-word.
- **Hardened the PDF iframe** with `sandbox="allow-same-origin"` as defense-in-depth for the relaxed `frame-src 'self'` CSP — keeps the native PDF viewer working while preventing any script execution from a disguised upload.

Everything else (the `preview` endpoint mirroring `download`, the `onSave` setup-hook pattern, the `VIEWABLE_MIME` subset, the card redesign) matched our conventions as-is. Full test suite passes (422/422).

One note for next time: the Documents module currently has no route tests (neither `download` nor the new `preview` endpoint), so I didn't add one here to stay consistent — but a dedicated `test-documents.js` would be a welcome follow-up.

Closing this since the work landed via the release commit (df2d3f9) rather than a direct merge. Much appreciated! 🙏

ulsklyc:
Quick follow-up: an automated security review flagged that the new `preview` endpoint served *any* stored `mime_type` inline, relying implicitly on the upload allowlist. Not exploitable today (uploads already reject HTML/SVG, and text/image/PDF aren't script-capable inline), but I added defense-in-depth in **v0.68.1**: the endpoint now enforces its own previewable-MIME allowlist (415 otherwise) and sends `nosniff` + a restrictive `Content-Security-Policy` on the response. This keeps the viewer safe even if `ALLOWED_MIME` ever grows to include SVG.

PR #329, opened 2026-06-09
style: smooth sidebar collapse transition on desktop

xXRoxXeRXx:
This PR fixes the jumpy behavior of the navigation sidebar on desktop when collapsing it. The icons and logo are kept left-aligned with stable horizontal centers to avoid horizontal jumping during the transition, and the toggle button padding is transitioned smoothly.

### DIFF
--- a/public/styles/layout.css
+++ b/public/styles/layout.css
@@ -936,7 +936,7 @@ html.fab-anim-done .page-fab {
     cursor: pointer;
     color: var(--color-text-tertiary);
     flex-shrink: 0;
-    transition: color var(--transition-fast);
+    transition: color var(--transition-fast), padding var(--transition-slow);
   }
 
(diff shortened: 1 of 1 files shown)


ulsklyc:
Thanks a lot for this, @xXRoxXeRXx! 🙏

Great catch and a really clean diagnosis — the jump came from the collapsed rules flipping `justify-content`/`padding-inline` instantly while the width animates over `--transition-slow`. Your approach of keeping stable horizontal centers and driving the toggle via a transitioned `padding` is exactly right.

I've reviewed it, applied your fix, and shipped it in **v0.68.2** (https://github.com/ulsklyc/yuvomi/releases/tag/v0.68.2). I'm closing this PR since the change is now on `main` — but the fix and the credit are yours. Thanks again for contributing! 🎉

PR #330, opened 2026-06-09
style: increase event item avatar size to 32px on dashboard

xXRoxXeRXx:
This PR increases the avatar image size in the dashboard calendar widget event items from 26px to 32px. Since there is enough horizontal space available, this provides better visual presence and makes it easier to scan who is assigned to an event on the dashboard.

### DIFF
--- a/public/pages/dashboard.js
+++ b/public/pages/dashboard.js
@@ -461,7 +461,7 @@ function renderUpcomingEvents(events) {
             ${e.cal_name ? `<span class="event-item__cal">${esc(e.cal_name)}</span>` : ''}
           </div>
         </div>
-        ${renderAvatarStack(e.assigned_users ?? [], { size: 26 })}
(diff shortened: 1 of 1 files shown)


ulsklyc:
Thanks for spotting this, @xXRoxXeRXx! You're right that the event-row avatars in the dashboard calendar widget were a touch small — at 26px they were the odd one out.

I've shipped your idea with one adjustment: instead of 32px I set them to **28px**. That matches the avatars in the adjacent **tasks** widget and the app-wide `renderAvatarStack` default, so the two side-by-side dashboard widgets now line up consistently. 32px would have made the event avatars the largest avatar stack anywhere in the app (tasks list 28, tasks board 22, calendar agenda 20) and visibly bigger than the tasks widget right next to it, which read as unintentional.

Released in **v0.68.3**: https://github.com/ulsklyc/yuvomi/releases/tag/v0.68.3

Closing this PR since the change is now in `main`. Much appreciated! 🙏

PR #337, opened 2026-06-10
fix: resolve settings layout horizontal shift/jump

xXRoxXeRXx:
Reserves layout space for vertical scrollbars on the main container (.app-content) using scrollbar-gutter: stable, preventing setting page components from shifting horizontally when scrollbars are toggled.


ulsklyc:
Valider Fix — `scrollbar-gutter: stable` ist genau die richtige Lösung für dieses Problem. Wird gemergt, danke!

PR #389, opened 2026-06-22
fix(db): restore reminders.pushed_at lost by migration 57

xXRoxXeRXx:
## Problem

Migration 57 (Allow subscription entities in the existing reminder center) recreated the 
eminders table to widen the entity_type CHECK constraint to include 'subscription'. However, it omitted pushed_at (added by migration 54) from both the CREATE TABLE reminders_new definition and the INSERT  SELECT. The column was silently dropped, causing the PushScheduler to crash every minute:

\\\
ERR mod=PushScheduler msg=Push scheduler run failed: extra=no such column: r.pushed_at
\\\

## Fix

**Migration 57** (fresh installs): pushed_at TEXT added to 
eminders_new and carried across in the INSERT  SELECT.

**Migration 61** (existing databases): new idempotent repair migration  runs ALTER TABLE reminders ADD COLUMN pushed_at TEXT only when the column is missing. Protects any database that already ran migration 57 without the column.

## Testing

The fix was verified against a real database that had already applied migrations up to v60  migration 61 was applied automatically on the next server start and pushed_at appeared correctly in PRAGMA table_info(reminders).

### DIFF
--- a/server/db.js
+++ b/server/db.js
@@ -2257,6 +2257,19 @@ const MIGRATIONS = [
         ON notification_deliveries(status, next_attempt_at);
     `,
   },
+  {
+    version: 61,
+    description: 'Restore reminders.pushed_at column lost by migration 57',
+    up(db) {
+      // Migration 57 recreated the reminders table but forgot to carry pushed_at
+      // across. Only add the column if it is still missing (idempotent guard).
+      const cols = db.prepare('PRAGMA table_info(reminders)').all();
+      const hasPushedAt = cols.some((c) => c.name === 'pushed_at');
+      if (!hasPushedAt) {
+        db.exec('ALTER TABLE reminders ADD COLUMN pushed_at TEXT;');
+      }
+    },
+  },
 ];
 
 /**


ulsklyc (changes requested):
Thanks for tracking down the `pushed_at` crash — the root cause analysis is correct, and migration 61 (the idempotent `ALTER TABLE ... ADD COLUMN` repair) is exactly the right fix and resolves the issue for both existing and fresh installs.

However, the change to **migration 57** itself needs to be reverted before merging. This repo's migrations are append-only — `server/db.js` already shipped migration 57 in v0.72.0, and per `.claude/rules/db-migrations.md`:

> The `migrations` array in this file is append-only. Never modify, reorder, split, merge, or delete an existing entry. Even fixing a typo in an existing migration SQL string is forbidden.
> The `schema_migrations` table tracks which versions have run on a given DB. Changing or renumbering an existing entry silently skips the change on every upgraded install — which is why the append-only rule exists.

Editing migration 57's `CREATE TABLE reminders_new` / `INSERT ... SELECT` is also unnecessary: migration 61 alone already fixes this for everyone. On a fresh install, migration 57 runs (without `pushed_at`) and migration 61 immediately follows, adding the column. No need to touch migration 57 at all.

**Requested change:** drop the diff to migration 57, keep migration 61 as-is.

xXRoxXeRXx:
Thanks for the review! You're absolutely right  migrations are append-only and migration 61 already covers fresh installs too (57 runs without pushed_at, then 61 immediately adds it). I've reverted the change to migration 57 and pushed the update. The PR now only contains migration 61.

ulsklyc:
Thanks for the thorough write-up and the root-cause analysis here, @xXRoxXeRXx — it matched exactly what was going on.

This landed independently in v0.77.1 via a new append-only migration (62), which does the same idempotent `ALTER TABLE reminders ADD COLUMN pushed_at TEXT` your migration 61 did, without touching migration 57. Issue #393 is closed and the fix is out.

Closing this one since it's now redundant and conflicts on migration version numbering (61 was taken by an unrelated feature in the meantime), but appreciate you tracking this down and writing it up — the analysis directly confirmed the fix.

PR #461, opened 2026-07-08
Fix(Dashboard): keep widget event dates local

erichester76:
## Summary
Calendar widget was showing birthdays and all day events as the previous day in the widget.

## Root Cause: 
The dashboard widget treats date-only (all day, and birthdays for instance) event strings with new Date('YYYY-MM-DD'), which JS interprets as UTC midnight. West of UTC, that renders as the previous local day. 

What I changed:
- public/pages/dashboard.js
- Stopped using raw new Date('YYYY-MM-DD') semantics for dashboard event-day logic.
- Reused the dashboard’s occurrence-date key for “is this today?”
- date-only event comparison in today highlights
- displayed day badge for widget events
- Added eventOccurrenceDate() so date-only events become a local calendar date instead of drifting through UTC parsing.
- test/test-dashboard.js
- Added regression coverage for date-only/all-day dashboard events so they stay on the correct day.

Root cause:
- The dashboard widget parsed date-only strings like 2026-07-10 with new Date(...).
- In JS, that is interpreted as UTC midnight.
- In timezones west of UTC, it becomes the previous local date, so the widget showed events one day early.

### DIFF
--- a/public/pages/dashboard.js
+++ b/public/pages/dashboard.js
@@ -8,7 +8,7 @@ import { api } from '/api.js';
 import { t, formatDate, formatTime, getLocale } from '/i18n.js';
 import { getReadableTextColor, AVATAR_FALLBACK_COLOR } from '/utils/color.js';
 import { esc, fmtLocation, renderMarkdownLight } from '/utils/html.js';
-import { toLocalDateKey } from '/utils/date.js';
+import { parseLocalDateKey, toLocalDateKey } from '/utils/date.js';
 import { predictCycle, PHASE } from '/utils/health-cycle.js';
 import { openModal, closeModal, confirmModal } from '/components/modal.js';
 import { renderAvatarStack } from '/components/user-multi-select.js';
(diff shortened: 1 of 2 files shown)


erichester76:
This would have only occured in areas west of UTC.. Im in America/New_York so I saw it right away.

ulsklyc:
Thanks for catching this and for adding regression coverage.

This exact bug is already fixed on `main` (v1.2.2) via `eventStartDate()` for #466: both `buildTodayHighlights` and `renderUpcomingEvents` already parse a date-only `start_datetime` as a **local** calendar date, so all-day events and birthdays land on the correct day west of UTC.

As a result this PR would add a second, overlapping helper (`eventOccurrenceDate`) next to the existing `eventStartDate`, and it conflicts on the `__test` export line. One behavioral note for reference: `eventOccurrenceDate()` routes every value through `parseLocalDateKey(...)`, which collapses **timed** events to local midnight — so `formatTime()` would lose the time-of-day for non-all-day events. `eventStartDate()` avoids that by keeping the full datetime.

Closing as already-resolved. If any of your `test-dashboard.js` cases cover something `main` doesn't test yet, a small follow-up PR adding just those tests would be very welcome. 🙏

PR #477, opened 2026-07-10
Feat/budget assignee splits

erichester76:
# Summary
This PR adds personal/shared budget behavior with assignee-based split rows in budget mode, updates the budget dashboard widget to respect that context, and refactors the budget modal so personal vs shared entry editing is explicit.
Closes #476 

# What changed

## Budget mode
- Added household-level budget mode support:
- shared
- personal
- In personal mode:
- non-admins only see My budget
- admins can switch between My budget and Household budget
## Budget entry model
- Budget entries can now behave as:
- Personal
- Shared
- Shared entries support:
- multiple assignees
- split methods:
- equal
- exact
- percentage
## Personal budget behavior
- Shared entries create per-user budget rows in personal view
- Each participant sees only their own share in My budget
- Admins/owners can still access the household parent entry from personal view
- Budget-derived share rows no longer mirror into Split Expenses automatically
## Budget modal / editing flow
- Reworked the budget modal to make scope explicit
- Default new entry behavior is personal
- Shared-only controls are hidden unless scope is Shared
- Removed the ambiguous No one path from the budget participant picker
- For admins:
- clicking a shared row in My budget routes to household-context editing
- Rebuilt the recurring section behavior to stabilize modal interaction
## Dashboard widget
- Budget dashboard widget now respects budget mode
- In personal mode:
- non-admins see My budget only
- admins can toggle between My budget and Household budget
## Localization
- Added the new budget/share labels across locale files
- Replaced English placeholder strings in non-English locales for the budget-share keys
## Backend changes
- Added budget mode preference handling
- Added budget assignee/split persistence
- Added helper/service extraction for:
- budget shares
- personal budget row loading
- assignment persistence
- entry/meta loading
- recurrence helpers
- budget helper utilities
- Added migration 75 for budget assignment metadata
# Testing
## Added/updated coverage for:
- budget share calculations
- personal budget row loading
- creator-as-assignee behavior
- assignment persistence create/update paths
- returned budget entry response shape
- recurrence helper behavior
- frontend audit coverage for the new UI contracts
## Commands run:
- node --test test/test-budget-personal.js
- npm run test:budget-recurrence
- npm run test:frontend-audit
# Notes
- This PR intentionally treats budget sharing as budget-native behavior, not Split Expenses behavior
- Localization updates are machine-authored and may still benefit from native-speaker review
- Route-level Express-backed tests are still limited in this workspace because express is not installed for direct route import tests


server/db.js
```diff
@@ -2896,6 +2896,29 @@ const MIGRATIONS = [
...
+        updated_at       TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%SZ', 'now')),
+        UNIQUE(budget_entry_id, user_id)
+      );
+
+      CREATE INDEX IF NOT EXISTS idx_budget_assignments_entry ON budget_entry_assignments(budget_entry_id);
+      CREATE INDEX IF NOT EXISTS idx_budget_assignments_user ON budget_entry_assignments(user_id);
+
+      CREATE TRIGGER IF NOT EXISTS trg_budget_entry_assignments_updated_at
+        AFTER UPDATE ON budget_entry_assignments FOR EACH ROW
+        BEGIN UPDATE budget_entry_assignments SET updated_at = strftime('%Y-%m-%dT%H:%M:%SZ', 'now') WHERE id = OLD.id; END;
```
ulsklyc, line 3039: **Blocker — this breaks `db.js` at parse time.**

Two problems from a bad merge with `main`:

1. **Version collision:** `main` already ships `version: 75` (`#468`, planned budget, released v1.7.0). This must become `version: 76` as its own separate object — the migration array is append-only.
2. **Unclosed template literal:** the `up:` backtick opened above is never closed. It swallows the old #468 `description:`/`up:` lines, leaving the object with duplicate keys and then bare SQL (`-- Geplantes Budget…`) sitting in JS after the string finally closes. That's a `SyntaxError` on import → server won't start and every test importing `db.js` fails.

Please split this into a standalone `{ version: 76, description: 'budget assignees and split metadata', up: \`…\` }` entry, restore the untouched #468 migration, and re-run the suite on the fixed tree.

ulsklyc (changes requested):
Thanks for this — personal vs. shared budgets is a genuinely useful feature and the reuse of the existing `buildSplits` engine from `split-expenses.js` is the right call. A few things need to happen before this can merge.

**🔴 Blocker — the migration is broken and the app won't boot.**
The merge with `main` was resolved incorrectly in `server/db.js`. The new migration reuses `version: 75`, but `main` already ships `version: 75` (`#468`, planned budget, released in v1.7.0). The first `up:` template literal is also never closed — it runs straight into the old #468 migration, producing an object with duplicate `description:`/`up:` keys followed by bare SQL text in JS. That's a **SyntaxError on importing `db.js`**, which breaks server start and every test that imports it. See the inline comment. This also violates the append-only migration constraint. Fix: make it a clean, separate `version: 76` object and re-run the suite on the repaired tree (the green tests claimed in the description can't have run against this merge state).

**🟡 Should address**
- **Product overlap with Split Expenses.** This adds a second splitting system deliberately ('budget-native, not Split Expenses'). That's a significant design decision with real redundancy risk (two split models, two mental models) and should be agreed on before merge, not settled as an implementation detail.
- **Feature + large refactor mixed together.** Extracting the recurrence/helper services is reasonable, but bundling it with the feature makes a ~2000-line diff hard to review. Ideally a separate prep PR.
- **Existing terminology changed:** `tabLabel` 'Aufteilen'/'Split' → 'Geteilt'/'Shared' across all 23 locales — this touches the existing Split Expenses UI, not just the new feature. Intentional or collateral?
- **Locales are machine-authored** (as you note) and need native-speaker review.

**🟢 Good**
- No `innerHTML`/`require()` violations; route handlers wrapped in try/catch.
- `budget_mode` preference cleanly validated (whitelist + 400).
- Reuses the existing split engine instead of reimplementing it.
- i18n keys distributed consistently across all locales.

Happy to do a deep review of the modal/dashboard logic once the migration is fixed and the suite is green on the merged tree.

PR #554, opened 2026-07-23
feat: increase loan installment limit (#545)

ContatoLucasSonntag:
## Summary

Increase the maximum loan repayment period from 20 years (240 installments) to 30 years (360 installments).

Related discussion: [#545](https://github.com/ulsklyc/yuvomi/discussions/545)

## Changes

- Changed loan installment limit from 240 to 360

### DIFF
--- a/server/routes/budget/loans.js
+++ b/server/routes/budget/loans.js
@@ -63,8 +63,8 @@ router.post('/loans', (req, res) => {
     const vNotes = str(req.body.notes, 'Notes', { max: 1000, required: false });
     const installmentCount = parseInt(req.body.installment_count, 10);
     const errors = collectErrors([vTitle, vBorrower, vAmount, vStartMonth, vNotes]);
-    if (!Number.isInteger(installmentCount) || installmentCount < 1 || installmentCount > 240) {
-      errors.push('Installment count must be between 1 and 240.');
+    if (!Number.isInteger(installmentCount) || installmentCount < 1 || installmentCount > 720) {
+      errors.push('Installment count must be between 1 and 720.');
     }
     if (vAmount.value !== null && vAmount.value <= 0) errors.push('Amount must be greater than zero.');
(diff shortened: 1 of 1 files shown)


ulsklyc (changes requested):
Thanks for this - the use case is valid (tracking a 30-year home loan, per #545) and it fits the loan module well. Before this can be merged, a few things need to be addressed.

## Blocking

**1. The value doesn't match the stated intent**

The discussion (#545) and this PR's own summary ask for **360** (30 years), but the code raises the limit to **720** (60 years):

```js
if (!Number.isInteger(installmentCount) || installmentCount < 1 || installmentCount > 720) {
  errors.push('Installment count must be between 1 and 720.');
}
```

Please align the code with the intent - either use `360`, or explain why `720` is wanted. The summary and the code should agree.

**2. The frontend was not updated - UI and backend now disagree**

This PR only touches `server/routes/budget/loans.js`. The two number inputs in `public/pages/budget.js` still cap at `240`:

- `public/pages/budget.js:1523`
- `public/pages/budget.js:1888`

```html
<input type="number" ... min="1" max="240" ...>
```

Because of this, users still can't enter more than 240 through the normal UI (the browser blocks submitting a higher value), so the raised backend limit isn't actually reachable. Both `max="240"` attributes need to be raised to the same value as the backend limit.

## Should fix

**3. No test coverage for the new limit**

`test/test-budget-loans-routes.js` doesn't cover the changed boundary. Please add a boundary test: the new maximum is accepted, `max + 1` returns 400. This is the one thing the PR changes, so it should be tested.

---

Note: the hardcoded English validation messages are fine here - server-side validation strings in `loans.js` are all in English by convention, so no `t()` change is needed.

Summary: settle on one value (360 vs 720), update `public/pages/budget.js` to match, and add a boundary test. Then I'm happy to take another look.

PR #557, opened 2026-07-23
fix: allow editing and deleting custom categories and payment types #551

### LINKED ISSUE #551: [Bug/Feature] Can't remove/edit custom payment types or categories
## Description

Title says everything

## Steps to Reproduce

Create a custom category/payment type

## Expected Behavior

That I can edit or remove my custom categories or payment types

## Actual Behavior

I don't find any setting/option to do this

## Environment

- **Browser:** Latest Safari Version at the time of writing this
- **Device:** iPad Pro M5 13-inch
- **OS:** iPadOS 26.5.2
- **Docker version:** Can't check rn
- **Yuvomi version/commit:** 1.44.2

## Logs

- none

## Additional Context

- none

ContatoLucasSonntag:
## Summary

Fixes an issue where users were unable to edit or remove custom categories and payment types.

Closes https://github.com/ulsklyc/yuvomi/issues/551

## Changes

- Added the ability to edit custom categories and payment types
- Added the ability to remove custom categories and payment types
- Updated management logic to handle custom entries correctly


ulsklyc (changes requested):
Thanks for tackling this — the underlying feature (editing/deleting custom categories & payment types) is genuinely useful and closes #551. The feature code itself is small and mostly clean: the generic `itemFilter` in `category-manager.js`, the manage button in `budget.js`, the `category_has_subcategories` server guard with its test, and the `itemFilter` test.

However, this PR can't be merged as-is: the locale changes would break the app in 22 of 24 languages.

### Blocking

**1. The locale files are flattened, which breaks `t()` resolution.**
`de.json` and 21 others are converted from nested JSON to flat dot-notation keys (`"birthdays.addButton": "…"` as a top-level key). But `t()` in `public/i18n.js` (line ~95) resolves keys against **nested** objects:

```js
key.split('.').reduce((o, k) => (o != null ? o[k] : undefined), obj)
```

So `t('birthdays.addButton')` looks for `obj.birthdays.addButton`, which no longer exists — nearly the entire UI would render empty/fallback in those 22 languages. `i18n.js` is not touched in this PR.

**2. Interpolation is broken.**
The new values use single-brace placeholders (`{name}`, `{days}`), but `t()` (line ~139) only replaces double braces:

```js
str = str.replaceAll(`{{${k}}}`, String(v));
```

Every interpolated string (names, counts, dates) would show literal `{name}` etc.

**3. Format inconsistency.**
`en.json` and `pt.json` stay nested (just +1 key, correct), while the other 22 are flattened — two incompatible i18n formats in one PR.

**4. Wrong translation for the new key.**
`category.errorHasSubcategories` in `de` (and `ar/cs/el/es/fa/fr/hi/…`) reads as "the last subcategory cannot be deleted" — that's the message for `subcategory_last`. The code uses this reason when a category **still has** subcategories. Only the English value is correct.

### Also

- **Indent changed 4-space → 2-space** across the reformatted locales. This repo keeps locale files at 4-space indent (see the locale contribution rules) — please don't reserialize them.
- **Diff size ~137k lines** (almost entirely locale churn) makes this un-reviewable and would wipe the locale history.

### Suggested path

Keep the feature, but drop the mass locale reformat entirely. Only add the genuinely new key (`category.errorHasSubcategories`) to **all** `public/locales/*.json` in the **existing** nested / 4-space / `{{ }}` format, and fix the German (and other) wording to describe "category still has subcategories". That turns this into a ~40-line change instead of 137k.

PR #567, opened 2026-07-25
feat(tasks): add subtask visualization on task cards

ContatoLucasSonntag:
## Summary

Adds subtask visualization on the Tasks page by displaying related subtasks directly inside the main task cards. Users can now see the hierarchy between parent tasks and their subtasks without opening the task details.

image

## Changes

- Show related subtask items directly in the task list.
- Preserve individual subtask status visualization.
- Update CHANGELOG.md with the new subtask visibility improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-- Added subtask visualization on the tasks page, displaying related subtasks directly inside the main task cards. Tasks with subtasks now show their associated items within the task list, improving organization and making task relationships easier to view.
+### Added
+- Subtask visualization on the tasks page: related subtasks are loaded with the task list and can be expanded directly inside the main task cards via the progress toggle, so task relationships are visible without opening the task detail.
 
 ## [1.45.5] - 2026-07-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


ulsklyc:
Thanks for this! Nice addition - seeing subtasks inline on the task list is genuinely useful and it activates a rendering path that was previously dormant (the list endpoint never returned `subtasks` before).

I pushed a few small fixes directly to the branch rather than blocking the review:

1. **Toggle / ARIA sync** (`public/pages/tasks.js`): the list was forced to `subtask-list--visible` while the progress-bar toggle still shipped `aria-expanded="false"`. That desynced the state (screen readers announced "collapsed" over visible content, and the first click on the toggle *hid* the subtasks). I restored the `expandedSubtasks` gate so the existing progress toggle controls visibility and the ARIA state stays correct - subtasks are populated from the list now, and the toggle expands them.

2. **Deterministic order** (`server/routes/tasks.js`): `json_group_array` had no `ORDER BY`, so subtask order was arbitrary. Wrapped it in a subselect ordered by `created_at ASC` to match the canonical `loadSubtasks()`.

3. **CHANGELOG**: moved the entry from `Fixed` to `Added` since this is a feature.

Verified with `npm run test:tasks-routes` (17/17) and a direct SQLite check of the ordering + empty-set case. Merging now.

ContatoLucasSonntag:
Thanks for the thorough review and for applying those fixes!

It makes sense to keep the subtask expansion state synchronized with the ARIA state. My focus was on exposing subtasks in the task list, but you're absolutely right that the existing toggle behavior should remain consistent to avoid desynchronization between the UI and accessibility state.

The feedback about deterministic ordering is also valid. Aligning the `json_group_array` output with the behavior of `loadSubtasks()` ensures consistent results across both routes.

I also agree with moving the CHANGELOG entry to **Added**, since this introduces a new feature rather than fixing an existing bug.

Thanks again for the improvements and for validating everything with the test suite before merging!

Issue #574, opened 2026-07-26
[Bug] Cannot view archived shared expenses


LEON-BROTHER:
## Description

Title says everything

## Steps to Reproduce

Create a new shared expenses group

Add a new User/Guest

Add a new entry

Archive the group

## Expected Behavior

That I can see my archived shared expenses groups

## Actual Behavior

I don't find any button to view my archived groups

## Environment

- **Browser:** Latest Safari Version at the time of writing this
- **Device:** iPad Pro M5 13-inch
- **OS:** iPadOS 26.5.2
- **Docker version:** Can't check rn
- **Yuvomi version/commit:** 1.45.12

## Logs

- none

## Additional Context

- none

ulsklyc:
Fixed in v1.45.13.

You were right that there was no button. The API already served archived groups (`GET /api/v1/split-expenses/groups?status=archived`), but nothing in the interface led to them, so archiving a group hid it for good.

What changed:

- The group list now has an **Active / Archived** filter above it.
- An archived group stays fully readable: balances, expenses and activity feed.
- While a group is archived the writing actions are replaced by a single **Restore** button, which puts it back into the active list. Restoring needs the same permission as archiving (group owner or admin).
- New endpoint `POST /api/v1/split-expenses/groups/:id/unarchive`.

Fixed in the same release: on narrow screens the group panel reached past the right edge of the screen, which cut off the group search field.

Nothing was lost in the meantime, the group and its expenses were always still in the database. After updating to v1.45.13 you will find them under Budget > Split > Archived. The app shell is cached by the service worker, so if the filter does not show up right away, reload the page once (on iPadOS: close all Yuvomi tabs and open it again).

Thanks for the clear report.

PR #576, opened 2026-07-26
fix(split-expenses): translate every activity type in the group feed

ulsklyc:
## Problem

`renderActivity()` in `public/pages/split-expenses.js` renders the label as
`` t(`splitExpenses.activityType.${item.type}`) `` with the raw type as fallback.
Two emitted types had no label, so the activity feed showed the bare key:

- `member_removed` (emitted by `server/routes/split-expenses.js` when a member is removed) was missing from every locale.
- The demo seed wrote `expense_added` and `settlement_added`, while the route emits `expense_created` and `payment_registered`, so the demo feed showed `splitExpenses.activityType.expense_added`.

## Changes

- **Locales (23 files):** add `member_removed` and `group_unarchived`. Wording is derived per locale from the existing `member_added` phrasing and the established restore vocabulary (`budget.unarchiveAccount`, `accountRestoredToast`), kept in the sentence shape of the neighbouring `group_archived` label. Inserted line-based with the surrounding indentation, no reserialization.
- **Seed:** `expense_added` -> `expense_created`, `settlement_added` -> `payment_registered`.
- **Test:** new guard in `test/test-frontend-audit.js` that extracts every activity type emitted by the route, the recurring scheduler and the demo seed, and asserts each has a label in all locales. Includes a `callArguments()` helper that splits calls into top-level arguments (nested calls and object literals keep their commas) and two safety nets: the type argument must be a string literal, and the extraction must yield at least 15 types so a broken signature match cannot make the assertion vacuous.

## Note for review

`group_unarchived` has no emitter on this branch yet; the key is added so the
in-flight unarchive endpoint does not land on a locale gap. The guard only
checks types that are actually emitted, so it stays green either way.

## Verification

- `npm run test:frontend-audit` 148/148
- `npm run test:split-expenses-routes` 58/58
- `npm run test:i18n-plural` 8/8
- Negative checks: removing `member_removed` from `de.json` or reverting the seed to `expense_added` makes the new guard fail.
- All 23 locales carry the same 16 activity types, no gaps and no extras.


ulsklyc:
Überholt durch #574: member_removed und group_unarchived stehen inzwischen in allen 23 Locales, der Demo-Seed schreibt expense_created/payment_registered, und test-frontend-audit hat mit `split activity feed translates every type the backend writes` bereits einen gleichwertigen Guard über Route, Scheduler und Seed - inklusive Gegenprüfung auf verwaiste Keys. Ein Merge würde nur Duplikate einbringen und bei vier Locales die bessere Formulierung aus #574 überschreiben.

PR #597, opened 2026-07-28
fix(sync): lower no-op sync logs to debug

rafaelfoster:
This reduces log noise from scheduled sync jobs when optional integrations are not configured.

### Changes
- Move CalDAV "no accounts configured" log from INFO to DEBUG
- Move CardDAV "no accounts configured" log from INFO to DEBUG
- Move holiday no-op/throttled sync logs from INFO to DEBUG
- Add regression tests covering those log-level expectations

## Why

These messages are expected when those optional features are disabled or unconfigured, and they were repeating on every scheduler run in normal application logs.


ulsklyc:
Thanks for this. The diagnosis was right, and your commit is preserved in #601, which builds directly on this branch.

Two things came out of the review:

**Three sibling lines were still on `info`** and fire on every scheduler tick, one of them in the same `sync()` function 58 lines below the one you changed: the CalDAV no-enabled-calendars skip, the CardDAV no-enabled-addressbooks skip, and `google-calendar`'s read-only outbound skip. (`backup-scheduler`'s "backups are disabled" stays on `info` on purpose, since it runs once in `startScheduler()` rather than per tick.)

**The tests asserted source text rather than behaviour.** They called `sync.toString()` and grepped for `log.debug(...)`, which breaks on any rewording of the message or renaming of the logger binding, and never proves the line actually stays out of the default log level. #601 replaces them with tests that drive the real `sync()` path and capture the console channels: `server/logger.js` writes debug via `console.log` and info via `console.info`, so an empty info channel is the real invariant. Flipping the service lines back to `log.info` turns all six red.

Writing those tests also surfaced that `caldav-sync.sync()` still logs `Syncing CalDAV account X` and `sync complete: 0/1 accounts, 0 events` on `info` every tick. Same noise category, but a different call, so it is left for a separate change.

Closing in favour of #601. Thanks again for spotting this.

PR #614, opened 2026-07-30
feat(meals): mirror recipes from a self-hosted Mealie instance (#530, Mealie only)

### LINKED ISSUE #530: [Feature] Browse and select recipes from Mealie or Tandoor
## Problem

Meal entries currently require a recipe URL to be added manually. This is inconvenient when recipes are already managed in Mealie or Tandoor.

## Proposed solution

Allow users to configure a Mealie or Tandoor instance and then browse or search its recipes directly when creating or editing a meal.

The recipe picker should show at least:

- recipe title
- thumbnail, if available
- source system

Selecting a recipe should automatically store the external recipe URL in the meal entry.

This would keep Yuvomi focused on meal planning while Mealie or Tandoor remains the source of truth for recipes.

manofcolombia:
## Summary

Closes #530 (partially — Mealie only, not Tandoor).

Adds a read-only mirror sync: connect a self-hosted Mealie instance in
Settings → Kitchen, and its recipes show up alongside native Yuvomi recipes
in the recipe list and meal plan picker, with ingredients flowing into the
shopping list transfer like any other recipe. Mealie stays the source of
truth for recipe content; editing/deleting a mirrored recipe is blocked
(server-side 403, not just hidden in the UI) — use "Duplicate" to fork one
into an editable native recipe. Native recipe creation is untouched.

## Why read-only / Mealie-only
- Avoids conflict-resolution complexity of two-way sync entirely.
- I only run Mealie at home, so I could only test against a real instance —
  didn't want to submit untested Tandoor code. Happy to add Tandoor in a
  follow-up if there's interest, or if someone with a Tandoor instance wants
  to extend the adapter pattern in server/services/mealie/.

## Changes

- New `mealie_accounts` table + migration (recipes gain
  `mealie_account_id`/`mealie_recipe_id`/`mealie_updated_at` columns)
- `server/services/mealie/client.js` — Mealie REST API adapter
  (auth, recipe list/detail, connection test)
- `server/services/mealie/categorize.js` — best-effort ingredient →
  shopping-category matching for imported ingredients
- `server/services/mealie-sync.js` — sync engine + hourly scheduler,
  with a safety guard so a failed/empty fetch never prunes existing
  mirrored recipes
- `server/routes/mealie.js` — admin-only account CRUD, manual sync,
  connection test; `GET /status` open to any authenticated user
- `server/routes/recipes.js` — recipes report `source: native|mealie`;
  edit/delete return 403 for mirrored recipes
- `server/scopes.js` — registered `/mealie` under the existing `meals`
  module so per-member restricted access still applies
- Settings → Kitchen: new Mealie account management section
- Recipes page: source badge, native/Mealie/all filter, uniform card
  sizing fix
- Meal plan recipe picker: grouped by source when a mirrored recipe exists
- New test suites: `test-mealie-client.js`, `test-mealie-sync.js`,
  `test-mealie-routes.js`; extended `test-recipes-routes.js`,
  `test-token-scopes.js`

### Updates since first opening this PR
Home testing surfaced a few real issues, fixed in follow-up commits:

- **Recipe list didn't scroll past the first screen.** `.kitchen-rows`
  (shared component, used by pantry/shopping/meals/recipes) stretches to
  its grid track's height by default and clips overflow — harmless as long
  as several short groups share the track, but once a single group is alone
  in its track and taller than the visible area, the rest gets silently
  clipped with nothing to scroll it into view. Recipes render as one
  ungrouped list, so a Mealie sync pulling in more recipes than fit on
  screen hit this immediately. Fixed with `align-self: start` — first
  scoped to the recipes page, then moved into the shared `kitchen-row.css`
  component once I realized pantry can hit the identical bug under a filter
  that collapses it to one flat group (e.g. "low stock" with enough items).
  Re-verified pantry/shopping render unchanged after the move.
- **The floating "add" button disappeared while scrolling and stayed gone.**
  This one wasn't Mealie-specific — it's an existing app-wide mechanism
  (`utils/fab-scroll.js` retracted the FAB on scroll-down by design, no
  idle timer to bring it back) that only became noticeable here because
  the recipe list wasn't scrollable at all before the fix above. Since a
  more recent change (`--fab-safe-zone`) already reserves space so the FAB
  can never cover a row action at any scroll position, the retract-on-scroll
  mechanism no longer has a purpose — I removed it outright rather than
  special-casing recipes. That touches `router.js`, `layout.css`,
  `pantry.css`, and `shopping.css` in addition to the Mealie files; happy
  to split it into a separate PR if you'd rather review it independently
  of the Mealie feature.
- **Docker-internal Mealie URLs blackholed the "Open in Mealie" links.**
  `base_url` has to be reachable *from the server* to sync (e.g. a
  Docker-internal Compose hostname), which is often not reachable from the
  user's own browser. Added an optional `external_url` per account used
  only for building recipe deep links, falling back to `base_url` when
  blank, with a "Edit link URL" action in Settings so it can be set after
  the fact without deleting the account (which cascade-deletes its
  mirrored recipes).

### Multi-angle review pass
Since I can't review the JS myself, I ran a structured multi-angle review
(line-by-line diff scan, removed-behavior audit, cross-file call-site tracing,
plus reuse/simplification/efficiency/altitude/conventions passes) against the
branch before marking it ready. It surfaced two real bugs, which are fixed:

- A mirrored recipe with no ingredients, no notes, and no link (possible if
  a sync run never resolved Mealie's `groupSlug`) fell through to the same
  "click the title to edit" shortcut used for detail-less *native* recipes,
  bypassing the read-only UI gating (the server still 403s the save, but the
  user hit a dead end instead of never seeing an edit affordance in the
  first place). Guarded on `source`; that specific edge case now renders as
  plain, non-interactive text instead.
- Editing an account's `external_url` only reached recipes the *next time
  they changed in Mealie itself* — the sync skips recomputing anything for
  unchanged recipes, `recipe_url` included, so the old (possibly
  Docker-internal, browser-unreachable) link stuck around indefinitely.
  Now stores the recipe's Mealie slug and rebuilds `recipe_url` from it on
  every sync pass, with no re-fetch needed and a guard against clobbering a
  good link if a given pass doesn't resolve `groupSlug`.

Also added while reviewing: recipe thumbnails, and the Mealie badge moved
from the expanded detail panel into the collapsed row itself (a mixed
native/mirrored list previously gave no way to tell them apart before
clicking each one). Mealie's media endpoint needs the same Bearer token as
everything else, so the browser can't hotlink it — added a
`GET /recipes/:id/mealie-thumbnail` proxy that fetches the bytes server-side
and streams them through, same MIME allowlist/security-header pattern as
the existing DMS thumbnail proxy.

### More mobile testing
A few more real issues turned up testing on an actual phone:

- **Mouse-wheel scroll only registered over the narrow reading column.**
  `.kitchen-list` (the actual `overflow-y: auto` scroller for pantry/shopping/
  recipes) also carried the 720px reading-column `max-width` itself, so on a
  wide desktop window its hit-testable box was only that narrow column —
  hovering the empty space to the right did nothing. Moved the `max-width`
  onto the scroller's children instead (`.kitchen-group` for grouped tabs,
  `.kitchen-rows` directly for recipes), so the visible card stays the same
  width but the scroller itself spans the full available width.
- **The Mealie badge landed in a different spot on every row on mobile,
  sometimes on a different line entirely.** Root cause: the recipe name is a
  flex item with the browser's default `flex-basis: auto`, which uses the
  *full unwrapped text width* to decide what fits on a line. A short title
  fit next to the thumbnail; a long one didn't, and instead of just
  wrapping its own text, the *entire name* moved to its own line — two
  different row shapes (3 lines vs 4) depending on nothing but title length,
  with the badge's position along for the ride. `flex-basis: 0` fixes that
  (the item always "fits", then grows, then its text wraps normally), and
  the badge now gets a forced, consistent line of its own the same way the
  ingredient count already does. Verified across four titles from 1 to 36
  characters — identical row shape and badge position every time.
- **The source filter row cost 121px of header on mobile vs. pantry's 65px**,
  for three options almost always sitting on "All". Replaced the chip row
  with the same treatment pantry already uses for "manage storage
  locations": one icon button in the header, opening the existing
  popover-menu component (the same one row overflow-actions use) instead of
  a bespoke always-visible row.

Also added generic-thumbnail handling for recipes without a Mealie image:
Mealie's media endpoint 404s rather than serving one itself
(mealie-recipes/mealie#4804), so a missing/deleted image now renders a
muted placeholder icon instead of nothing or a broken-image icon, and the
ingredient-count column got a reserved width so it (and the badge before
it) don't drift a few pixels per row depending on digit count or
singular/plural wording.

## Checklist

- [x] `npm test` passes

## Testing
Manually tested end-to-end against a real Mealie instance at home: account
setup, sync, recipe browsing/filtering, scheduling a mirrored recipe into
the meal plan, shopping-list transfer, and confirming edit/delete are
blocked on mirrored recipes. This was tested on a fresh instance of yuvomi as well pre-existing running instance. Since first opening the PR I've also
re-tested the scroll fix and the FAB fix against a locally seeded instance
with 50 mirrored recipes, confirmed the "Edit link URL" flow, and verified
the thumbnail proxy end-to-end against a local stand-in HTTP server standing
in for Mealie (real Bearer-token request, real byte response, rendered and
loaded in an actual browser) rather than only against mocked unit tests.

## Disclosure
I built this with heavy use of Claude (AI assistance) — I'm not a JS
developer, so I can't personally review the implementation line-by-line.
I did drive the design decisions (read-only mirror, scope boundaries),
tested it thoroughly against a real Mealie instance, and ran a security-
focused review pass before opening this. Posting it as a working
proof-of-concept for #530 and would appreciate a close look — happy to
iterate on anything that doesn't fit the project's conventions. Also, I am happy to provide screenshots or access to my test yuvomi and mealie instances if that is helpful from a UX evaluation aspect

## Finally
I will not be offended in anyway if you wish to keep LLM driven code out of this project. I did not see any provisions for this in the contributing or code of conduct markdown, but I can absolutely understand why other projects have taken this stance due to the sheer amount of load it places on the maintainers.


ulsklyc:
Merged and shipped in [v1.73.0](https://github.com/ulsklyc/yuvomi/releases/tag/v1.73.0). Closing manually because the ten commits were squashed and rebased onto main, so GitHub does not pick them up as merged. Authorship is preserved on the commit.

Thank you for this, and for the disclosure at the bottom of the description. To answer it directly: the code is judged on what it does, and this holds up. The read-only mirror was the right call on scope, the 403 sits on the server rather than only in the UI so the two cannot drift, the token never leaves the server and is stripped from API responses, the thumbnail proxy carries the same MIME allowlist and security headers as the DMS one, and the sync guards against a failed fetch pruning existing mirrors. The colours went into `tokens.css` with light and dark variants instead of being hardcoded at the call site, which is the detail most contributions miss. The self-review pass in the description caught two real bugs before I saw them.

What changed on the way in:

- **Migrations 111-113 became 118-120.** Those numbers were taken in main while this PR was open. This is also what made the branch conflict.
- **The FAB and kitchen-rows work came out.** Not because it was wrong, but because main had already landed both independently as #634, with stricter guards in `test-frontend-audit.js` than the PR carried. Your analysis of the FAB retract matched the one in main almost point for point, including the iOS address bar as the trigger. Same for `align-self: start` on `.kitchen-rows` and the `.kitchen-list` scroller width. That was the bulk of the remaining conflict.
- **`innerHTML` in the source filter** became `replaceChildren()` + `insertAdjacentHTML()`. It is a hard constraint here (there is a hook that blocks it), and every value was already `esc()`-wrapped so it was never an XSS issue, just the wrong primitive.
- **Manual `"` escaping** on the external-URL field became `esc()` from `utils/html.js`.

One thing I have not settled yet and want to flag, since you asked for a close look: `MealieAdapter` uses raw `fetch()` on the admin-supplied URL, which goes around `server/utils/ssrf.js`. That module exists specifically because three subsystems taking user-supplied URLs had drifted apart, and this is a fourth. It is admin-only, so it is not urgent. The awkward part is that the consistent fix, an opt-in like `DOCUMENT_STORAGE_WEBDAV_ALLOW_PRIVATE_NETWORK`, would block a Docker-internal Mealie by default, which is the normal case for this integration. That trade-off is mine to make, not something to hand back to you.

Documented in `docs/SPEC.md` (data model and Recipes UI), `docs/test-suites.md`, README and the landing page.

ulsklyc:
Correction to my merge note above: I got one of the four points wrong, and it
was the one that took credit away from you.

I wrote that the kitchen-rows work "came out" because main had already landed
it as #634. That is true for the FAB retract removal, which #641 deleted
`utils/fab-scroll.js` for a few hours before this merge. It is not true for
the CSS. #641 never touched `kitchen-row.css`; it only changed `router.js`,
`layout.css`, `pantry.css`, `shopping.css`, `sw.js`, `fab-scroll.js` and
`test-frontend-audit.js`.

    git log -S "align-self: start" -- public/styles/kitchen-row.css
    8e8ae8c7 Zak Lantz  feat(meals): Rezepte aus einer selbst gehosteten Mealie-Instanz spiegeln (#530)

Single hit, and it is your commit. Same for moving the reading-column
`max-width` off `.kitchen-list` onto its children so the scroller spans the
full width. Both shipped, both are yours, and they are load-bearing for every
tab that renders `.kitchen-list`, not just recipes. Your read on why a lone
tall group gets silently clipped was correct and is now the behaviour of the
shared component.

The claim also sits in the commit message of 8e8ae8c7, which I cannot rewrite
now that it is tagged. This comment is the correction of record.

One consequence worth naming: I said the guards in `test-frontend-audit.js`
were stricter than the ones the PR carried. That holds for the FAB
(lines 1975-1979 assert the file is gone and that `router.js` and `sw.js` no
longer reference it). There is no guard on `align-self: start` for
`.kitchen-rows` at all, so the reasoning you put in the CSS comment is the
only thing holding it. I am adding one.

PR #621, opened 2026-08-01
Feat/credit card invoice management

ContatoLucasSonntag:
## Summary

Discussions: https://github.com/ulsklyc/yuvomi/discussions/541

## Changes

image

## Checklist

- [ x] `npm run test:tasks-routes` passes
- [ x] Follows [CONTRIBUTING.md](../CONTRIBUTING.md) conventions
- [ x] No new frontend dependencies (vanilla JS, no frameworks)
- [x ] CHANGELOG.md updated (if user-facing change)


ulsklyc:
Thanks for putting this together, and for linking it to discussion #541 - credit-card invoices are a feature I do want in Yuvomi, and `credit` already exists as an account type, so this is the right place for it. The concept is welcome. This particular implementation isn't mergeable yet, and I'd rather be specific about why than leave you guessing.

**One blocker is a data-loss regression outside the feature.** In `POST /api/v1/budget`, the new insert passes literals where the existing recurrence values used to go:

```js
0, vRrule.value,
'monthly', 0, null, accountRef.value,
```

`isRecurring`, `interval`, `isVirtual`, `storeAmount` and `fullAmount` are still computed just above, but nothing reads them any more. The effect is not limited to installment purchases - it hits **every** new budget entry:

- a recurring entry (subscription, rent, standing cost) is stored with `is_recurring = 0` and never repeats
- a virtual yearly budget of 1200 is stored as a one-off 1200 instead of the smoothed 100/month, with `recurrence_full_amount = NULL`

Reproduced against this branch: `POST` with `is_recurring: true` comes back with `is_recurring = 0`, and a virtual yearly entry stores `amount = -1200, virtual = 0` where main stores `amount = -100, full = -1200`.

*(Correcting myself here: the first version of this comment claimed the test suite doesn't cover this. It does. `npm run test:budget-entries-routes` fails on this branch at "POST /: virtuelles Budget glättet den Jahresbetrag auf den Monatsanteil", 35 passed / 1 failed. The checklist in the description ticks `npm run test:tasks-routes`, which is the wrong suite for a budget change - a plain `npm test` would have surfaced this before review.)*

**`closing_day` doesn't do what the field promises.** The UI labels it "Abrechnungstag" / closing day, and #541 asks for it in the sense that gives credit cards their shape: a purchase made after the closing day lands on the *next* statement. `statementFor()` cuts strictly on calendar months:

```js
WHERE b.account_id = ? AND b.date >= ? AND b.date < ?   // `${month}-01` .. `${nextMonth}-01`
```

`closing_day` is only read to decide whether an *empty* statement auto-closes. So a card with closing day 15 still bills the 1st to the 31st. That's a design decision, not a typo - either the field carries the billing period (and `statementFor()`, the history route and the installment dates all follow from it), or the feature is calendar-month-based and the field should be named and documented as "when the statement closes", not as the period boundary. I'd want that settled before reviewing the rest, because it changes the shape of most of this code.

**A `GET` that writes.** `statementFor()` inserts into `budget_account_invoices` when it runs, so `GET /accounts/invoices` and `GET /accounts/:id/invoices/history` mutate state. GET is not covered by the CSRF middleware in this app, and a route whose response depends on how often it was called is hard to reason about. Auto-closing belongs in the `POST` action or in a scheduled pass.

**Project conventions the diff doesn't follow** - none individually hard, together a lot of rework:

- *UI text.* All user-facing strings go through `t('key')` (CLAUDE.md, hard constraint). The PR adds 29 keys, then hardcodes roughly fifteen more strings in Portuguese and English: "Relatório de faturas" (button label, title and modal heading), "Pagar fatura", "Excluir/Alterar compra parcelada", "Reabrir fatura", "Todas as parcelas", "Somente esta parcela", "Esta e as próximas parcelas", "Valor do pagamento", "Conta de débito", "Pagamentos", "Estornar", "Estornar último pagamento", "Pagamento realizado", plus `aria-label="Invoice cycle"`. Each needs a key in all 23 `public/locales/*.json`.
- *German values.* Several of the added keys read as machine translation and would ship as-is: "Fakturtotal", "Fakturastatus", "Zahlte Raten", "Schlusselungstag" (not a German word - the server returns it in validation errors), `invoiceInstallmentsShort: "Rechnungen"` where the column means Raten, `invoicePaidLabel: "Bezahlte Rechnungen"` for what is a paid amount. `de` is the reference locale here, so these want a native pass.
- *Colours.* `budget.css` hardcodes twelve values (`#b45309`, `#2563eb`, `#fef3c7`, …). Every colour, radius, shadow and font size comes from `public/styles/tokens.css` - that's what makes dark mode work; `#fef3c7` as a background is unreadable there.
- *Breakpoint.* `budget.css:581` uses `560px`; `test:frontend-audit` fails on it (canonical: 640/768/1024/1440).
- *Comments.* Four code comments are in Portuguese (`entries.js` around the installment update and the delete cascade, `budget.js` at the invoice section). The codebase comments in German.
- *Migrations.* 114-120 collided with numbers already taken in main. Also: 116, 118 and 120 are a correction chain on the same table (create, then add `paid_amount`, then rebuild for a `CHECK`) - since none of this has shipped, that's one migration.
- *API surface.* Six new routes, none in `server/openapi.js`, and `test-budget-structure.js`'s route table wasn't updated (it fails).

**Performance.** `/accounts/:id/invoices/history` calls `statementFor()` per month, and each call runs 3-4 queries plus loads every entry of that month. Three years of history is ~140 queries and the full entry set in one response.

**Where it stands.** I have a local branch rebased onto current main (`v1.73.0`) with the migrations renumbered to 121-124 and the correction chain folded into one, the recurrence regression fixed, and the route-table guard updated - the budget and schema suites pass on it. Say the word and I'll push it so you can pull it into this PR; I did not want to force anything onto your branch uninvited. The remaining items are yours, particularly the `closing_day` question, which I'd rather you decide than have me guess at.

If you'd like to keep going: I'd suggest splitting it. Credit-card metadata on the account (bank, limit, closing/due day) plus the available-limit display is a small, self-contained PR I could take quickly. Statements and installment purchases are the part that needs the billing-period question answered first.

Leaving the PR open for now rather than closing it, since the feature is wanted and the question above is answerable.

ContatoLucasSonntag:
Hi! First of all, thank you for taking the time to review this in such detail. I really appreciate the thorough feedback, and it's clear you put a lot of effort into testing the implementation.

You're absolutely right about the recurrence regression. That was an unintended mistake during the refactor, and I completely missed that the computed values were no longer being used in the insert. Thanks for catching it, and also for correcting the note about the test suite.

Regarding `closing_day`, I agree this is the biggest architectural question in the PR. My original intention was to introduce the account metadata first and then build the statement system incrementally, but I can see that the current implementation leaves the field implying behavior that isn't actually implemented. Before moving forward with the invoice logic, I'll rethink that design so the data model and billing period semantics are consistent.

The point about GET routes mutating state also makes sense. I hadn't considered the implications for CSRF and cacheability, and I agree that auto-closing should happen through an explicit action or background process instead of during a read request.

The remaining issues (translations, German localization, design tokens, breakpoint conventions, comment language, OpenAPI updates, route table, migration cleanup, and performance) are all valid as well. I'll go through them one by one so the implementation matches the project's standards.

I also really appreciate your offer to push your local branch. Since you've already fixed the recurrence regression, renumbered the migrations, and updated the route table, I'd be happy to use that as the new base instead of duplicating work. Please go ahead and push it—I think that will save both of us time and let me focus on the remaining design and implementation changes.

I also like your suggestion about splitting the work. It makes the review much more manageable. I'll rework this into smaller PRs, starting with the credit-card account metadata and available-limit functionality, and then tackle statements and installment purchases separately once the `closing_day` behavior is properly defined.

Thanks again for the detailed review and for leaving the PR open. I appreciate the guidance, and I'll work through the remaining issues before requesting another review.

ulsklyc:
Thanks for the reply, and for taking the review the way it was meant.

**First, a correction I owe you.** Two of the German values I quoted are not in your PR. I wrote "Zahlte Raten" and "Fakturastatus"; the branch actually has `"Bezahlte Raten"` and `"Rechnungsstatus"`, and both are correct German. I was sloppy quoting from a scratch list instead of the file. The point stands for the ones that are really there (`invoiceTotalLabel: "Fakturtotal"`, `invoicePaidLabel: "Bezahlte Rechnungen"` for what is an amount, `invoiceInstallmentsShort: "Rechnungen"` where the column means Raten, and `Schlusselungstag` in the server-side validation errors), but it was five items, not seven, and I should have checked before writing that paragraph.

**On splitting, and what I pushed.** You said yes to both taking my branch as a base and splitting into smaller PRs, and those two pull against each other: if the `closing_day` design is going to change, most of what is in my rebased branch (`statementFor()`, the history route, the installment dates) changes with it, and the first small PR you named needs almost none of it. So rather than hand you 2000 lines to carve up, I pushed only the part that is genuinely independent:

[`feat/credit-card-metadata`](https://github.com/ulsklyc/yuvomi/tree/feat/credit-card-metadata) - one migration, ~300 lines, full suite green.

It has the issuing bank and the credit limit on the account, the available-limit figure in the account list (limit minus what is owed; a balance in your favour does not raise it above the limit), the fields shown in the account dialog only for credit-card accounts and cleared when the type changes away, all strings through `t()` in all 23 locales with a native German pass, tokens instead of hardcoded colours, and three test cases. I kept your authorship on the commit. Take it as-is, or cherry-pick from it, whichever suits you.

One deliberate difference from what I suggested earlier: I left `closing_day` and `due_day` out of it. The columns themselves would survive any answer to the question below, but a field labelled "Abrechnungstag" that does not bound anything is exactly the thing I flagged, and I did not want to ship it just to make the PR look complete. They can come as their own migration with the logic that gives them meaning.

**The question I still need from you.** Two coherent answers, and I would like you to pick one rather than have me guess:

*(A) The billing period is the real one.* A purchase after the closing day belongs to the next statement, so the period runs from the day after last month's closing day through this month's. This is how physical cards work and how #541 reads. It means `statementFor()` takes a date range instead of a month, the history route iterates periods instead of calendar months, and every statement needs a label answering "which month is 16 Jan to 15 Feb?".

*(B) Statements are calendar months.* `closing_day` only decides when the month's statement is frozen. Simpler, and most of your existing code already does this - but then the field needs a name and a description that say so, in the UI and in the docs.

I lean towards (A), because (B) does not match any card a household actually holds, and someone who fills in "15" will expect (A). But it is more work, and it is your PR, so it is your call. Either way the answer settles the shape of the statement code, which is why I would rather have it before reviewing that part line by line.

**One more finding for whenever you get to the statement code.** Paying an invoice writes a credit entry onto the card account dated the day of payment. Pay January's statement on 10 February and that credit lands inside February's statement, which sums every entry of the calendar month, so February comes out lower by the amount you just paid. The counter-entry is right for the account balance and wrong for the statement total; the two need to be told apart. Worth designing for rather than patching later.

My rebased branch with the statement work is still sitting here. Once (A) or (B) is settled I will either push it too or throw away the parts that the answer invalidates, whichever is less work for you. No rush from my side.

ulsklyc:
I left you with a design question, and that was the wrong call on my part. The billing period is a decision about how the budget module works, not about how your contribution is written, and you should not have to guess which answer I would accept. So here it is, decided:

**It is (A). A purchase after the closing day belongs to the next statement.**

The reason is the one I already gave, and thinking about it since has not moved me off it: (B) does not describe any card a household actually holds. Someone who types "15" into a field labelled "Abrechnungstag" expects the period to run from the 16th to the 15th. Shipping (B) would mean the field is honest only in the docs, and the first support question would be why the statement covers the calendar month.

What follows from that, so you are not deriving it from a one-line answer:

- `statementFor()` takes a date range instead of a month. The period ends on this month's closing day and starts on the day after the previous one, with the usual short-month clamp (closing day 31 in February is the 28th or 29th).
- The history route iterates periods, not calendar months.
- Every statement needs a label. "16 Jan to 15 Feb" has to be called something in the UI, and I would name it after the month it closes in, with the actual range next to it.
- Installment dates follow the period, not the month.

**Two things that are now easier than they were on 2 August.**

First, the account metadata part is merged: #684 puts `credit_bank` and `credit_limit` on the account with the available-limit figure in the list, your commit with your authorship, tests green. That is the split we agreed on, and it means the statement work no longer has to carry it. I left `closing_day` and `due_day` out on purpose - they belong with the logic above, as their own migration.

Second, a warning about rebasing: main has moved to v1.85.0 and the migration numbers have shifted again. 121 to 124 are taken now, the next free one is 132 (131 goes to #684). I would not spend an evening rebasing the statement code onto that, because most of the conflicts sit in exactly the lines that (A) rewrites. Starting `statementFor()` from the period definition is likely less work than reconciling the month-based version with main and then converting it.

**And the finding from my last comment, which (A) does not solve by itself.** Paying a statement writes a credit onto the card account dated the day of payment, so paying January's bill on 10 February lands inside February's period and lowers that statement by the amount you just paid. The counter-entry is right for the account balance and wrong for the statement total. Worth deciding up front how the two are told apart - a flag on the entry, or a separate table for payments - rather than patching it once the numbers look odd.

No deadline from my side, and the draft status is fine. I would just rather you spend the time on code than on guessing what I want.

ulsklyc:
Hi Lucas, I saw you closed this one, and the obvious thing first: that part was on me, not on you.

I parked the billing period question with you on August 2nd and only decided it on the 5th, and by then the branch had been sitting in draft against a conflicting base for days. It was a product decision about how the budget module works, and it was never yours to guess. Asking a contributor to pick the answer a maintainer would accept is a reliable way to stall a branch, and that is what happened here.

So, with no pressure attached and either answer genuinely fine: **do you want to pick this up again?**

If you do, the guessing part is gone. Decision (A) stands: a purchase after the closing day belongs to the next statement, and the period runs from the day after the last closing day to the current one. Concretely that means `statementFor()` takes a date range instead of a month, the history route iterates periods, each statement needs a label for "January 16 to February 15", and the payment of a statement must not land in the total of the following period. That last one matters: the credit is correct for the account balance and wrong for the statement total, so the two have to stay distinguishable.

Two practical notes if you start again. Branch fresh off current `main` instead of rebasing the old one: main is at v2.6.1 now and the next free migration number is 136, and the repeated collisions (114 to 120, then 121 to 124) are a good part of why this branch kept going stale. The account metadata from #684 is already in, so the statement work sits on top of it rather than next to it.

If you would rather not, that is completely fine and I will build it myself. The design, the field layout and the installment model in this branch are yours, and I would say so in the changelog either way. I would just rather ask than assume.

Thanks for everything you have sent so far, several of which have shipped, and sorry for the part of this that was mine.

PR #622, opened 2026-08-01
fix(layout): preserve sidebar scroll position

ContatoLucasSonntag:
## Changes

- Fixed: the sidebar no longer forces scrolling back to the active item.
- The sidebar now preserves the user's scroll position when selecting an item, preventing it from jumping between the first and last items after reaching either end of the list.

### DIFF
--- a/public/router.js
+++ b/public/router.js
@@ -3054,10 +3054,20 @@ function rebuildNavigation({ updateLabels = true } = {}) {
   }
 
   if (navSidebarItems) {
+    // replaceChildren recria toda a árvore da navegação (por exemplo, após
+    // mudança de rota, idioma ou módulos) e o browser zera a rolagem do
+    // container. Preservamos a posição manual antes e depois do re-render.
+    const previousScrollTop = navSidebarItems.scrollTop;
     const sidebarEls = sidebarNavItems();
     navSidebarItems.replaceChildren(...sidebarEls);
     if (window.lucide) window.lucide.createIcons({ el: navSidebarItems });
-    requestAnimationFrame(() => positionSidebarIndicator());
+    requestAnimationFrame(() => {
+      navSidebarItems.scrollTop = Math.min(
+        previousScrollTop,
+        Math.max(0, navSidebarItems.scrollHeight - navSidebarItems.clientHeight),
+      );
+      positionSidebarIndicator();
+    });
   }
   if (bottomItems) {
     const moreBtn = bottomItems.querySelector('#more-btn') ?? moreNavButtonEl();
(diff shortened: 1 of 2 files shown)


ulsklyc:
Merged and shipped in [v1.73.0](https://github.com/ulsklyc/yuvomi/releases/tag/v1.73.0). Closing manually because the branch was rebased onto main, so GitHub does not pick the commits up as merged.

Thanks for this. The diagnosis was right and it is a real bug: `replaceChildren()` resets the list's scroll position on every rebuild, and rebuilds happen on each route change as well as on language and module-list changes. Preserving the position across the rebuild is exactly the fix.

Two adjustments on the way in:

- I kept the auto-scroll to the active item. It was added deliberately (audit finding F-01): with a long module list, the active item and its indicator pill could both sit below the fold, so the navigation lost its "you are here" anchor. Its explanatory comment was still in place above the removed block. With your scroll-preservation in front of it, it no longer fires for an item that is already visible, which is what made it feel like it was fighting the user. So both behaviours coexist now, and the jumping is gone.
- Dropped `height: 100dvh` / `max-height: 100dvh` on the sidebar (it is `position: fixed` with `top: 0; bottom: 0`, so the height is already determined) and `-webkit-overflow-scrolling: touch` (iOS < 13). `min-height: 0` and `overscroll-behavior-y: contain` stayed, and I added a note on why `min-height: 0` is load-bearing here.

One note for future PRs: code comments in this repo are in German. Yours was in Portuguese, so I rewrote it in place.

ContatoLucasSonntag:
Thanks for looking into this and for the update.

I tested the latest version, but unfortunately the issue is still present on my side.

The change appears to fix one code path, but there are still cases where the original behavior can be reproduced.

ulsklyc:
Thanks for testing it again.

Two things so I can chase the right code path instead of guessing:

**1. Which version is running on your side?** The Changelog panel in the navigation shows it. The fix shipped in 1.73.0, and self-hosted setups sometimes still serve the previous image after a pull.

**2. Repro details.** The fix covers the navigation rebuild path (route change, language change, module list change). There are other call sites that reposition the sidebar without going through it, so the exact trigger decides where to look:

- What happened right before the jump: switching route, changing language, changing the module list, or reloading the app?
- Where was the sidebar scrolled before that: top, middle, bottom?
- Where does it jump to: back to the top, to the bottom, or to the active item?
- Browser, OS, and window height? The sidebar only scrolls when the module list is taller than the viewport, so it may not reproduce at all in a tall window.

A screen recording works too if that is easier than writing it out.

PR #623, opened 2026-08-01
feat(tasks): add default expanded subtasks setting

ContatoLucasSonntag:
## Summary

Adds a new setting that allows users to choose whether subtasks should be expanded by default in the Tasks page.

## Changes

- Add a new Tasks setting for default subtask visibility.
- Allow subtasks to be expanded by default.
- Preserve manual expand/collapse behavior.


ulsklyc:
Merged and shipped in [v1.73.0](https://github.com/ulsklyc/yuvomi/releases/tag/v1.73.0). Closing manually because the branch was rebased onto main, so GitHub does not pick the commits up as merged.

Nicely done. It slots into the existing module-options leaf without inventing a new pattern, reuses the `expandedSubtasks` option that `renderTaskCard` already had, and the locale files came back at exactly +4 lines each with no reserialisation, which is more care than that detail usually gets.

One change on the way in: `api.get('/preferences')` sat in the shared `Promise.all` without its own fallback, so a failure there would have dropped the whole page into the load-error branch, leaving tasks, users and categories empty even though their own requests had succeeded. It is a display preference, so it now carries `.catch(() => ({ data: {} }))` like the calendar and dashboard loaders do. Also pulled the two "three toggles" counting comments in the leaf and its guard up to date.

Documented in `docs/SPEC.md` under Tasks.

Issue #618, opened 2026-07-31
[Bug] CalDAV event sync for non-admins is not working


BVV2:
## Description

None of the family members can set the calendar events to a configured CalDAV sink, except the admin.

## Steps to Reproduce

1. Set up a CalDAV connection.
2. Go to the Calendar.
3. Open an event menu.
4. Expand "More settings" to see a "Sync target" dropdown.
5. Find and select CalDAV connection from earler.
6. Save event, check if the CalDAV server got it successfully.
7. Log into a non-admin account.
8. Go to the Calendar.
9. Open an event menu.
10. Expand "More settings" to see a "Sync target" dropdown.
11. Only "Store locally" is present.

## Expected Behavior

Expected the CalDAV connection to be available for all users, or an admin setting to grant the user the ability to use it.

## Actual Behavior

The CalDAV connection is not there, only "Store locally" is present.

## Environment

- **Browser:** Version 149.0.7827.114 (Official Build, ungoogled-chromium) (64-bit) (Other browsers seem to have same results)
- **Device:** Desktop
- **OS:** Ubuntu 24.04 
- **Server** TrueNAS 25.10.4 - Goldeye
- **Yuvomi version/commit:** App Version:0.96.8 Version:1.0.17
- **CalDAV version:** Locally hosted Radicale v3.7.5

## Logs

No logs related to the problem found aside from normal flood.

ulsklyc:
Fixed in **v1.66.1**.

You were right that this was not intended behaviour, and it turned out to be narrower than a missing permission: the "Sync target" dropdown was reading the calendar list through the *account management* routes (`/caldav/accounts`, `/google/calendars`), which are admin-only. A non-admin got a 403, the frontend swallowed it silently, and "Store locally" was all that was left.

Saving a target was never restricted. `POST`/`PUT /api/v1/calendar` has always accepted `target_caldav_account_id` from any member, so the restriction was an accident of the read path rather than a permission boundary. The same applied to Google calendars, which just was not reported.

The dropdown is now served by a dedicated read route, `GET /api/v1/calendar/sync-targets`, available to every authenticated user. It returns only the display name and target key of the calendars an admin has enabled (and, for Google, that are writable) - no credentials, server URLs, or usernames. Account management stays admin-only, so no admin setting is needed: any family member can now pick a target your admin has switched on.

Two things improved along the way: it is a single request instead of one per CalDAV account, and an expired Google token no longer empties the CalDAV group in the same dropdown.

[...]

Issue #593, opened 2026-07-28
[Bug] Yuvomi doesn't sync calendar deletion


aserl-cell:
## Description

I've been testing the calendar syncing function, and noticed that if I create an event on Yuvomi, after it syced with the Google calendar, it won't delete the event from the google calendar when I deleted the event from Yuvomi. Just want to also note that if the event was deleted from Google calendar, it does sync through. Similarly, changes  in event details made from Google's end will sync into Yuvomi. 

## Steps to Reproduce

1. Go to Calendar
2. Create an event, and wait to sync to Google calendar. 
3. Delete event from Yuvomi. 

## Expected Behavior

Event is deleted from Google calendar as well. 

## Environment

- **Browser:** Chrome 150
- **Device:** Android phone
- **Yuvomi version/commit:** 1.48.2

ulsklyc:
Fixed in **v1.51.0**. Thanks for the precise report, and especially for noting which direction still worked, that pointed straight at the cause.

Outbound sync only ever *created* events in Google. Once an event had been pushed, it was flipped to a "comes from Google" state, and the outbound branch only ever looked at purely local events, so from that moment nothing you did to it in Yuvomi reached Google again. Your deletion was one symptom; editing an event had the same silent no-op, which is why the inbound direction you tested kept working perfectly.

What now happens when you change a synced event in Yuvomi:

- **Delete** removes it from Google as well.
- **Edit** is sent along: title, notes, location, colour, all-day, start and end time, and the repeat rule. Assignment, visibility, icon and attachments stay in Yuvomi, since Google has no equivalent for them.
- **Moving it to a different Google calendar** moves it there instead of leaving it behind. This needs write access to both calendars.

The change is tried immediately when you save and retried by the next sync run if Google cannot be reached, so a short outage does not swallow it. An incoming sync no longer overwrites a local edit that has not made it to Google yet.

One related bug turned up while fixing this, which may also have hit you: moving an event between two synced calendars *inside Google* could make it disappear from Yuvomi. The source calendar reports the event as cancelled while the destination still lists it under the same ID, and the deletion was applied by ID alone, so it could remove the entry the destination had just written. That is fixed too.

[...]

PR #676, opened 2026-08-05
Feat/outlook.com sync & graph push

cberane:
## Summary

Implements the proposal from discussion #675.

Outlook.com no longer speaks CalDAV, so the only write path is the Microsoft
Graph API. This adds a new calendar provider for personal Microsoft accounts
(outlook.com / M365 Family) — deliberately one-way: Yuvomi stays the source of
truth, and changes made in Outlook are overwritten on the next run.

## Changes

- New Outlook calendar provider (Microsoft Graph, personal accounts): one-way
  push, multi-account OAuth against `/consumers`
  (`MS_CLIENT_ID`/`MS_CLIENT_SECRET`/`MS_REDIRECT_URI`), plain `fetch`, no SDK,
  no new dependencies; reconnect replaces tokens (upsert over `ms_user_id`),
  `invalid_grant` sets `needs_reauth` instead of failing silently
- Auto-sync per account: a dedicated target calendar + owner; all events
  visible to that member are pushed automatically, assignees appended to the
  title (`Dinner (Anna, Ben)`, convention #482); externally synced events
  (Google/CalDAV/ICS) are excluded to avoid duplicates; explicit per-event
  targets (`target_outlook_*`) win on collision
- Drift detection: each run lists `id`+`changeKey` once per linked calendar —
  events edited in Outlook are PATCHed back to the Yuvomi state, events deleted
  there are recreated; events created directly in Outlook are left untouched
- Push state lives in `outlook_event_links` per (event, account) with content
  hash (unchanged = 0 requests), tombstone semantics and the Graph ETag — no
  handoff to `external_source='outlook'`, the #593 queue stays untouched;
  migration v134
- UI: Outlook card under Settings → Calendar sync (connect/reconnect, auto-sync
  target calendar, account owner, calendar checkboxes, sync-now, delete);
  Outlook groups per account in the event dialog's sync-target picker;
  `outlook:<accountId>|<calendarId>` as third target format, so an Outlook
  calendar is also selectable as a member's default sync target (#620);
  15 new i18n keys in all 24 locales
- Admin-only management routes (parity with `caldav_accounts`); members pick
  targets via `/calendar/sync-targets` (#618); empty runs stay below the
  default log level (#601), missing `MS_*` config fails loudly like the Google
  provider; `MS_*` deliberately excluded from the installer (the Entra app is
  created only after installation)
- Docs: step-by-step Entra app registration for private individuals
  (`docs/installation.md`), SPEC tables and provider section, GDPR third-country
  entry for Microsoft Graph, `.env.example`, README calendar line, CHANGELOG
- Tests: `test:outlook-calendar` (31) and `test:outlook-event-target` (6),
  wired into `npm test`; routes guard 46→55

## Checklist

- [x] `npm test` passes


ulsklyc:
## Review: Outlook.com push via Microsoft Graph

First: I owe you an apology. You asked in #675 whether the proposal was fine before opening a PR, and I never answered. You then built the whole thing anyway and have been waiting since 7 August. That is on me, not on you, and it is the worst way to treat 3000 lines of careful work.

**The answer is yes.** Outlook.com dropping CalDAV leaves a real hole that ICS subscriptions do not fill, and I want this in Yuvomi.

I have gone through the branch and run the suites locally. The craft is genuinely good:

- The test chain is wired in all three places (`scripts`, the `test` chain, `docs/test-suites.md`). Most contributions miss at least one.
- 33 + 6 tests pass, and they test the things that actually break: RRULE to `pattern`/`range`, all-day conversion, content-hash no-ops, `changeKey` drift in both directions, tombstone deletes, visibility loss, `invalid_grant` to `needs_reauth`.
- All 23 locales with real translations, correct indentation, no plural violations.
- SPEC, both READMEs, `installation.md`, `PRIVACY-FOR-SELFHOSTERS.md` and `SECURITY.md` all updated.
- No new dependency, no SDK, no `innerHTML`, and you followed the CalDAV multi-account shape instead of inventing a third pattern next to Google. That was the right call.

Here is what needs to happen before it can go in.

### Blocking

**1. The migration number is taken.** Your `outlook_accounts` / `outlook_calendar_selection` / `outlook_event_links` block plus the two `ALTER TABLE calendar_events` sits at **v134**. `main` is at **v145** now. It needs to move to **v146**. Migrations in this repo are append-only and never renumbered after the fact, so this has to be the last entry at the time you push.

**2. `main` has moved 430 commits since you branched.** The good news: only five files conflict, and four of them are mechanical (`CHANGELOG.md`, `package.json`, `docs/install.html`, `test/test-installer-schema.js`). All 23 locale files merge cleanly, which says something about how carefully you placed those keys.

Please use a **merge**, not a rebase. For the `CHANGELOG.md` conflict, resolve by meaning: your lines belong under `[Unreleased]`, and the released blocks keep only what actually shipped.

If you would rather I take the merge over and push it into your branch, say the word. I did that for another contributor recently and it is no trouble.

### Should fix

**3. The `MS_*` variables belong in `env-schema.js`, not in the exception list.**

The PR leaves `tools/installer/env-schema.js` untouched and instead adds the three keys to `INTENTIONALLY_NOT_IN_INSTALLER` in `test/test-installer-schema.js`, with the reasoning that the Entra app registration only exists after installation.

That reasoning applies word for word to `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` / `GOOGLE_REDIRECT_URI`, and there we decided the other way: they sit in the schema as `type: 'user'`, `required: false`, `group: 'google'`, and the wizard offers them as optional fields. Google Drive does the same. `env-schema.js` is the source of truth for env vars in this repo, and `.env.example` is the annotated reference, so documenting a variable only in the latter leaves the schema behind.

Please add a `group: 'outlook'` block to `env-schema.js` mirroring the Google one, and drop the three entries from the exception list. As it stands the guard is green because it was switched off.

**4. About 113 en dashes in the locale values.** For example: `"Yuvomi bleibt führend – Änderungen in Outlook werden überschrieben."` The existing files carry two to four of those in total each, so this roughly doubles them in one go. House style is a plain `-`. No test catches this, which is why I am flagging it by hand.

### One product question, and it is mine to answer, not yours

The drift reconciliation (detecting events edited or deleted in Outlook via `changeKey` and reasserting the Yuvomi state) is the consistent reading of "one-way, Yuvomi is the source of truth", and your implementation is careful about it. But it is also the first place where Yuvomi actively overwrites something a user typed somewhere else, and a user who edits a time in Outlook and finds it reverted 15 minutes later will read that as a bug.

I want to keep the behaviour. What I would like added is the warning where the decision is actually made: `outlookPushHint` says the right thing, but it lives in the sync target picker. The same sentence should appear in the connect flow, **before** an account is linked.

### Not your problem

Two things I checked and decided against holding you to, so you do not fix them by accident:

- `graphRequest` has no fetch timeout. Correct, but neither does `google-calendar.js` nor `caldav-client.js`. That is a separate, repo-wide issue.
- Tokens are stored in plain text. Same as `caldav_accounts.password` and the Google refresh token in `sync_config`. There is no encryption anywhere in `server/` today.

### One practical note

Your CI has never run. Fork PRs from first-time contributors sit at `action_required` until a maintainer approves the workflow run, which is why the PR shows "no checks reported". That is on my side and it applies to every push, not just the first. Ping me here when you have pushed and I will release the run so you are not waiting on a check nobody triggered.

Thanks again for the patience, and sorry it took this long.

cberane:
> If you would rather I take the merge over and push it into your branch, say the word. I did that for another contributor recently and it is no trouble.

Hopefully I have the time in the next days to fix it myself. If not, I will let you know and would be thankful, for you to fix the blocking issues of the PR.

cberane:
@ulsklyc I am currently working on the merge and I hope the work will be done this evening :-)

# Blockers
1. I will bump the migration version number
2. I am currently merging the files
# Should Fix
3. I will fix the `MS_*` env declarations
4. I found multiple `en dashes` and `em dashes` throughout the project - not only locales. They existed before. I will fix mine anyway, but the issue wasn't obvious. To make the "house style" more obvious, I recommend to mention it in the contribution guidelines.
# CI
When I am ready, I will ping you.

cberane:
@ulsklyc Done and pushed — this is the ping for the CI run. :-)

Following up on my list from above:

**1. Migration:** by the time I merged, main's schema head had moved on to **v153**, so the Outlook block is now **v154** rather than the v146 from the review. It's the last entry in the array, and I re-checked against a fresh fetch of main immediately before pushing.

**2. Merge:** done as a merge commit. main had moved further here too — 560 commits and 9 conflicting files by merge time — but the resolution followed your guidance: CHANGELOG by meaning with my lines under `[Unreleased]`, and the locales all merged cleanly. The route-table guard is at 57 now (48 from main + 9 Outlook routes), and the full test chain passes after the merge.

**3. `MS_*` env declarations:** moved into `env-schema.js` as `group: 'outlook'` mirroring the Google block (`type: 'user'`, `required: false`, secret masked), exception entries removed. Heads-up: the schema guards pulled in more than the schema file — `install.html` now has an Outlook card in the calendar step (redirect URI derived from the planned origin, like Google's), `install.sh` asks for the credentials, the Portainer compose and Unraid template pass them through, and all 24 web + 24 CLI installer locales carry the new keys. I also added an "Outlook installer wiring" test next to the Google Drive one so none of this can drift.

**4. Dashes:** all 137 en dashes from my branch replaced with `-` (the 113 in the review was lines, not characters), plus one in a test-file header. `outlookOwnerNone` now follows the existing `"- Nobody -"` convention. As mentioned, the pre-existing ones on main are untouched — I still think a line in the contribution guidelines would make the house style visible, and I'm happy to do the repo-wide cleanup as a separate PR if you want it.

**One correction on the drift warning:** `outlookPushHint` was already in the connect flow — `buildOutlookProvider()` in `sync-calendar.js` emits it before the `configured` check, so it's visible before any account is linked. What didn't have it was the sync-target pickers. So I added the same sentence (same locale key, no new translations) there instead: under the target dropdown in the event modal and under the default-target setting, shown only while an Outlook target is selected. If you'd rather keep it to the connect flow only, say the word and I'll drop that commit.

ulsklyc:
Merged as `0b26eec4`. Thank you for seeing this through, and for doing it this thoroughly after the wait I put you through.

I went over the branch again before merging rather than trusting the green tick, and everything you listed checks out:

- **Migration** is v154 and last in the array, main was at v153.
- **The merge** landed on the current main head, so nothing of main got reverted underneath it. The place I actually worried about was the `test` chain in `package.json`: a conflict in that one enormous line would have swallowed suites silently. I split it into links and diffed it sorted. 221 to 223, nothing lost, both Outlook suites wired in.
- **`git diff` between your head and the merge commit is empty**, so what is on main is exactly what I reviewed.

Two things you did that were more than what I asked for:

**The installer guard.** I asked you to move `MS_*` into `env-schema.js`. You turned the exception inside out: `OUTLOOK_KEYS` now sits next to `GOOGLE_DRIVE_KEYS` as a positive expectation, so the test demands the keys are in the schema instead of excusing their absence. That is the right fix. The original entry in `INTENTIONALLY_NOT_IN_INSTALLER` was the interesting part of the whole review, because the guard was green precisely because it had been switched off, and the same reasoning would have excused `GOOGLE_CLIENT_ID` too. Following the schema through to `install.html`, `install.sh`, the Portainer compose, the Unraid template and all 48 installer locales, plus a wiring test so it cannot drift, was yours, not mine.

**Your correction on `outlookPushHint` is right and I was wrong.** `buildOutlookProvider()` does emit it before the `configured` check, so the connect flow already had it. The pickers were the gap. Keep both commits: someone who inherits an already-connected instance never walks through the connect flow, and the sync-target dropdown is exactly where they decide to point an event at Outlook. That is the moment the warning has to be readable.

**On the dashes, you are right and the fault is mine.** The rule is not written down anywhere a contributor can find it. It lives in a guard that only covers `README.md` and `README.de.md`, and main carries 137 in code comments and 364 in the CHANGELOG. You had no way of knowing. I will add it to CONTRIBUTING.md.

I would rather not take you up on the repo-wide cleanup, though. The CHANGELOG entries are release history and I do not want to rewrite them after the fact, and touching several hundred code comments would produce a diff that is noise for everyone reading `git blame` later. The rule earns its keep in user-facing text, which is what the guard already covers. Documenting it is the fix, not a sweep.

It is already out: **v2.26.0**, https://github.com/ulsklyc/yuvomi/releases/tag/v2.26.0. A minor, because it adds nine public routes under `/api/v1/calendar/outlook/`. The one thing I changed on the way was the three em dashes in your CHANGELOG entry, since that text ships as the release notes and into the Umbrel store listing, which is exactly the user-facing text the rule is meant for. The store notes I wrote by hand, as always, and they lead with the one-way warning and the advice to point it at a fresh, empty calendar rather than one someone has been keeping by hand.

And I have closed #675 as resolved. Ten days of silence on that thread was not acceptable, and the answer should have come before you wrote three thousand lines, not after.

ulsklyc:
Done: `73ecd6e0`. The rule is in CONTRIBUTING.md now, in both places it was missing: under Code Conventions as the rule itself, and in the Changelog section with the reason attached, since that is where getting it wrong actually costs something.

Your finding is in the text explicitly. It says that existing code comments on main are inconsistent and are deliberately left that way, so `git grep` is not a reliable guide here. That sentence exists because grepping is exactly what led you to conclude there was no rule, which was a reasonable conclusion to draw. Two em dashes in CONTRIBUTING.md itself went with the change, since the rule would otherwise have been sitting right next to its own violation.

PR #686, opened 2026-08-06
feat(meals): generalize Mealie recipe mirror into provider adapters, add Tandoor (#530)

### LINKED ISSUE #530: [Feature] Browse and select recipes from Mealie or Tandoor

manofcolombia:
## Summary

PR #614 shipped a read-only recipe mirror for Mealie only, and explicitly left
`server/services/mealie/` set up to become an adapter pattern once a second
provider showed up. #530 asked for exactly that generalization plus Tandoor
support. This PR does both: generalizes the schema/routes/frontend to a
provider-neutral adapter pattern (mirroring the existing Paperless/Papra
pattern in Document Storage) and adds `TandoorAdapter` as the second provider.

Verified live against real Mealie and Tandoor test instances (sync, 403
write-guard, thumbnail proxy, ingredient transfer, cascade delete).

Closes #530
Refs #614

## Changes

- **Schema**: migration v132, rename-only (`ALTER TABLE ... RENAME TO/COLUMN`,
  no data movement) — `mealie_accounts` → `recipe_provider_accounts` (+
  `provider` column, `CHECK (provider IN ('mealie','tandoor'))`),
  `recipes.mealie_*` → `recipes.provider_*`. FK cascades, unique index, and
  updated_at trigger carry over automatically.
- **Adapters**: `server/services/recipe-providers/index.js` factory +
  `mealie.js`/`tandoor.js`, one shared interface (`testConnection`,
  `listRecipeSummaries`, `getRecipe`, `recipeUrl`, `fetchThumbnail`).
  `TandoorAdapter` built against Tandoor's real DRF API (Bearer auth,
  `/api/recipe/`), not guessed from Mealie's shape.
- **Routes**: `/mealie` → `/recipe-providers`, provider-aware account
  CRUD/test/sync, unchanged auth/scoping (`meals` module).
- **Frontend**: Settings → Kitchen gets a provider picker; Recipes/Meals
  source badges, colors (`--source-mealie`/`--source-tandoor` tokens), and
  all 24 locales generalized off the hardcoded Mealie-only strings.
- **Security fixes found in review**:
  - Both adapters now go through the SSRF-hardened client
    (`server/utils/http.js`), opt-in via `RECIPE_PROVIDER_ALLOW_PRIVATE_NETWORK`,
    wired through the installer (schema, web wizard, Portainer/Unraid, docs,
    all 24 installer locales) — same pattern already used for ICS/WebDAV.
  - Tandoor thumbnail proxy no longer follows a server-supplied absolute
    image URL to an arbitrary host with the account's Bearer token attached;
    now origin-pinned to the configured `base_url`, 502 on mismatch.
- **Bug fixes**: Tandoor pagination parsed the DRF `next` link by
  string-slicing against `base_url` (breaks behind a reverse proxy with a
  different host) — now host-agnostic URL parsing. Tandoor's light-mode
  badge color darkened 4.24:1 → 5.5:1 for WCAG AA. Provider badge no longer
  shifts position on recipes with zero ingredients.
- **Tests**: new migration v131 suite (rename mechanics, FK/index/trigger,
  cascade), Tandoor adapter/sync/route suites, an SSRF-adoption regression
  test (pins that every `*_url`-backed module actually calls the guard), and
  a generalized badge-contrast test that discovers providers from CSS
  instead of hardcoding Mealie/Tandoor. `npm test`: 140/140 green.
- CHANGELOG.md updated (Added/Changed/Fixed/Security).

## Checklist

- [x] `npm test` passes (140/140)
- [x] CHANGELOG.md updated

## Disclosure
Same as before. This code was entirely generated with claude. I believe I have kept true to the contributing and code of conduct md while still providing the request where we now have a generalized external recipe provider mirror with 2 adapter. Assuming this is approved, I'll start looking into recipe sage next as the 3rd adapter.


ulsklyc:
Merged as 5b3ccd9d (via 1aa06e42), rebased onto v2.2.3 with your authorship kept on the commit. Thank you - this is the most careful contribution the repository has had in a while. The SSRF hardening you added in your own review round, and in particular pinning Tandoor's upstream-supplied image URL to the account origin before attaching the Bearer token, is exactly the failure nobody would have found later. Issue #530 had 15 thumbs up; this closes it.

What changed during the rebase, so the diff you see on main is not a surprise:

- **The migration is v134, not v132.** `main` had already used 132 (tasks archive) and 133 (shopping item order) since your branch forked. A colliding number is worse than a wrong one: an existing installation records the version as applied and silently skips the rename, then boots against `mealie_accounts` with code that only knows `recipe_provider_accounts`. `test-recipe-provider-migration.js` moved with it (base is now `<= 133`).
- **Locale conflicts resolved in your favour** - the mealie* keys in `main` had not moved since the fork, so the conflicts were context-only. The em dashes went out of the new lines: this repository writes `-`. Chinese kept its `——`, which is correct punctuation there rather than a stylistic import.
- **`recipes.js`**: your "render the ingredient count even at 0" rule stayed, on the current class name. `.kitchen-row__meta` became `.list-row__meta` in the meantime, and the 15ch rule that keeps the badge in place now hangs off that name.
- **`tokens.css`**: your provider-badge comment replaced only the old Mealie heading; the 2026-07-30 note above it about the meal colours is a recorded review decision and stayed.
- **The changelog entries moved** from the `[Unreleased]` block they were written into (long since released as 1.85.0) to the current one. I also sharpened the SSRF entry into an upgrade notice: for anyone whose Mealie answers on a Docker-internal address - the common case, and the one your own URL hint describes - the sync stops until `RECIPE_PROVIDER_ALLOW_PRIVATE_NETWORK=true` is set. That is the right default, but it needs to be visible before someone reports it as a broken sync.

Full `npm test` is green on the merge.

ulsklyc:
Released in [v2.3.0](https://github.com/ulsklyc/yuvomi/releases/tag/v2.3.0). Both GHCR images carry the tag, and the Umbrel store PR is updated.

Thanks again - and a note on what your work is still doing after the merge. Because the adapter interface is now the shape of the thing, the docs describe a provider-neutral module rather than "Mealie, and also Tandoor": the README, the landing page and `docs/SPEC.md` all read that way now, and a third provider is an adapter plus a badge colour. That was the promise #614 left behind, and you are the one who kept it.

The SSRF entry is in the release notes as an upgrade notice rather than a security footnote, since a Docker-internal Mealie is the common case and its sync stops until `RECIPE_PROVIDER_ALLOW_PRIVATE_NETWORK=true` is set. If that generates questions, they will land on #530 and I will handle them - you should not be on the hook for support on a default I agreed with.

PR #691, opened 2026-08-06
feat(budget): support revenue subcategories and fix related tests

ContatoLucasSonntag:
# feat(budget): support revenue subcategories and fix related test failures

## Summary

This Pull Request improves the **Budget** module by adding support for subcategories to **Revenue** categories and improving the **New Entry** form layout.

### Budget

* Move the **Subcategory** field below the **Category** field in the **New Entry** form.
* Allow creating subcategories for **Revenue** categories.
* Keep the subcategory list synchronized with the selected category.
* Allow selecting revenue subcategories when creating or editing budget entries.

## Motivation

Previously, subcategories were only available for expense categories, limiting how revenue entries could be organized.

With this change, revenue categories can now be organized hierarchically. For example:

* Salary

  * Monthly Salary
  * Overtime
  * Bonus

* Investments

  * Dividends
  * Interest
  * Returns

## How to Test

1. Open the **Budget** page.
2. Go to the category manager.
3. Create or edit a **Revenue** category.
4. Create a subcategory for that category.
5. Open the **New Entry** form.
6. Select **Revenue** as the entry type.
7. Select the configured category.
8. Verify that the **Subcategory** field appears directly below the **Category** field.
9. Select a subcategory.
10. Save the entry and verify that it is stored correctly.

## Expected Result

* Revenue categories support subcategories.
* The **Subcategory** field is displayed below the **Category** field.
* Revenue entries can be saved with a selected subcategory.

## Tests

In addition to validating the changes to the **Budget** module, this Pull Request also fixes test failures **that are unrelated to the implemented Budget functionality**:

* ✅ `test/test-shopping.js`

  * `Detail-Refresh aktualisiert die Zeile, ohne das .shopping-item zu ersetzen (Swipe-Closures bleiben intakt): ein Kategoriewechsel muss die Gruppierung neu aufbauen`

* ✅ `test/test-detail-view.js`

  * `Kontaktdaten kommen als Text ins DOM, nie als Markup`

The following test is still failing and **was not fixed** in this Pull Request because it is unrelated to the changes introduced here:

* ⚠️ `test/test-password-normalization.js`

The changes in this Pull Request do not affect the behavior of this test. It will be addressed in a separate Pull Request.

### DIFF
--- a/public/pages/budget.js
+++ b/public/pages/budget.js
@@ -932,7 +932,7 @@ function renderEntries() {
     const recurTag  = e.is_recurring
       ? ` <span class="budget-recur-mark" role="img" aria-label="${t('budget.recurringLabel')}"><i data-lucide="repeat" class="icon-sm" aria-hidden="true"></i></span>${e.recurrence_virtual ? ' ' + t('budget.virtualBudgetBadge') : ''}`
       : (e.recurrence_parent_id ? ` <span class="budget-recur-mark" role="img" aria-label="${t('budget.recurringInstanceLabel')}"><i data-lucide="corner-down-left" class="icon-sm" aria-hidden="true"></i></span>` : '');
-    const categoryMeta = isIncome || !e.subcategory
+    const categoryMeta = !e.subcategory
       ? categoryLabel(e.category)
       : `${categoryLabel(e.category)} · ${subcategoryLabel(e.subcategory)}`;
     const acctName = accountName(e.account_id);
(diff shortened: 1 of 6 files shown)


ulsklyc:
Merged as 471fa3f6 (via ab118636), with your authorship kept on the commit. Thank you.

Two things I changed while integrating, so you know what landed:

- **The two test changes are not in.** `test-shopping.js` and `test-detail-view.js` were failing for you because of CRLF line endings on your checkout, not because of anything in the tests. The repository already declares `* text=auto` in `.gitattributes`, so the fix belongs in your local git config (`git config core.autocrlf input`) rather than in the guards. Loosening a regex to accept `\r\n` would weaken it for everyone else permanently.
- **`expenseSubcategories` is now `subcategories`** in the `/budget/meta` response. The server had always grouped every subcategory into it regardless of type; only the name claimed otherwise. Since your change makes that visible to users, the name had to follow.

The feature itself went in as you wrote it, including moving the field out from under "Advanced" - it belongs next to the category it qualifies.

ulsklyc:
Released in [v2.3.0](https://github.com/ulsklyc/yuvomi/releases/tag/v2.3.0). Thank you.

One thing worth recording, because it is the reason this change was small: the server had always grouped every subcategory into that meta field without looking at the category type. Only the three gates around it were expense-only - the create route, `validateSubcategory()` and the entry form - and each of them stated the rule independently. That is why removing them needed no migration and no backfill: the data model never had the restriction, only the code did, in triplicate.

The field is now called `subcategories` in the API response, since it had never contained only expense ones.

PR #692, opened 2026-08-06
feat(notifications): add generic webhook notification channel

ContatoLucasSonntag:
# feat(notifications): add generic webhook notification channel

image

## Discussions
Discord notifications support #660 https://github.com/ulsklyc/yuvomi/discussions/660

## Summary

This Pull Request introduces a new **Webhook** notification provider, allowing Yuvomi to send notifications to external services through configurable HTTP endpoints.

## Changes

- Add **Webhook** as a new notification provider.
- Allow configuring a destination webhook URL.
- Add optional Bearer Token authentication for secured endpoints.
- Enable integration with external services that support HTTP webhooks.

## Motivation

The notification system was previously limited to the providers natively supported by the application.

By introducing a generic webhook provider, users can integrate Yuvomi with a wide variety of third-party services and automation platforms without requiring dedicated integrations.

Examples include:

- Home Assistant
- n8n
- Make
- Zapier
- Discord-compatible endpoints
- Custom APIs
- Self-hosted automation services
- Any service capable of receiving HTTP webhook requests

## How to Test

1. Navigate to **Settings → Notifications**.
2. Create a new notification channel.
3. Select **Webhook** as the provider.
4. Configure the destination webhook URL.
5. (Optional) Configure a Bearer Token.
6. Save the channel.
7. Trigger any notification supported by the application.
8. Verify that the webhook request is successfully delivered to the configured endpoint.

## Expected Result

- Webhook channels can be created and configured successfully.
- Notifications are delivered to the configured webhook endpoint.
- When configured, the Bearer Token is sent using the `Authorization` header.
- Invalid endpoints return an appropriate error without affecting the application's notification flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Administrators can now send reminders to generic HTTP webhooks alongside Web Push, Gotify, and ntfy. Each webhook receives a documented JSON payload through `POST`, may use an optional write-only Bearer token, can be verified from Notification settings, and participates in the existing per-channel retry and deduplication flow.
+
 ## [1.87.0] - 2026-08-06
 
 ### Added
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,6 +62,7 @@ npm run test:meals
 npm run test:calendar
 npm run test:ncb            # notes, contacts, budget
 npm run test:reminders
+npm run test:notifications    # Web Push and external notification channels
 npm run test:dashboard
 npm run test:api
 npm run test:ics-parser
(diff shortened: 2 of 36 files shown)


ulsklyc:
Thanks for this one, and sorry for the wait. Two of your other PRs (#691, #693) went out today in [v2.3.0](https://github.com/ulsklyc/yuvomi/releases/tag/v2.3.0). This one I want to keep, but not in its current shape, and I owe you the reasoning rather than a verdict.

**The problem is the fixed payload.** The provider sends:

```json
{ "event": "notification", "notification": { ... }, "sentAt": "..." }
```

That is a Yuvomi-shaped body. Home Assistant and n8n accept any JSON, so for those two it works as advertised. Discord does not: a webhook there requires `content` or `embeds`, and anything else comes back as 400 "Cannot send an empty message". So the "Discord-compatible endpoints" line in the PR description does not hold, and discussion #660 - the one this links to - stays unsolved.

**And #660 asks for two different things.** The original post wants Discord notifications, which is the case above. The only comment on it, which is where the upvotes are, asks for the opposite direction: *"a notifications section and an API to call it and then it appears in Yuvomi, so it can be used with Discord bots, Home Assistant and more."* That is inbound - external systems writing into Yuvomi. This PR is outbound. I read that second one as its own feature, not something this channel should try to cover, and I will keep it open in #660.

**What would make this land: a payload template.** One text field on the channel, with `{{title}}`, `{{body}}`, `{{url}}` and `{{tag}}` placeholders, left empty means today's default body. Then a Discord user pastes `{"content": "{{title}} - {{body}}"}`, a Slack user pastes their own shape, and Home Assistant users leave it blank. One generic provider covers all of them, and Yuvomi never needs a service-specific adapter - which is exactly what your PR description promises. It is a small change: the template field in `notification-channels.js` next to `baseUrl`, a substitution in `webhook.js`, one input in the settings page, plus the i18n keys.

Two smaller things for whenever you pick it up: the branch needs a rebase onto v2.3.0 (it conflicts now), and the new locale strings use em dashes - this repository writes `-` throughout, in all languages except Chinese where `——` is correct punctuation.

Do you want to take the template on? I am happy to leave it with you, and equally happy to build it myself if you would rather not - your call, and either way the idea and the groundwork are yours.

ContatoLucasSonntag:
Thanks for the detailed feedback and explanation!

Yes, please feel free to take over and implement the payload template. I really like the approach you suggested, and I agree that it makes the webhook provider much more flexible for Discord, Slack, Home Assistant, n8n, and other services without requiring service-specific adapters.

Unfortunately, I don't have much time available at the moment to work on the implementation and rebase, so I'd be happy for you to continue from the groundwork already in this PR.

Thanks again for taking the time to review it and for keeping the idea moving forward!

PR #693, opened 2026-08-06
feat(dashboard): add photo screensaver to prevent display burn-in

ContatoLucasSonntag:
# feat(dashboard): add photo screensaver to prevent display burn-in

## Summary

This Pull Request introduces a photo screensaver that is displayed when the dashboard is idle, helping prevent display burn-in on tablets and wall-mounted devices.

## Changes

- Add an idle screensaver for the dashboard.
- Display photos instead of keeping the dashboard visible while the device is inactive.
- Integrate the screensaver with Immich as the photo source.
- Automatically return to the dashboard after user interaction.

## Motivation

Previously, the dashboard remained visible when the device was not in use, increasing the risk of display burn-in on devices with static screens.

This implementation displays photos while the device is idle, providing a more visually appealing experience and helping preserve the display.

## How to Test

1. Open the dashboard.
2. Leave the device idle until the configured timeout expires.
3. Verify that the photo screensaver is displayed.
4. Touch the screen or interact with the device.
5. Verify that the dashboard is restored immediately.

## Expected Result

- The photo screensaver is displayed after the idle timeout.
- Photos are loaded from the configured Immich instance.
- User interaction immediately closes the screensaver and returns to the dashboard.
- The dashboard is no longer displayed continuously while the device is idle, reducing the risk of burn-in.


ulsklyc:
Merged as cfe19307 (via 9c30e60d), rebased onto v2.2.3 with your authorship kept. Thank you - a wall tablet that burns its own dashboard into the panel is a real problem, and the server side of this was careful work: the API key never reaches the browser, and the thumbnail proxy refuses an asset id that is not a UUID before it builds an outgoing request.

Five things changed on the way in, all in the frontend:

- **`innerHTML` is banned in this repository** (a save hook and `test:frontend-audit` both enforce it, no exceptions). `overlay.innerHTML = '<p></p>'` is now `createElement` + `append`.
- **`Intl.DateTimeFormat` is not used directly in pages or components** - the caption now goes through `formatDate()` from `public/i18n.js`, so it follows the user's date-format preference like every other date in the app.
- **`fetch()` is not called directly from the frontend** - the photo list goes through `api.get()`, which carries CSRF and session handling.
- **The settings copy guard** wanted the leaf to actually say what its description promises: `immichPurpose` now names the screensaver.
- **Two guards had to be told about the new surface**: the settings IA distribution (admin 7 -> 8) and the installer schema, where the three `IMMICH_*` variables are now recorded as a deliberate exception - the screensaver is set up in the app, not in the install wizard.

I also extended `test-screensaver.js` from 2 to 6 cases: `requireAdmin` on the configuration routes, the quiet unconfigured `/photos` response, and the UUID requirement for both the album and the asset id - both of them end up inside an outgoing Immich request, so they are worth pinning. The `__test` export in the component went away; it was never reachable from a test.

ulsklyc:
Released in [v2.3.0](https://github.com/ulsklyc/yuvomi/releases/tag/v2.3.0), together with your budget change from #691. Thank you for both.

Two of your contributions shipped in one release, and they are quite different kinds of work: #691 found a rule that was written down in three independent places and only enforced in one, and this one added a whole surface with a server-side credential boundary. The screensaver also picked up documentation of its own - `docs/immich-screensaver.md`, a card on the install page, and a section in `docs/SPEC.md` - because it is the first feature aimed squarely at a wall tablet rather than a phone.

The frontend conventions I corrected on the way in are all written down in `CONTRIBUTING.md` and enforced by hooks and `test:frontend-audit`; running `npm test` locally would have flagged the `innerHTML` line before the CI did. Not a complaint - just the shortest path past that whole class of review comment next time.

PR #697, opened 2026-08-08
Let API tokens act as a selected family member

JakeTheRabbit:
## Summary

- let an administrator choose the family member an API integration token acts as
- retain the administrator creator separately for audit while applying the subject's identity, role, ownership, and configured module permissions
- expose the subject selector in the existing API-token settings page and include subject metadata when listing tokens
- reject split-expense guests and cascade token deletion when either its creator or subject is deleted
- preserve existing tokens by backfilling their creator as subject

## Why

External connectors such as bank imports need least-privilege `budget:write` credentials, but imported personal transactions must belong to the selected household member rather than whichever administrator provisioned the integration.

## Security and compatibility

- token scopes remain an additional allow-list and cannot expand the selected member's module permissions
- non-admin subject roles cannot use admin-only routes
- the raw token remains one-time-only in the UI
- existing tokens keep their creator-as-subject behaviour after migration
- split-expense guest accounts cannot be selected

## Verification

- `node --experimental-sqlite --test test/test-api-token-subject.js` (4 passed)
- `node --experimental-sqlite --test test/test-token-scopes.js` (16 passed)
- `node --experimental-sqlite --test test/test-permissions.js` (16 passed)
- `node --experimental-sqlite --test test/test-permissions-routes.js` (13 passed)
- `node --test test/test-openapi-structure.js` (5 passed)
- `node --test test/test-frontend-audit.js` (215 passed)
- `node --experimental-sqlite test/test-db.js` (37 passed)

The local host intentionally blocks npm as a supply-chain control, so the focused commands above were run directly; the new test is also wired into the repository's normal `npm test` chain for CI.


JakeTheRabbit:
Rebased this PR onto exact upstream `v2.1.0` (`8f9dfbb1`). The branch is now one commit ahead of `main` and zero behind.

Updated focused verification:

- API-token subject tests: 4/4
- token scopes: 16/16
- auth user identity: 1/1
- permissions: 16/16
- module registry: 24/24 plus parity 8/8
- OpenAPI structure: 5/5
- schema reconciliation: 4/4
- `git diff --check origin/main..HEAD`: clean

The bounded upstream test chain reached step 91 of 183: 88 steps passed; the `test-shopping.js` and `test:detail-view` failures reproduced identically on pristine v2.1.0 and are not introduced by this PR. `test:password-normalization` exceeded the bounded 184-second run, so no claim is made for steps 91-183.

ulsklyc:
Merged as e5f70a29. Thank you - and I owe you a correction on my own first read.

I had this filed as a large change to the auth core for a niche case. Then I checked what the alternative would look like and found that `budget_entries.owner_id` is fixed to the creator on insert, by design (#476/#505), and is explicitly not settable on update either. So there is no other way to attribute an imported transaction to the member it belongs to. The obvious alternative - accepting `owner_id` in the request body - is the weaker one: it lets any caller claim any ownership on every call, where a token subject is fixed once by an admin and not selectable afterwards. Your design is at the right layer, and the separation of creator (accountable, audited) from subject (identity, permissions) is the part that makes it defensible.

Three things changed on the way in:

- **The migration is v135, not v134.** v134 landed a few hours before this, with the Tandoor recipe adapters in #686. Same reason as always: a colliding number is recorded as applied and then silently skipped on an existing installation.
- **The second commit is not in.** `fix(auth): reconcile collided token-subject migration` rebuilt `reconcileCriticalSchema` so the new `after` block runs for every entry on every start, to survive exactly the v134 collision above. With a free number there is no collision to survive, and I would rather not change that self-healing path for a feature that does not need it. Only the first commit was taken; the reconcile code is untouched.
- **The subject picker got its own i18n key** (`settings.apiTokenSubjectLabel`, "Acts as") instead of borrowing `settings.familyRoleOther`. A borrowed key reads correctly right up until someone rewords the original.

I also wrote the reasoning into `docs/SPEC.md` under API Tokens - specifically why the subject sits in the token rather than in the request - so the next person to touch this finds the argument rather than re-deriving it.

Full `npm test` is green. It will go out with the next release.

PR #728, opened 2026-08-12
docs(modules): say how a module with its own backend stays safe

JakeTheRabbit:
## Summary

- add a "Modules With A Backend Service" section to `MODULES.md`
- state the storage boundary: no direct `yuvomi.db` access, `/api/v1` only
- state the identity rule: re-check the session server-side through `GET /api/v1/auth/me`, never trust a user id or role sent from the module's own browser code
- state that the module's service needs its own CSRF pair, and that unattended jobs use a scoped API token rather than a session

## Why

`MODULES.md` covers browser-only modules well. A module that needs stored state, scheduled work, or a third-party credential has to run a service of its own, and the guide currently says nothing about what that service owes core beyond the one paragraph on the service-worker cache path.

The two mistakes that follow are both easy to make and expensive:

- Opening `yuvomi.db` directly. A second writer breaks core's migrations, and the schema moves between releases. It works on the day it is written and dies on the next upgrade - and the report lands on you, not on the module author.
- Trusting a user id or role sent from the module's own browser code. A module page is not a trusted caller, and treating it as one turns a page bug into an authorization bypass.

## Scope

Docs only. Every mechanism described already exists: the session cookie, `/api/v1/auth/me`, API token scopes, and the `/api/` service-worker rule already documented one paragraph earlier. No core change is implied or required, and no code paths are touched.

Written from building https://github.com/JakeTheRabbit/yuvomi-extensions, a sidecar platform that runs modules this way against an unmodified Yuvomi image.

### DIFF
--- a/MODULES.md
+++ b/MODULES.md
@@ -96,6 +96,8 @@ Do not open `yuvomi.db`. It is core's private storage: the schema changes betwee
 
 Re-check identity on the server for every request. Forward the incoming Yuvomi session cookie to `GET /api/v1/auth/me` over the internal Yuvomi URL, and trust only that response for the user id, role, and permissions. The browser half of a module is not a trusted caller: never accept a user id or role from a request body.
 
+Cache that answer briefly rather than resolving it on every call. Yuvomi rate-limits `/api/` to 300 requests per minute per IP, and a service that does not forward the caller's address spends that budget from its own container IP for all of its users at once - the first symptom is a `429` for everyone. A few seconds of cache keyed on the session cookie is enough, and short enough that a logout still takes effect.
+
 Yuvomi's CSRF token protects Yuvomi's endpoints, not a module's. State-changing routes on the service should independently require:
 
 - a valid Yuvomi session, verified as above;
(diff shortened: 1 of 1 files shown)


ulsklyc:
Thanks - this is written from real practice and it lands in the right place. `MODULES.md` had nothing to say about a module that owns state, and the two failure modes you name (opening `yuvomi.db`, trusting a user id from the page) are exactly the two that would have generated reports against core.

I verified every mechanism you describe against the code and they all hold: `/api/v1/auth/me` at `server/auth.js:1266`, the double-submit CSRF pair at `server/middleware/csrf.js`, and `budget:read` / `budget:write` exist verbatim in `server/scopes.js:31`. Your point that Yuvomi's CSRF token does not protect a module's endpoints is also literally true in code: API-token requests skip the middleware entirely (`server/middleware/csrf.js:30`).

Two things I am adding rather than sending back to you, since they are small and I would rather not have you wait on a round trip:

1. **Rate limit.** `app.use('/api/', apiLimiter)` allows 300 requests per minute per IP (`server/index.js:163-222`). "Re-check identity on the server for every request" means one extra `/api/v1/auth/me` per incoming request, and unless the service forwards the caller's IP those all count against a single container IP - so the budget is shared across every user of the module at once, and the failure is a 429 for all of them. I am adding a sentence recommending a short-lived cache of the resolved identity keyed on the session cookie.

2. **Token issuance is admin-only.** `POST /auth/api-tokens` carries `requireAdmin` (`server/auth.js:1363`) and the settings page is `adminOnly: true`. "Issue an API token in Settings" is not available to every member, so I am naming the path and the restriction.

Merging with those two edits pushed to your branch.

PR #729, opened 2026-08-12
docs(modules): say how a module survives a yuvomi upgrade

JakeTheRabbit:
## Summary

- add a "Compatibility Across Yuvomi Releases" section to `MODULES.md`
- point modules at `GET /api/v1/version` and the authenticated OpenAPI document as the self-check surface
- describe degrading to read-only instead of failing part way through a write
- state the surface third-party modules may rely on: `/api/v1` and the public browser libraries, not `server/` internals or undocumented response fields

## Why

`module.json` carries the module's own version and nothing about the Yuvomi release it was written against, and loading is not gated on a compatibility range. A module calling an endpoint that a later release renamed therefore keeps loading and fails at the point of use - in front of the user, part way through a write.

The endpoints needed to avoid that already exist. This documents the pattern, and more usefully draws the line on what is supported, so a module built on a private helper is the module author's problem rather than a regression report filed against core.

## Scope

Docs only. No new endpoints, no core change, no code paths touched.

Touches a different part of `MODULES.md` than the backend-service PR, so the two apply cleanly in either order.

### DIFF
--- a/MODULES.md
+++ b/MODULES.md
@@ -92,6 +92,19 @@ Yuvomi scans `modules/` and validates each `module.json`. Invalid modules are sh
 
 Admins can enable, disable, and order modules in Settings -> Personal -> Navigation. Ordering is per user and open to every member; the enable/disable switches are admin-only. Copying a new folder into `modules/` makes it appear there automatically.
 
+## Compatibility Across Yuvomi Releases
+
+`module.json` records the module's own version, not the Yuvomi version it was written against, and Yuvomi does not gate loading on a compatibility range. A module that calls an endpoint a later release renamed or moved therefore keeps loading and fails at the point of use, in front of the user.
+
+Two endpoints let a module check for itself before it acts:
+
+- `GET /api/v1/version` returns the running Yuvomi version.
+- The authenticated OpenAPI document describes the operations that version actually serves.
+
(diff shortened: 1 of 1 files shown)


ulsklyc:
Thanks - the diagnosis is right. `module.json` really does carry only the module's own version and loading is not gated on a compatibility range (`server/services/modules.js:59-90`), so a module that calls a renamed endpoint does keep loading and does fail in front of the user. Documenting the pattern is worth doing.

Three edits I am making before merge:

1. **The OpenAPI document is admin-only.** `/api/v1/openapi.json` and `/openapi.json` both carry `requireAuth, requireAdmin` (`server/index.js:322-324`). As written, "compare the operations it requires against that document when it starts" returns 403 for every family member who is not an admin, which is most of them - so the self-check as described is not runnable in the normal case. I am reframing it as a development-time and admin-facing check, and giving the runtime path that does work without admin: watch the version, and treat a `404` or `405` on a previously working endpoint as the signal to degrade.

2. **`/api/v1/version` needs a credential.** Without a session or API token the response omits `version` entirely (`server/index.js:266-276`). Fine for a module page, not fine for an unattended job, so I am saying so.

3. **The stability sentence.** This is the one that is not really a docs change. "The supported surface for third-party modules is `/api/v1`" turns an internal API into a public contract - and until now I have changed `/api/v1` freely (#593, #617 and #638 all moved endpoints or response shapes). I want to give module authors the line you are asking for without committing a solo project to compatibility guarantees it cannot keep, so I am softening it to what I can actually hold: `/api/v1` is the surface to build on, and breaking changes to it get called out in the CHANGELOG. Everything you list as outside the surface stays outside it.

Merging with those edits pushed to your branch.

PR #737, opened 2026-08-12
fix(calendar): Fix default date selector if in day view

LycidasFfm:
## Summary

Added functionality in the day view of the calendar to directly open a event on that specific day.

## Changes

- 

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Creating an event from day view now defaults to the day you're looking at.** Clicking an empty slot in the day grid already opened the new-event form on that day, but the "+" button in the toolbar and the floating action button ignored it and always defaulted to today, so switching a few days ahead and reaching for either "+" silently created the event on the wrong day. Both now pass the day currently shown in day view through to the form; the other views are unaffected, since their reference date isn't a specific day the user picked.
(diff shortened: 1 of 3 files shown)


LycidasFfm:
Hey @ulsklyc and thanks for Yuvomi, really love to use it!
Just added a tiny thing, that already annoyed me several times ;)

ulsklyc:
Merged, with one commit added on top. Thank you for this one, and in particular for the diagnosis rather than just the report: you had located the cause exactly, and that is the expensive half of the work.

I took it further than your patch, so here is what changed and why.

**The cause was not specific to day view.** Your reasoning for leaving the other three alone was that their reference date is not a specific day the user picked, and that is true as far as it goes. But the consequence is the same wherever you have paged away from today. In month view, standing in September and reaching for "+", the appointment was created in August. Week view had it too, and agenda view, where the cursor is literally the first day of the list on screen. Same fall-back to today, same surprise, only further away.

**So the fix is one rule instead of one special case:** today wins as long as the current view actually shows today, otherwise the first day of the period on display. That keeps the old behaviour exactly where it was right - for anyone who does not page around, nothing changes at all - and it answers all four views without a per-view exception.

Two details worth naming, in case you look at the code:

- It deliberately does not go through `getRangeForView()`, which would have been the obvious reuse. That function's month span is the 42-day grid and starts in the previous month, so September would have proposed 31 August. The month's own first day is the right answer here.
- The empty state of the search keeps falling back to today, and that is intentional. A result list is not a period, so there is nothing on screen for a proposal to refer to.

The guard that came with it checks the rule across every create call rather than a list of known lines, so a new button that forgets the date fails the test rather than slipping through. It is red against the state before your commit (four calls without a date instead of one), which is the only thing that makes it worth having.

Your commit is in `main` unchanged and under your name. The `## [Unreleased]` entry now describes the whole thing and credits you; your branch needed a rebase because four releases had landed since you opened this, and the only conflict was that changelog block.

It ships in the next release.

PR #741, opened 2026-08-12
feat(inventory): a new module for tracking owned belongings

keinzufall85-neo:
Based on https://github.com/ulsklyc/yuvomi/discussions/696, this can be seen as an RFC for the (potential) Inventory module. When you answered my initial post, I was already quite far building it, so I thought I'd rather push then drop it.

It (to my best effort) follows every coding and contribution guidline that Yuvomi gives. It's also already rebased onto 2.7.1 and did pass all tests. Happy to hear your comment. :)

## Summary

Adds **Inventory**, a new core module for tracking owned household belongings — what you have,
where it's stored, and when something about it needs attention. Proposed in
https://github.com/ulsklyc/yuvomi/discussions/696; built in stages so each is independently
reviewable, and further stages beyond what's here (multi-photo galleries, calendar integration for
deadlines) are still open.

One deliberate departure from the original proposal: items are tracked individually rather than as
**stacks of duplicates** (the Discussion suggested a Pantry-like stacking model for multiples of the
same thing). Pantry already owns the "stackable, quantity-tracked consumable" use case; Inventory is
aimed at durable, individually-identifiable belongings (serial numbers, warranties, condition),
where each unit's own history matters. Happy to revisit if that's a hard requirement.

## Changes

- **Stage 1 — core CRUD:** items with brand, model, serial number, category, storage location,
  purchase price/date, warranty length, condition and status. Deliberately no current-value/resale
  field — per discussion #696, a manually maintained number nobody updates is worse than no number
  at all. Storage locations are a two-level hierarchy (e.g. "Garage" →
  "Werkzeugschrank"), renameable/sortable through the shared category-manager component, same as
  Pantry Locations. Categories are a manageable list (five seeded defaults), same pattern as Task
  Categories — including localizing the seeded set via `label_key` rather than a hardcoded name, so
  they translate instead of staying German regardless of the app's language.
- **Stage 2 — document linking:** attach receipts, warranty cards, or manuals from the Documents
  module, reusing Budget's existing visibility-filtered document-linking mechanism.
- **Stage 3 — budget linking:** link a purchase, refund, repair, or later accessory to a Budget
  entry, with a role per link. Creating an item straight from a booking (via a new "Add to
  inventory" hook in the Budget entry modal) prefills the purchase price. Visibility follows
  Budget's own rules exactly, including personal budget mode.
- **Stage 4 — deadlines:** a derived warranty-end reminder (30 days out, computed on the fly, never
  stored) plus a dedicated, subscribable read-only ICS feed for warranty end dates.
- **Custom tracked dates:** up to 10 additional per-item dates beyond the warranty (TÜV, service,
  insurance renewal, ...), each with its own reminder lead time — join the same reminder system and
  the same ICS feed (renamed from "warranty deadlines" to "inventory deadlines" to reflect the
  broader scope).
- **List redesign:** two-level browse (metric cards + category overview, then a category's items
  grouped by storage location), "All" / "Needs attention" filter chips scoped to a category, a
  single photo per item, a read-only detail view before editing, and a nav badge for items with a
  soon-expiring or overdue warranty/tracked date. The landing page's "Needs attention" metric card
  is itself a filter too — tapping it flattens the list across every category to just the items
  that need a look, reusing the same view search results already fall back to.
- **Nav placement:** lives in the Household section (moved there from Finance once the module's own
  weight settled — it's primarily a record of long-lived household items, not a finance tool).
- Full i18n across all 24 locales, OpenAPI registration (and MCP for free via the existing
  OpenAPI bridge), permissions/scopes wiring, and 14 test suites (all cross-referenced in
  `docs/test-suites.md`).
- `docs/SPEC.md` and `README.md`/`README.de.md` updated with the new module's data model and
  feature description.

## Checklist

- [x] `npm test` passes


ulsklyc:
**Yes, this lands.** I want to say that first, because you built 100 commits on an open question and deserve an answer rather than a review queue.

Then the reasoning, and one change I am asking for before anything else.

## What decided it

In #696 I left one thing open, and it was not whether the code would be good: *"every module is a permanent line item in the navigation for every household, including the ones that will never track a bike."* That question is still there, and this PR does not answer it, it goes around it. So I am answering it now, and the answer has a condition attached.

Three things moved me toward yes:

- **You extended the existing reminder system instead of building a second one.** Migration 139 rebuilds the `reminders` table because SQLite cannot widen a CHECK with ALTER, and `entity_type = 'inventory_item'` then joins the same delivery, retry and dedup path everything else uses. That is the difference between a module and a module-shaped island, and it is the part I would otherwise have had to ask for.
- **The test chain is actually wired.** All 14 suites in `package.json` and in `docs/test-suites.md`, with descriptions that name the trap rather than the file. `test:reminders-entity-type-migration` calling out `foreignKeysOff` as mandatory, because the DROP would otherwise cascade the delivery log away with it, is the kind of note I write for myself.
- **You took both remarks from #696 seriously**, including the one where I agreed with you against your own field list: no current-value field, and the budget link follows the v1.67.0 attachment pattern instead of inventing a second one.

Your departure from the proposal, individual items rather than stacks, is the right call and for the right reason. Pantry owns "stackable and consumable". I would have pushed back had you brought stacking with you.

## The condition: it ships off by default

The navigation argument from #696 does not disappear because the code is good, so it gets solved rather than outvoted: **Inventory arrives disabled and each household switches it on.**

Concretely that is one migration adding `inventory` to `disabled_modules` where it is not already there. It runs on fresh installs too, so new households start without it and existing ones do not find an eighteenth entry after an update they did not ask for. Everything else stays as you built it.

This would be the first module with that default, and that is exactly why it fits: it is also the first whose audience is visibly a subset. A household that will never track a bike never sees it; one that wants its spreadsheet gone turns it on once. And if Inventory turns out to be what half the installs use, flipping that default later is a one-line change with an obvious signal behind it.

## Three smaller things

1. **Rebase onto v2.10.1.** Your branch sits on 2.7.1 and conflicts now. Migrations 137-143 still fit, `main` is at 136, but only as long as nothing else carrying a migration merges before you. I would rather not spend that window on something else in the meantime, so tell me roughly when you can get to it.
2. **Drop the eight `scripts/add-inventory-*-i18n-keys.mjs`.** They are scaffolding from your build process and would become permanent files nobody runs again.
3. **The deadlines feed token.** You store `inventory_deadlines_feed_token` household-wide in `sync_config`, while the calendar feed keeps `users.calendar_feed_token` per user. Two models for one concept, and the household-wide one cannot be revoked per person: whoever subscribed keeps access until everybody loses it. Please move it onto the per-user pattern the calendar feed already uses.

## Then

Once it is rebased and CI has run, I will review it properly. It has never been through CI, so that is the first real signal for both of us. Expect the review to be long rather than harsh: 96 files get read, not skimmed.

Thanks for building it, and for asking in #696 before you started. That is why this is a conversation about a default value instead of a rejection.

ulsklyc:
Follow-up on point 1: **I took the rebase off your hands.** Your branch is now on v2.10.1, it merges cleanly, and CI has run for the first time - green on both Node versions.

I pushed it straight to `feat/inventory`, so this PR is up to date without you doing anything. Two things worth knowing about how, since it is your branch:

**It is a merge, not a rebase.** Your branch sat 105 commits deep against 81 in main, and 24 of the overlapping files are locale files. A rebase replays every one of your commits against those, which is a hundred chances to lose something quietly in a JSON file nobody reads line by line. A merge resolves the same conflicts once and leaves your staged history untouched - which matters here, because the stages are how you asked for this to be reviewed. If you would rather have a linear history, say so and I will redo it as a rebase.

Only two files actually conflicted:

- **`package.json`** - the test chain. main gained `test:browser-loader-stubs` and `test:rrule-ui` while you were building; I took main's chain and put your inventory block back where you had it, after `test:subscriptions`. 196 suites, no duplicates.
- **`CHANGELOG.md`** - your six Unreleased lines kept verbatim, above the entry from main. I only pulled the em dashes to `-`, which is what the rest of the repo uses.

Your migrations needed no renumbering: 137-143 still follow straight on from main at 136.

**Two guards then failed, and both are worth a look because they are new since your base** (v2.8.0, so you could not have known):

1. `page-measure--narrow` was missing on the page root. The reading measure moved from the row carriers to the page, so without the role the toolbar and control rows end up next to their own body instead of flush with it.
2. The FAB needs `data-dock-label` from the shared `newLabel.*` register. On a pointer device the primary action docks into the module header as a labelled button, and without it that button is silently blank. I added `newLabel.inventory` across all 24 locales - please check the noun for your language if it is not one of the ones I would bet on.

Both are in a separate commit so you can see exactly what changed in your code, and nothing else of yours was touched.

That leaves points 2 and 3 from my earlier comment with you: the eight `scripts/add-inventory-*-i18n-keys.mjs`, and moving the deadlines feed token onto the per-user pattern. Plus the default-off migration. Once those are in, I will do the full read.

3581 tests pass locally, same as CI.

ulsklyc:
Nachzug zwei: your branch is now on **v2.11.0**, which went out a moment ago. CI is green again.

Only the CHANGELOG conflicted this time, and I resolved it by meaning rather than by text position: your six Inventory lines stay under `[Unreleased]` where they belong while this PR is open, and the `[2.11.0]` block keeps exactly what actually shipped. `package.json` and `public/sw.js` merged on their own, since 2.11.0 added no new test suite and the test chain was already correctly merged from last time.

**Your migrations are still untouched.** 137-143 sit flush on main at 136, and 2.11.0 deliberately carries none - the per-user setting it added uses the existing config table rather than a new column, precisely to keep that window yours.

Nothing changed in your code in this push. The three points from before are still with you whenever you have time:

1. the eight `scripts/add-inventory-*-i18n-keys.mjs`
2. the deadlines feed token onto the per-user pattern
3. the default-off migration

One practical note for when you do push: workflow runs on this PR sit at "action required" until a maintainer approves them, which is why it looked like CI had never run before. Ping me here after you push and I will approve it right away, so you are not waiting on a green tick that nobody triggered.

keinzufall85-neo:
All three points are in, plus one finding I deliberately did not act on.

1. The eight scaffolding scripts are gone. 497 lines. Nothing referenced them — not package.json, not the workflows, not docs/, not the tests — and .dockerignore excludes scripts/ from the image anyway, so they only ever existed in the tree. You were right that they were the exception: no other module left one behind.

2. The deadlines feed token is per user. Migration 144 mirrors migration 61 exactly — column plus partial UNIQUE index on users — and deletes the old sync_config row, which is itself the withdrawal that was previously impossible: an existing subscription URL stops working instead of lingering as an orphaned key. isValidFeedToken (bool) gave way to findUserIdByFeedToken, so a token resolves to its owner rather than just answering "valid".

One scope call worth weighing rather than discovering on the read-through: the admin gate came off with it. It was justified solely by the artifact being household-wide — my own module header argued exactly that — and server/routes/calendar/feed.js, the pattern you pointed me at, has none. Keeping it would have left members holding an individually revocable token they could not mint, which is half the point. /api/v1 also carries no module permissions for sessions (only API tokens do), so GET /inventory/items is already open to every member; the gate was not drawing a line that existed elsewhere.

The feed's content stays household-wide. Items have no owner and no visibility column, so buildInventoryDeadlinesFeed deliberately takes no user id — personal is the access, not the selection. I left a comment saying so, because the diff alone reads like an omission.

The two feeds share the pattern, not the token space: separate columns, separate indexes, separate functions. A new test pins that boundary — a calendar token does not open the inventory feed, and clearing one subscription leaves the other resolving. The only shared code remains escapeICSText/foldLine, as before.

While in there: /feed/inventory-deadlines/{token}.ics was the only feed route missing from the OpenAPI registration. Its calendar counterpart is in core.js; this one now is too.

3. Inventory ships disabled. Migration 145, merging inventory into disabled_modules rather than replacing the array — a household may already have modules switched off, and a blind write would silently switch them back on. Corrupt or non-array JSON counts as "nothing disabled", matching parseDisabledModules. No separate fresh-install path was needed: migrate() runs the whole array on a new database, so the single entry covers both audiences. There is no other seed path — I looked for one before assuming.

The finding I did not act on. Per-person revocation now works in the data model, but members cannot reach it in the product. /settings/sync/calendar is adminOnly, and findSettingsLeaf returns null for non-admins, so the route does not resolve — and that page hosts both the calendar feed section and the inventory one.

This is the shape you have already fixed twice:

registry.js:72 (#695) — tasks_default_target writes per user, "und dieses Blatt darf deshalb nicht adminOnly sein"
registry.js:108 (Critique 2026-07-27) — order and mobile slots are per-user with no admin check server-side; the leaf sat behind adminOnly anyway, so five of six family members could not configure their own navigation
Ours is the same: per-user setting, no server-side admin check, admin-only leaf. The page splits cleanly — CalDAV, ICS subscriptions and calendar import are household concerns and belong behind the gate; the two feed sections are the per-user tail.

I did not fix it, for two reasons. It is not a regression from this PR — the calendar feed has had it since migration 61, and I inherited it by matching the pattern you asked for. And fixing it properly means a new non-admin leaf plus i18n across 24 locales, changing your calendar feed's surface inside a PR about inventory. Splitting out only the inventory half would recreate the "two models for one concept" you asked me to remove, so that is the one option I would argue against. Happy to do it here if you would rather it not wait.

On the merge: keep it as a merge — no need to redo it as a rebase. Your reasoning about the locale files holds, and the staged history is still readable, which was the part I cared about.

**On being current** (edited to correct the paragraph that stood here): the branch is now merged up to `69cf1754`, i.e. through the v2.12.0 release, and is no longer behind. My earlier note claimed this round conflicted in all 24 locale files — that was wrong. I had read a `git merge-tree` "changed in both" as a conflict when it only means both sides touched the file; every locale file merged cleanly and `test:i18n` passes at full key parity. Apologies for the false alarm.

The only hand resolution was `CHANGELOG.md`, and the 2.12.0 release made it worth doing carefully: with main's Unreleased entries moved into a dated block there was nothing left to conflict against, so git merged it cleanly *and wrongly* — the eight Inventory lines landed inside `[2.12.0]` and `[Unreleased]` came out empty, claiming as shipped what is still an open PR. `[2.12.0]` is now byte-identical to main and the Inventory entries are back under `[Unreleased]`.

Migrations are unaffected throughout: main is still at 136 and 137–145 sit flush behind it, now across nine main commits and a release.

Tests: npm test green — 3610 assertions, 197 suites (one new: test:inventory-default-off-migration). Both new migrations are also exercised end to end by test-ics-subscription.js, which replays the real MIGRATIONS array against a fresh database.

ulsklyc:
All three are in, and I checked them rather than read them. Green across the board on my side too: full suite, plus the two new ones on their own (`inventory-default-off-migration` 7/7, `inventory-deadlines-feed` 15/15).

**1. Scaffolding.** Gone, and you were right that nothing referenced them.

**2. The feed token.** Migration 144 mirrors 61 exactly, the partial UNIQUE index is there, and the old `sync_config` row is deleted, which is the withdrawal that did not exist before. `findUserIdByFeedToken` over `isValidFeedToken` is the right call: a token that resolves to an owner is what makes a single revocation possible at all. Your test "Ein Rückzug trifft genau ein Abo, nicht alle" pins exactly that, and the one next to it pins that the two token spaces stay separate. Both are the assertions I would have asked for.

**3. Disabled by default.** Migration 145 merges instead of replacing, treats corrupt JSON the way `parseDisabledModules` does, and needs no separate seed path. Function-form `up` is supported (`server/db.js:5349`, four existing cases) and `sync_config.updated_at` exists, so the `ON CONFLICT` clause is sound. 137-145 sit flush behind main at 136 and nothing existing was touched.

## The admin gate coming off: confirmed, keep it off

I verified all three of your reasons instead of taking them: `server/routes/calendar/feed.js` carries no `requireAdmin`, `/api/v1/inventory` hangs off the global `requireAuth` with no module gate (`server/middleware/` has only csrf, require-admin and validate), and so the feed exposes nothing a member cannot already fetch from `GET /api/v1/inventory/items`. A gate there would have drawn a line that exists nowhere else in the codebase. Your reasoning was correct and the change stands.

## The finding you left: right call, and it is bigger than inventory

You were right not to fix it here, and the argument is stronger than the one you made. `registry.js:170` is `adminOnly: true`, and the whole `sync` domain is admin-only, so those two feed sections cannot stay there and be reachable. But the fix is not new ground: **the same move is already in that file three times**, each with a comment explaining it. `calendar-defaults` (line 56), `task-defaults` (line 72, #695), `navigation` (line 108). Per-user setting stranded on an admin-only leaf, moved to a leaf under `personal`.

So this is the fourth run of an established pattern, and it belongs in its own change for the reason you gave: pulling only the inventory half would rebuild exactly the two-models-for-one-concept I asked you to remove. It touches the calendar feed, which has nothing to do with inventory, inside a 10k-line PR about inventory.

I will do it as a follow-up right after this merges: one leaf under `personal` carrying both feeds, with CalDAV, ICS subscriptions and calendar import staying behind the gate as household concerns. Not your work to carry.

One note for that follow-up while it is fresh: `regenerateFeedToken` returns a token even when the user id is null, where the `UPDATE` matches no row. Yours does this because `calendar/feed.js:45` does, so it is not a defect in your PR - I am writing it down so the follow-up fixes both ends rather than one.

## On the merge

Nothing left for either of us: you pulled main up to v2.12.0 yourself at 09:51, so the branch is current and `git merge-tree` reports zero conflicts. Your 07:42 note about the 24 locale files was overtaken by your own commits.

Merging now. Thank you for the patience through three rounds of this - the module is better for every one of them, and the process notes you wrote along the way made reviewing it fast.

keinzufall85-neo:
By the way, I stumbled upon the following, potentially harming gate behaviour: `claude-code-review.yml:177-181 ` selects Claude's comments with no time filter, and `:186` fails only when the total is 0. Since the plugin deliberately stops on a PR it has already commented on (`:193`), the first review satisfies the check for every push after it. It proves a PR was reviewed once, not that the current head was reviewed.

PR #744, opened 2026-08-12
fix(tasks): preserve subtasks when spawning recurring task occurrences

### LINKED ISSUE #742: [Bug] Subtasks are not preserved with recurring parent tasks
### Yuvomi version

2.7.1

### Deployment

### Description

When a recurring task has subtasks, those subtasks should remain linked to the parent task and follow its recurrence.

Currently, when the parent task generates its next recurring occurrence, the subtasks are not correctly recreated and linked to the new occurrence of the parent task.

This causes recurring tasks that work as checklists or have multiple steps to lose their subtask structure after the first occurrence.

### Steps to reproduce

Create a new task.
Configure the task as recurring (for example, daily or weekly).
Add one or more subtasks to the recurring parent task.
Complete the parent task and its subtasks.
Wait for or trigger the next recurrence.
Open the newly generated occurrence of the parent task.
Check its subtasks.

When a recurring parent task generates a new occurrence:

The new parent task should also receive copies of all subtasks from the original recurring task.
The new subtasks should be linked to the new occurrence of the parent task.
Each new subtask should start with its appropriate initial state (for example, pending/not completed).
The subtask structure should be preserved across all future recurrences.
Completing a subtask in one occurrence should not affect the same subtask in previous or future occurrences.

Example:

Weekly task
├── Subtask A
├── Subtask B
└── Subtask C

When the next weekly occurrence is generated:

Weekly task (new occurrence)
├── Subtask A (new)
├── Subtask B (new)
└── Subtask C (new)

The new subtasks should be independent records linked to the new occurrence of the parent task.

### Browser / device / OS

Windows / Chome

### Logs

```shell

```

### Additional context

Subtasks should follow the recurrence lifecycle of the parent task.

Recurrence should be controlled by the parent task. Subtasks do not necessarily need their own independent recurrence configuration.

Whenever Yuvomi creates the next occurrence of a recurring parent task, it should copy the subtask structure and associate the new subtasks with the new occurrence of the parent task.

This would allow recurring tasks to work as reusable checklists or workflows while keeping each occurrence and its completion state independent.

t0msh:
## Summary
Fixed an issue where completing a recurring parent task with subtasks created the next recurring instance without copying its subtask structure, causing recurring checklists and multi-step workflows to lose subtasks after the first occurrence.

Closes #742 

#### `server/routes/tasks.js`

1. **`spawnRecurrenceFollowup(task)`**:
   - Queries all existing subtasks under the completed parent task: `SELECT * FROM tasks WHERE parent_task_id = task.id ORDER BY id ASC`.
   - For each subtask, inserts a new subtask record linked to `newTask.id` (`parent_task_id = newTask.id`).
   - Resets subtask `status` to `'open'`.
   - Preserves title, description, category, priority, points, visibility, `assigned_to`, and `created_by`.
   - Links cloned subtasks to their origin subtask via `recurrence_origin_id = sub.id`.
   - Shifts subtask `start_date` and `due_date` relative to the next recurrence date if dates exist.
   - Copies assignments (`task_assignments`) and tags (`task_tags`) associated with each subtask.

2. **`discardRecurrenceFollowup(taskId)`**:
   - Updated touched check: `SELECT 1 FROM tasks WHERE parent_task_id = ? AND (status != 'open' OR recurrence_origin_id IS NULL) LIMIT 1`.
   - Allows un-doing completion of a parent task to cleanly discard untouched spawned followup tasks and their cloned subtasks, while preserving followup tasks if any subtask was modified or manually added by the user.

#### `test/test-tasks-recurrence.js`
Added three unit tests:
_(titles here in English, but tests are labelled in German to match other unit tests)_
1. `Subtasks of a series are copied and reset when the subsequent instance is spawned(#742)`: Verifies that subtasks (whether `done` or `open`) are cloned under the new recurring parent task instance and reset to `status = 'open'`.
2. `Undo hooking of a series with pristine subtasks deletes the follow-up instance including subtasks (#742)`: Verifies that un-checking a parent task discards the followup task and its untouched cloned subtasks.
3. `Undo check of a series with a subtask completed retains the follow-up instance (#742)`: Verifies that un-checking a parent task preserves the followup task if any subtask on the followup instance was completed or edited.

- 

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Preserve subtasks when spawning recurring task occurrences.** Completing a recurring parent task with subtasks now copies its subtask structure with reset status (`open`) to the next occurrence, preserving checklists and multi-step workflows across occurrences (#742, #744). Undoing completion discards pristine spawned follow-up tasks while preserving any follow-up tasks whose subtasks were completed or edited.
+
 ## [2.8.2] - 2026-08-14
 
 ### Changed
--- a/server/routes/tasks.js
+++ b/server/routes/tasks.js
@@ -976,9 +976,60 @@ function recurrenceFollowupOf(taskId) {
 }
 
 /**
+ * Prüft, ob Unteraufgaben einer Folgeinstanz durch den Benutzer verändert wurden
+ * (editiert, erledigt, hinzugefügt oder gelöscht).
+ */
+function isFollowupSubtasksTouched(followup) {
+  const originTaskId = followup.recurrence_origin_id;
+  const originSubtasks = originTaskId
+    ? db.get().prepare('SELECT * FROM tasks WHERE parent_task_id = ?').all(originTaskId)
+    : [];
+  const currentSubtasks = db.get()
+    .prepare('SELECT * FROM tasks WHERE parent_task_id = ?')
+    .all(followup.id);
+
+  if (currentSubtasks.length !== originSubtasks.length) return true;
+
+  const originDueDate = originTaskId
+    ? db.get().prepare('SELECT due_date FROM tasks WHERE id = ?').get(originTaskId)?.due_date
+    : null;
+
+  for (const sub of currentSubtasks) {
+    if (sub.status !== 'open' || !sub.recurrence_origin_id) return true;
+    const origin = originSubtasks.find((o) => o.id === sub.recurrence_origin_id);
+    if (!origin) return true;
+
+    if (
+      sub.title !== origin.title ||
+      (sub.description || '') !== (origin.description || '') ||
+      sub.category !== origin.category ||
+      sub.priority !== origin.priority ||
+      sub.assigned_to !== origin.assigned_to ||
+      sub.points !== origin.points ||
+      sub.visibility !== origin.visibility ||
+      sub.due_time !== origin.due_time
+    ) {
+      return true;
+    }
+
+    const subAnchorDate = originDueDate || origin.due_date;
+    const expectedStart = shiftedStartDate(origin.start_date, subAnchorDate, followup.due_date) ?? origin.start_date;
+    const expectedDue = origin.due_date
+      ? (shiftedStartDate(origin.due_date, subAnchorDate, followup.due_date) ?? followup.due_date)
+      : null;
+
+    if (sub.start_date !== expectedStart || sub.due_date !== expectedDue) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+/**
  * Nimmt die Folgeinstanz zurück, wenn ein Abhaken rückgängig gemacht wird (#650).
  * Nur unangetastete Instanzen verschwinden: hat jemand sie selbst erledigt (und
- * damit die Serie weitergeschrieben) oder ihr Unteraufgaben gegeben, steckt dort
+ * damit die Serie weitergeschrieben) oder ihr Unteraufgaben gegeben/erledigt/bearbeitet, steckt dort
  * Arbeit, die ein Klick auf die Vorgängerin nicht wegwerfen darf.
  * Rückgabe: Anzahl vorgemerkter CalDAV-Löschungen.
  */
@@ -986,10 +1037,7 @@ function discardRecurrenceFollowup(taskId) {
   const followup = recurrenceFollowupOf(taskId);
   if (!followup || followup.status !== 'open') return 0;
 
-  const touched = db.get().prepare(
-    'SELECT 1 FROM tasks WHERE parent_task_id = ? LIMIT 1'
-  ).get(followup.id);
-  if (touched || recurrenceFollowupOf(followup.id)) return 0;
+  if (isFollowupSubtasksTouched(followup) || recurrenceFollowupOf(followup.id)) return 0;
 
   // Vor dem DELETE vormerken, wie in DELETE /:id: danach sind UID und Objekt-URL
   // weg. Lokal erzeugte Folgeinstanzen sind nicht gespiegelt, dann ist das ein No-op.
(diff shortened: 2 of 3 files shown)


ulsklyc:
Thanks for this, and thanks especially for the diagnosis in #742 itself - it was accurate and saved me the search. The approach is right: the subtask structure belongs to the series, and copying it at spawn time with a reset status is the correct place to do it. Your three tests pass here, and the branch still merges cleanly onto `main` after v2.8.1.

I reviewed it in a worktree and found three things. Two of them I turned into probes so you can reproduce them rather than take my word for it.

### 1. An edited subtask no longer protects the follow-up instance

The new `touched` check catches a subtask that was completed (`status != 'open'`) or manually added (`recurrence_origin_id IS NULL`), but not one that was *edited*. Unchecking the parent then deletes the follow-up along with the work someone typed into it. Before this PR every subtask was a guard, so this is a narrow regression rather than a pre-existing gap.

```js
test('an edited (not completed) subtask protects the follow-up', async () => {
  const parent = insertTask({
    title: 'Review A', status: 'open', due_date: dayKey(-1), created_by: uid,
    is_recurring: 1, recurrence_rule: 'FREQ=WEEKLY',
  });
  insertTask({ title: 'Schritt', status: 'open', due_date: dayKey(-1), created_by: uid, parent_task_id: parent });

  await call('PATCH', `/${parent}/status`, { status: 'done' });
  const followup = openInstances('Review A')[0];
  const sub = db.prepare('SELECT id FROM tasks WHERE parent_task_id = ?').get(followup.id);

  // The user works on the new instance: not ticking it off, refining it.
  await call('PUT', `/${sub.id}`, { title: 'Schritt: mit Essigreiniger' });
  await call('PATCH', `/${parent}/status`, { status: 'open' });

  assert.ok(db.prepare('SELECT title FROM tasks WHERE id = ?').get(sub.id),
    'the entered work must not disappear');
});
```

I do not have a strong opinion on the fix. Comparing the clone against its origin row (title, description, dates) would catch it; so would simply keeping the old "any subtask protects it" rule for anything the user has opened. Your call - the tradeoff is between a leftover empty follow-up and a deleted edit, and the second one is the expensive direction.

### 2. The two date fallbacks disagree when the series has no due date

```js
shiftedStartDate(sub.start_date, task.due_date, nextDate) ?? sub.start_date,
sub.due_date ? (shiftedStartDate(sub.due_date, task.due_date, nextDate) ?? nextDate) : null,
```

`shiftedStartDate` returns `null` when either date is missing, and a completion-anchored series (`recurrence_from_completion`, #658) legitimately runs without a `due_date`. The `due_date` branch then falls back to `nextDate`, which is right, but the `start_date` branch falls back to the *old* value: the fresh subtask arrives already overdue.

```js
test('completion-anchored series without a due date re-dates the subtask', async () => {
  const parent = insertTask({
    title: 'Review B', status: 'open', created_by: uid,
    is_recurring: 1, recurrence_rule: 'FREQ=WEEKLY', recurrence_from_completion: 1,
  });
  insertTask({
    title: 'Teilschritt', status: 'open', start_date: dayKey(-30), due_date: dayKey(-30),
    created_by: uid, parent_task_id: parent,
  });

  await call('PATCH', `/${parent}/status`, { status: 'done' });
  const followup = openInstances('Review B')[0];
  const sub = db.prepare('SELECT start_date, due_date FROM tasks WHERE parent_task_id = ?').get(followup.id);

  assert.notEqual(sub.start_date, dayKey(-30), `start_date of the new subtask: ${sub.start_date}`);
});
```

Measured on your branch the clone keeps `2026-07-15`, thirty days old, while its `due_date` correctly moves to the next occurrence.

### 3. Typo in a comment

`spawnRecurrenceFollowup`'s docblock now reads "Es gibt nichts **to** pushen" - looks like a stray search-and-replace, it was "zu pushen".

---

Nothing here is a reason to start over; items 1 and 2 are a handful of lines. If you would rather hand it back, say so and I will finish it on top of your commit with attribution - but the PR is yours and I would rather you keep it.

t0msh:
Thanks for the reply, I'm quite happy to get these bits and get it right.
Taken on board, and have implemented the below:

- **`isFollowupSubtasksTouched(followup)`**: Added helper to inspect all subtasks of a spawned follow-up task. It checks if subtask count changed, if any subtask status is not `'open'`, if `recurrence_origin_id` is missing (manually added subtask), or if any field (`title`, `description`, `category`, `priority`, `assigned_to`, `points`, `visibility`, `due_time`, `start_date`, `due_date`) was modified by the user.
- **`discardRecurrenceFollowup(taskId)`**: Uses `isFollowupSubtasksTouched(followup)` so that unchecking a parent task preserves the spawned follow-up task whenever a user has edited or refined its subtasks.
- **`spawnRecurrenceFollowup(task)`**:
  - Uses `const subAnchorDate = task.due_date || sub.due_date;` when calculating shifted dates for subtasks.
  - Ensures subtask `start_date` and `due_date` correctly shift for completion-anchored series (`recurrence_from_completion`) without a parent `due_date`.
- **Comment Fix**: Fixed stray comment typo `"Es gibt nichts to pushen"` -> `"Es gibt nichts zu pushen"`.

I also added the changes/fix to the changelog, as it seems I neglected to do that on the original PR.

ulsklyc:
Merged. Thank you for turning this around so precisely, and for taking the two findings as findings rather than as an argument.

I checked the new revision rather than taking the summary at face value, and both are properly fixed:

The edited-subtask case now holds. I re-ran the probe from my earlier comment unchanged and it passes, and you went further than I asked by comparing the actual fields instead of just the status. The date case holds too, with `subAnchorDate` falling back to the subtask's own due date, so a completion-anchored series without a parent due date no longer hands its copy a month-old `start_date`.

I also ran a third probe that was not in my original list, to check you had not overcorrected: whether an untouched follow-up still gets discarded when the parent is unchecked. It does. That was the risk with a comparison this thorough, and it is not there.

Because four releases landed between your first push and now, I tested the combination rather than your branch alone: your commits merge into `main` with no conflict in code, and the full suite is green on the merged state, 150 suites. The only conflict was in the CHANGELOG, where your entry and today's release blocks both wanted the same lines; your entry sits under `[Unreleased]` and will ship with the next release.

One thing I noticed and deliberately did not ask you to change: `isFollowupSubtasksTouched` compares the copy against its **origin** subtask, so editing the origin after the spawn also marks the follow-up as touched. That is the safe direction of the two, and a stray follow-up is much cheaper than a deleted edit, so it stays as you wrote it.

Thanks as well for the changelog entry, and for the diagnosis in #742 that started all this. It named the right function before I had opened the file.

PR #788, opened 2026-08-16
refactor: one name for shared expenses, and the README as a short landing page

ulsklyc:
Two related pieces of work, one commit each.

## `refactor(i18n)`: one name for the shared expenses module

The module carried several names in **every single language** - ten of the twenty-four locales carried three or more. In German it was "Gemeinsame Ausgaben" as the page title, "Geteilte Ausgaben" as the receipt folder in Documents, and "Aufteilung" in a nav key. The existing "one module, one name" guard could not see any of it, because it compares `nav.*` against `settings.apiTokenScopeModules.*` and the folder lives under `documents.*`.

`splitExpenses.title` is now canonical in all 24 languages. `documents.splitExpensesFolder` follows it (11 locales differed), `nav.splitExpenses` is removed rather than renamed (it existed in every locale and no line of code read it), and the two guest-account 403 messages plus `docs/SPEC.md` follow it too. Code identifiers stay `split-expenses`.

`splitExpenses.tabLabel` deliberately keeps its short form. It sits seventh in a tab strip whose other six labels are four to eleven characters long, so it is the one place a second wording is allowed - a label, not a name. SPEC.md now records that as a decision instead of leaving it to look like drift.

**Migration v146** renames an existing receipt folder. `ensureFolder` looks the folder up by NAME and creates it otherwise, so changing the locale value alone would have handed every existing household a second folder on its next receipt, with the old receipts left behind in the old one.

Both guards are counter-tested against the previous state. The migration test covers the case where old and new differ only in capitalisation (`id`: "Pengeluaran bersama"): `COLLATE NOCASE` finds the source folder as its own conflict there, so the migration has to exclude its own id. Mutation-tested - drop that comparison and exactly one assertion fails.

## `docs(readme)`: rebuild the README as a short landing page

486 and 487 lines down to 241 and 245. The README was a second landing page next to yuvomi.cloud, maintained by hand in two languages: 17.9 screens on a phone, the first install command at 11848px of scroll depth, 52% of the text folded behind `<details>` - including the four module handoffs, the one argument a folder full of separate self-hosted tools cannot match.

The gallery, the 1400-word module deep dive and the FAQ move to yuvomi.cloud and `docs/SPEC.md`. What moved in: the four handoffs (verbatim from the landing page's strings), the operating facts a self-hoster decides on (image size, RAM, ports, volumes, outbound - none of the five were anywhere in 486 lines), and a closing call to action.

**The install block is split in two.** It used to be one copyable block whose third line was a comment asking the reader to edit `.env` and whose fourth line started the container, and `openssl rand -hex 32` did not appear once in the whole file.

Five one-way drifts against the landing page are closed: the "live demo" its own target says does not exist, the single-person household PRODUCT.md has listed as a confirmed audience since 2026-08-06, Inventory shipping disabled by default, the eighteen icon alt texts that read every module name twice to a screen reader, and the headline.

Measured on the rendered markup at 390px and 1280px:

| | before | after |
|---|---|---|
| image payload | 4.85 MB / 46 requests | 65 KB / 1 WebP |
| external icon requests | 18 | 0 |
| desktop depth | 9.4 screens | 6.0 |
| mobile table overflows | 3 | 0 |

The module table loses its icon column: at 390px those icons rendered at 1.05 x 1.05 px, because GitHub's `img{max-width:100%}` resolves against a 28px column. The table itself stays the canonical module list.

`test:readme-consistency` holds what a reviewer cannot - install block never fuses the human step into the start, stated counts are the counted ones, every landing-page module name appears in the README table, EN and DE keep the same shape, no dead link, anchor or dash. Counter-tested: six of its assertions fail against the previous state.

## Notes

- `npm test` green on the branch
- Migration v146 makes this a **minor**, not a patch
- CHANGELOG entry follows in release prep


ulsklyc:
All six review findings addressed in e2780be8, each verified against the code first.

**CodeQL, `test/docs-dict.js`** - correct. `stripTags` cut in a single pass. Now a fixpoint loop, the same shape and the same reason as `test/source-text.js` already carries.

**P1, migration v146** - the fix is in, but the severity does not hold up, and I want that on the record rather than quietly adopted. With both spellings present, both lookups do return the same row, as described. What follows is not an abort: the index scan returns `Pengeluaran Bersama` *first* (BINARY collation, capital before lowercase), so `source` is the target row, and the UPDATE writes the name it already has. No `UNIQUE` violation, no rename, no startup failure. Measured directly:

```
vorher : [{"id":2,"name":"Pengeluaran Bersama"},{"id":1,"name":"Pengeluaran bersama"}]
AKTUELLE Fassung: lief durch
nachher: [{"id":2,"name":"Pengeluaran Bersama"},{"id":1,"name":"Pengeluaran bersama"}]
```

The old code was correct by accident. It is now `SELECT ... WHERE name = ?` for the source and `WHERE name = ? COLLATE NOCASE AND id <> ?` for the conflict, so the result no longer depends on a query plan's ordering. The added test says out loud that it is green against both versions and is therefore a state assertion, not a mutation probe - this repo has been bitten by guards that looked like proof and were not.

**P2, outbound claim** - correct, and worse than stated. `public/router.js` calls `/api/v1/changelog` on the first shell load and every six hours after, `server/routes/changelog.js` fetches `api.github.com/repos/ulsklyc/yuvomi/releases`, and there is no env var to turn it off. The claim was wrong on the landing page too, where it originated. Both READMEs and both dictionaries in `docs/index.html` now name that one request instead of promising zero.

**P2, single-file export** - correct. Qualified in both READMEs and both dictionaries, in the operating facts and under "Before you commit".

**P2, Podman path** - correct, and a regression I introduced while shortening. `podman compose -f podman-compose.yml up -d` with its `podman-compose.yml` is back, with the `:Z` labels named as the reason.

**P2, `#faq` anchor** - correct. `SUPPORT.md` now points at the section that answers those questions. The more useful part: `test:readme-consistency` checked the README's own anchors and was green while an inbound link went nowhere, so it grew that direction. Counter-tested - restore the old link and the new assertion fails.

`npm test` green.

PR #837, opened 2026-08-23
Multiple filipino translation improvements

anobongjimwel:
## Summary

This pull requests contains several changes on some key parts of the ```fil.json``` file of the project which improves Filipino translation to sound more native. Initially made by @ulsklyc, I further improved some things that can be improved.

Other than that, @ulsklyc nice work on starting the translation. Mostly correct indeed.

## Changes

- All commits are within the ```fil.json``` file


ulsklyc:
Thank you for this - it is a careful pass, and the parts I can check hold up: 214 values changed, every `{{placeholder}}` preserved, indentation and line endings untouched, and the file merges into `main` cleanly and comes out key-complete against the German reference.

Two things need a follow-up, both the same shape: a module was renamed in one place and left standing in the others.

**Blocking - `Imbentaryo`.** `nav.inventory` is now `"Imbentaryo"`, but `settings.apiTokenScopeModules.inventory` is still `"Inventory"`. That trips a guard (`npm run test:i18n`, "Modulnamen sind in nav und in den API-Token-Scopes wortgleich"), which is why CI would not go green.

The guard only sees that one pair, but three more places carry the same name and were missed: `inventory.title`, `shortcuts.goInventory` and `documents.inventoryFolder`. Until now that name was identical in all four places in every one of the 24 locales, so this would have been the first split.

`documents.inventoryFolder` is the one that matters beyond looks. It is the folder receipts for inventory items are filed under, and `ensureFolder` looks it up **by name** - a different name silently creates a second folder and leaves the older receipts in the first. We had to clean exactly that up once before, in migration v146, for the shared-expenses folder.

**Worth fixing - `Paminggalan`.** You renamed the module from `Despensa` to `Paminggalan` and updated nine strings, but seven still say `despensa`: `pantry.emptyTitle`, `pantry.searchPlaceholder`, `pantry.loadErrorTitle`, `shopping.toPantryDescription` (+`_one`) and two `emptyHint` lines. The app would call the same thing by two names.

I have prepared both follow-ups on top of your commits rather than sending you back and forth over mechanical replacements, and extended the folder guard so the inventory folder cannot drift again. Your translation work is unchanged.

One question I genuinely cannot judge, so I would rather ask than overrule: `common.visibility.label` goes from `"Pagiging nakikita"` to `"Nakakakita"`. It labels a selector for who may see an entry, so I would expect a noun there ("visibility") rather than something that reads to me like "can see". Is that the natural phrasing in context?

Also worth knowing: the whole `inventory.*` section is still untranslated English, and it is not the only one - roughly 3800 values across the locales have never been translated. If you want to keep going, that section is a self-contained next step.

PR #840, opened 2026-08-23
fix(sync): resolve Google Calendar sync FOREIGN KEY crash

### LINKED ISSUE #839: [Bug] Google Calendar Sync not working
### Yuvomi version

2.32.0

### Deployment

Bare metal / from source

### Description

When the user with id 1 is deleted the Google sync does not import any events.

The error stems from [L940 in google-calendar.js](https://github.com/ulsklyc/yuvomi/blob/main/server/services/google-calendar.js#L940) as far as I can see the user id is harcoded to 1 in this.

### Steps to reproduce

1. Delete the default user from installation
2. Setup Google Calendar Sync
3. Try to sync

Expected: Sync works

Actual: No sync is happening and log shows error message in this style:
`{"ts":"2026-08-23T13:21:09.040Z","level":"error","mod":"Google","msg":"Upsert error for event xxxx:","extra":"FOREIGN KEY constraint failed"}
`

### Browser / device / OS

_No response_

### Logs

```shell

```

### Additional context

_No response_

Ennosuke:
## Summary

Google Calendar sync was not working when user with id 1 was deleted. Same as #4 

Fixes #839

## Changes

- Instead of harcoded 1 for "created_by" field use logic from iCloud fix

## Checklist

- [x] `npm test` passes

### DIFF
--- a/server/services/google-calendar.js
+++ b/server/services/google-calendar.js
@@ -888,6 +888,14 @@ function upsertGoogleEvents(items, calRefId = null, calColor = GOOGLE_COLOR, col
     // sonst Kalenderfarbe als Default.
     const evColor = (item.colorId && colorMap[item.colorId]) || calColor;
 
+    // created_by: ersten existierenden User verwenden (nicht hardcoded ID 1)
+    const owner = db.get().prepare('SELECT id FROM users ORDER BY id ASC LIMIT 1').get();
+    if (!owner) {
+      log.warn('No user in database - sync skipped.');
+      return;
+    }
+    const createdBy = owner.id;
+
     const existing = db.get().prepare(
       'SELECT id, outbound_dirty FROM calendar_events WHERE external_calendar_id = ? AND external_source = ?'
     ).get(item.id, 'google');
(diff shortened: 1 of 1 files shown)


ulsklyc:
Merged as 8256208e with a follow-up on top - thank you, the diagnosis was exactly right and the pointer to the iCloud fix saved me the search.

Two things I changed after taking it:

**The lookup moved out of the per-event callback.** It sat next to the color resolution, which runs once per event, so a full resync over a few thousand appointments would have prepared and run `SELECT id FROM users ORDER BY id ASC LIMIT 1` a few thousand times. `caldav-sync.js` resolves the same value once per run for that reason, and it does now here too - the row cannot change mid-sync.

**The early `return` also caught updates and deletions.** Placed before the `existing` branch, a database without users would have stopped those as well, and neither of them touches `created_by`. Now only the insert is skipped, and the warning is logged once when the owner is resolved rather than once per event.

**The suite could not have caught this.** Its seed user gets ID 1 - the comment above it even said `created_by = 1 in upsertGoogleEvents`, so the assumption that broke was written down and carried along. Two cases are in now: one owner with a different ID, one with no users at all. I checked both against the unfixed code and both go red, the first with the same `FOREIGN KEY constraint failed` from your report.

One note on the commit subject: it said "and sync all calendars", but the diff only contained the `created_by` change. I dropped that half when merging.

Ships in the next release.

PR #886, opened 2026-08-26
feat(schedule): add opt-in rotating shift patterns and read-only calendar overlays

mclgoerg:
## Summary

Adds a `schedule` module, per the data model posted in #786: households can define reusable shift types, assign them to cycle days (an anchor date + cycle length, covering both a fixed weekly timetable and an arbitrary rotation like Early/Early/Late/Late/Night/Night/Off/Off), and replace individual dates with a different shift or an explicit free day.

Follows the three architectural decisions from the discussion:

- **One cycle model**, not two features - a "Week A/Week B" timetable is just a 14-day cycle, so it and a rotating shift pattern share the same arithmetic instead of duplicating it.
- **A pattern, not calendar recurrence.** A rotation is a repeating sequence of *different* entries, which an RRULE can't express without splitting it into several unrelated series.
- **An overlay, not materialized events.** Entries are computed from anchor date + cycle position + overrides at read time; nothing is copied into `calendar_events`, so editing a pattern can't leave stale appointments behind, and a two-year rotation costs one row instead of ~700.

Resolution priority per day: an override wins over the newest applicable pattern (by `valid_from`) wins over nothing. A `NULL` override is an explicit free day. Every household member can read the overlay (per the discussion's "the family mostly needs to know Anna is unavailable Tuesday evening" framing); a member can write only their own schedule, and admins can write for anyone.

The calendar shows the overlay as a separate, explicitly toggleable, read-only layer - never as ordinary editable events - defaulting to a compact strip rather than a full block, matching the "quieter" signal preference from the discussion. The module ships disabled by default, the same way Inventory does.

## Discovered along the way

Rebasing onto `main` (which had moved from v2.43.0 to v2.46.0 while this was in progress) surfaced a real regression, unrelated to this feature: an earlier edit to `server/routes/preferences.js` had accidentally deleted the `MAX_WIDGET_OPTION_LENGTH` constant while adding `schedule` to the toggleable-modules list, breaking dashboard widget option validation (`ReferenceError`, caught by `test:preferences-dashboard`). Restored it as part of this PR since it was collateral damage from the same diff.

Also updated `test-inventory-default-off-migration.js`'s first case, which asserted the exact list of modules a fresh install ships disabled (`['inventory']`) against the fully-migrated global `db` singleton - now that `schedule` joins that same default-off list, the assertion is scoped to migration 145's own isolated effect instead (matching the file's other two test cases), so it won't need updating again for the next module that ships switched off.

## Test plan

- [x] `npm test` - full suite, 216/216 suites green, run after the rebase onto `main` @ v2.46.0
- [x] Migration renumbered 161 → 165 to sit after the four migrations (161-164) that landed upstream while this was in progress; `test:migrations-append-only` and `test:schema-mirror` pass
- [x] Manual verification in a running instance: shift type / pattern / override CRUD, cycle-day assignment, override precedence, calendar overlay toggle (compact + full-block), statistics tab, mobile layout, RTL/reduced-motion paths for the new statistics bar chart
- [x] All 24 locales carry the new i18n keys (`test:i18n`, `test:i18n-plural`); only `en` and `de` have real translations for the feature so far, matching how the rest of the module ships today - the other 22 carry English placeholder text, same as every other string in this feature

Closes discussion #786.

---
Drafted with Claude Code.


ulsklyc:
Thank you for this - it is a careful piece of work, and the architecture is exactly what we agreed in #786: one cycle model, an overlay rather than materialized events, opt-in. I have not changed any of that.

Rather than hand you a list and wait, I rebased your commit onto v2.47.0 and appended the review findings as separate commits: **#895**. Your commit is unchanged and still carries your authorship.

For the record, what the follow-ups cover:

1. **`POST`/`PUT`/`DELETE /shift-types` had no permission check** while every other route in the module asks `mineOrAdmin`. Reproduced: another member renames the household's early shift (200) and deletes an unused one (204). Your PR description promises "a member can write only their own schedule" - the rule now follows the `created_by` column you already added: anyone may add a type, only its creator or an admin may change or remove it.
2. **`catch { 409 'in use' }`** caught every error and blamed the same cause. Pulled out as `isStillReferenced()` so a test can ask the rule directly. Measured along the way: a refused `ON DELETE RESTRICT` arrives as `SQLITE_CONSTRAINT_TRIGGER`, not `_FOREIGNKEY`, even though its message says "FOREIGN KEY constraint failed".
3. **Two `innerHTML` assignments** in `public/pages/schedule.js`. Not your fault that no test caught it - the guard was a list of seven files and structurally could not see a new page. It now scans all of `public/` (#894).
4. **`style="--layer-color:#6C3AED"`** in the calendar markup, although your PR adds `--module-schedule` to tokens.css. The holiday layer next to it carries an inline value because its colour is a *user setting*; this one is a design decision.
5. **Documentation.** The module was in neither canonical source - not the README table, not `docs/SPEC.md`. On the site the module count is derived from the actual cards, so that meant writing the card too.

One thing worth passing on for next time: only CHANGELOG and package.json conflicted, but the `test` line in your package.json predates `test:caldav-object-urls` (added in v2.47.0). Taking your side of that conflict silently drops the suite from the aggregate.

I'll close this once #895 lands. Thanks again - this was a lot of ground covered, and the parts that needed changing were the edges, not the design.

mclgoerg:
Thank you so much for the feedback. I’ll take care of this once I contribute again.

ulsklyc:
Landed as **#895**, released in **v2.48.0** - your commit is in `main` unchanged and carries your authorship.

Closing this one so the two do not sit side by side. The module is the 19th in the table and is documented in `docs/SPEC.md`, the README tables and the OpenAPI spec.

Thanks again - the architecture held up under review, which is the part that is hard to get right. Whenever you come back to it, the follow-up commits in #895 are the whole list; nothing is waiting on you.

PR #919, opened 2026-08-28
feat: give third-party modules the same surfaces as core (widgets, permissions, i18n)

Cbrogg:
## Summary
Bring third-party drop-in modules to parity with core modules: dashboard widgets, household permissions, API token / MCP scopes, and UI translations — without changing core application code for each new module.

The optional sidecar API proxy is **not** in this PR. It is parked on `feature/ext_proxy` for a separate review.

Docs: `MODULES.md`. Changelog: `[Unreleased]`.

## Changes
- Validate optional `capabilities` in `module.json` (widgets, `ext:` permissions, API prefix) and expose normalized `capabilities` + `i18n` on `GET /api/v1/modules`.
- Load third-party dashboard widgets at runtime (`renderWidget`) with per-widget error isolation and optional `optionsSchema` dialogs.
- Merge extension entries into navigation, route guards, the permission catalog, and API token / MCP scope pickers (`ext:`, `:`).
- Resolve module UI labels from `locales/{locale}.json` via UI locale → `defaultLocale` → `en` → `de` → static manifest strings.
- Keep `permissions.js` off `db.js` by injecting the extension catalog at runtime (same pattern as scopes); precache `extension-i18n` / `extension-widgets`.
- Document the contract in OpenAPI and `MODULES.md`; add suites for capabilities, permissions, widgets, and i18n.

## Checklist
- [x] `npm test` passes
- [x] Follows `CONTRIBUTING.md` conventions
- [x] `CHANGELOG.md` updated (if user-facing change)


ulsklyc:
Thanks for the amount of work in here, and for asking first in #922 rather than dropping 2k lines unannounced. The manifest side is genuinely well built: `capabilities.widgets[].entry` is validated as a safe relative path that must exist on disk (`server/services/module-capabilities.js:127-133`), so a module cannot pull code from a CDN. That is exactly the constraint this project has, and you found it without being told.

The proxy is a different story. I ran the branch and confirmed four issues below; each one is on its own enough to block a merge. All four are in the same file pair, so they are fixable, but they need to be fixed before this can go in.

## Blocking

### 1. The CSRF bypass makes every proxied write forgeable

`server/middleware/csrf.js:33`

```js
if (req.path.startsWith('/extensions/')) return next();
```

`csrfMiddleware` is mounted as `app.use('/api/v1', csrfMiddleware)` (`server/index.js:459`), so inside it `req.path` is `/extensions/<id>/...` and this branch does fire.

The comment says sidecars "enforce their own CSRF on upstream", but the sidecar has no CSRF relationship with the browser: it never issued a token, it cannot validate one, and by the time it is called Yuvomi has already authenticated the request from the session cookie and authorised it against the permission catalog. Any page on the internet can now run

```js
fetch('https://household.example/api/v1/extensions/budget-v2/transactions',
      { method: 'POST', credentials: 'include', body: '...' })
```

and the browser attaches the session cookie. This is the one thing CSRF protection exists to stop, and the exemption is the only path under `/api/v1` that has it.

The other exemption in that function, `req.authMethod === 'api_token'`, is safe for a reason that does not transfer: a bearer token is not attached by the browser automatically. Cookie-authenticated requests have no such property.

There is no case for this line. If a sidecar cannot read `X-CSRF-Token` from a forwarded header, that is a sidecar problem, not a reason to open the door.

### 2. The user's Yuvomi session cookie is handed to the sidecar

`server/routes/extensions-proxy.js` (`forwardHeaders`)

```js
const headers = { ...req.headers };
delete headers.host;
delete headers.connection;
delete headers['content-length'];
```

`cookie` and `authorization` survive that. Every proxied request therefore gives the sidecar the caller's full Yuvomi session cookie, or their API token. A sidecar is third-party software the operator dropped into a folder; with the session cookie it can call any Yuvomi endpoint as that user, including admin ones if the user is an admin. That turns "this module may be reached" into "this module holds a full power of attorney for whoever uses it", which no permission entry in the catalog expresses.

Forward an allowlist instead of stripping a denylist: `content-type`, `accept`, `accept-language`, and an identity header Yuvomi mints itself (user id, display name, granted access level). The sidecar should learn who is calling from Yuvomi, not by holding their credentials.

### 3. Response headers from the sidecar land on the Yuvomi origin

`server/routes/extensions-proxy.js` (`applyUpstreamHeaders`) copies every upstream header except `transfer-encoding`, including `set-cookie` via `res.append`. The sidecar can therefore set or overwrite cookies on the Yuvomi origin: the session cookie, the CSRF cookie, anything. Session fixation through a module is not a scenario that needs a malicious sidecar, only a careless one.

The `_rawHeaders` field added to `server/utils/http.js:81` exists solely to make this copy possible, and should go with it. Allowlist the response headers as well: `content-type`, `content-disposition`, `cache-control`, `etag`.

### 4. `rest` is not confined to the namespace the route promises

```js
const targetUrl = `${upstream}/api/extensions/${encodeURIComponent(moduleId)}/${rest}${query}`;
```

`moduleId` is encoded, `rest` is not. Verified:

```
rest = '../../admin/secret'
  ->  https://sidecar.example/api/extensions/x/../../admin/secret
  ->  https://sidecar.example/api/admin/secret   (after URL normalisation in fetch)
```

Any authenticated member with access to one extension module can reach any path on that sidecar, including paths the sidecar considers internal because "only Yuvomi's `/api/extensions/` prefix is exposed". Reject any `rest` segment that is `.` or `..`, and encode each segment individually.

## Should fix

**`serializeBody` silently drops non-JSON bodies.** Anything that is not JSON, a string or a Buffer returns `undefined`, so a `multipart/form-data` upload arrives at the sidecar with an empty body and a `content-type` header that promises otherwise. That fails as a puzzle rather than as an error. Either stream the raw body through or reject unsupported content types with a 415.

**The proxy suite tests configuration, not the security properties.** `test/test-extensions-proxy.js` covers `normalizeUpstreamUrl`, `checkSSRF` and the 404/502 paths, all of which pass. None of the four issues above would have been caught by it, and all four are testable: a forged cross-origin POST, an assertion that `cookie` is absent from forwarded headers, an assertion that `set-cookie` does not survive the response, and the traversal string above.

**The branch is behind main and `test:changelog` is red on it** (`actual: ['2.51.2'], expected: []`). A rebase fixes it.

## On the direction

That is a separate question from the code, and mine to answer rather than yours to fix. The manifest-driven half (widgets, permissions, scopes, i18n) extends surfaces Yuvomi already owns, and it does so along the lines the project already draws. The sidecar proxy is a different commitment: it makes Yuvomi a gateway in front of software it does not ship, and the four issues above are the sort of thing a gateway has to get right permanently, not once.

I would rather take this in two parts, with the manifest half going first and the proxy discussed on its own. If you would like to split the branch that way, I will review the first part promptly.

Cbrogg:
Thanks for running the branch and for the precise write-up. All four blockers are real; I am not going to argue any of them. I also agree with the split: the manifest half belongs in this PR, the proxy is a gateway and should be reviewed as one.

This branch is now the manifest half only (widgets, permissions, scopes, i18n, catalog, docs), rebased onto current main so `test:changelog` is green. The proxy work is parked on `feature/ext_proxy` with the four issues fixed there, and I will open it as a follow-up PR rather than mix it back in.

## Blocking

### 1. CSRF bypass
Agreed — there is no case for that line. Sidecar CSRF has no relationship with the browser, and Yuvomi has already authenticated the session cookie. The `/extensions/` exemption is gone from this PR (the file is unchanged from main). Cookie-authenticated writes under `/api/v1` go through the same double-submit check as every other route. The `api_token` exemption stays, for the reason you gave.

### 2. Session cookie forwarded to the sidecar
Agreed. On the follow-up, `forwardHeaders` is an allowlist: `content-type`, `accept`, `accept-language`, plus identity headers Yuvomi mints (`X-Yuvomi-User-Id`, display name, role, granted access level) and an optional shared secret. `cookie`, `authorization`, and the other credential headers are not forwarded. The sidecar learns who is calling from Yuvomi, not by holding the session.

### 3. Upstream `Set-Cookie` on the Yuvomi origin
Agreed. Response headers on the follow-up are allowlisted to `content-type`, `content-disposition`, `cache-control`, `etag`. `set-cookie` is not copied. `_rawHeaders` on `safeRequest` existed only for that copy and is gone with it.

### 4. `rest` path traversal
Agreed, and thank you for verifying it. The follow-up rejects any `.` / `..` / empty segment with 400 and `encodeURIComponent`s each remaining segment so the request cannot leave `/api/extensions/:moduleId/`.

## Should fix
- Unsupported bodies are 415 rather than a silent empty upstream body.
- The proxy suite now asserts the four properties above (forged cookie POST without CSRF, no `cookie` on the outbound request, no `set-cookie` on the Yuvomi response, traversal string rejected). That stays in the proxy PR.
- This branch is rebased onto current main.

Happy to adjust the identity-header names if you have a preferred prefix. I would like the first part reviewed as soon as you have a moment.

Cbrogg:
@ulsklyc Changes have been made to the code; please review it again.

ulsklyc:
Second pass on the manifest half. The split was the right call, and the four blockers from the first round are gone - `server/middleware/csrf.js` is untouched in this diff, and the proxy is off the branch. The permission and scope injection reads well: keeping `permissions.js` off `db.js` by pushing the catalog in at runtime is the same shape the scopes already had, and `setModuleEnabled` re-runs `listModules` so a disabled module drops out of both catalogs immediately. `capabilities.widgets[].entry` is still validated as a safe relative path that must exist on disk, and `ID_RE` keeps `ext:{id}` safe as an attribute value.

Two blockers, both reproduced on this branch. Neither is caught by the three new suites - all of `test:extension-permissions`, `test:extension-widgets` and `test:extension-i18n` are green with both defects in place.

## Blocking

### 1. An extension `api.prefix` can take over a core module's scope

`server/services/module-capabilities.js:213-216` builds the scope prefix by stripping `/api/` off whatever the manifest declared, and `server/scopes.js:32-40` merges extension modules into `PREFIX_TO_MODULE` **after** the core ones. `Map.set` overwrites, so a manifest that declares `"api": { "prefix": "/api/tasks" }` reassigns the segment `tasks` to `ext:taskextras` - and `moduleForPath` is what the token-scope guard in `server/index.js:431` asks about every `/api/v1` request.

```
$ node --input-type=module -e "
import { setExtensionScopeModules, moduleForPath, tokenAllows } from './server/scopes.js';
import { buildExtensionCatalog } from './server/services/module-capabilities.js';
const cat = buildExtensionCatalog([{ id: 'taskextras', status: 'enabled', enabled: true,
  capabilities: { permissionModuleKey: 'ext:taskextras',
    permissionModule: { label: 'Task Extras', icon: 'box' },
    widgets: [], apiPrefix: '/api/tasks', scopeKey: 'ext:taskextras' } }]);
console.log('before', moduleForPath('/tasks/5'));
setExtensionScopeModules(cat.scopeModules);
console.log('after ', moduleForPath('/tasks/5'));
console.log('tasks:write allowed?          ', tokenAllows(['tasks:write'], moduleForPath('/tasks/5'), 'write'));
console.log('ext:taskextras:write allowed? ', tokenAllows(['ext:taskextras:write'], moduleForPath('/tasks/5'), 'write'));
"
before tasks
after  ext:taskextras
tasks:write allowed?           false
ext:taskextras:write allowed?  true
```

So an installed module grants any token holding `ext:taskextras:write` full write access to the **core** Tasks API, and silently revokes it from tokens that hold `tasks:write`. `/api/tasks` is not an exotic prefix either - a module that extends tasks is exactly the one that would pick it, so this breaks by accident before anyone tries it on purpose.

The fix is already half-written: `moduleForPath` has a dedicated `extensions/{id}` branch (`server/scopes.js:140-144`), and `MODULES.md` documents the prefix as `/api/extensions/my-module`. The code just does not require it. Make the validation match the docs - reject anything that is not `/api/extensions/{moduleId}` - and the collision cannot be expressed at all. Anchoring it to the module's own id also removes the second question, which is what should happen when two modules claim the same prefix.

### 2. A module's locale file can poison `Object.prototype`

`public/i18n.js`, `nestFlatLocaleDict`:

```js
node[parts[i]] = node[parts[i]] || {};
node = node[parts[i]];
```

For `parts[i] === '__proto__'` the left side is truthy (it *is* `Object.prototype`), so nothing is created - and the walk then continues **into** the prototype, where the last segment is written:

```
$ node --input-type=module -e "
import { nestFlatLocaleDict } from './public/i18n.js';
nestFlatLocaleDict({ '__proto__.polluted': 'PWNED', 'widget.title': 'harmless' });
console.log(({}).polluted);
"
PWNED
```

The input is `locales/{code}.json` from a module folder, fetched by `loadExtensionLocales` and parsed with no key filtering. What makes this worth blocking rather than shrugging at - a widget module runs its own code anyway - is *when* it happens: `reloadExtensionLocales` loads the bundles of every enabled module at startup, before any widget mounts and whether or not the user ever opens that module's page. A module that ships nothing but a menu entry and a translation file gets to write onto `Object.prototype` on every page of the app, and this codebase leans on `?? 'write'` / `?? 'allow'` defaults in exactly the places that a polluted prototype turns inside out (`public/permissions.js`, `moduleAccess`, `canSeeWidget`).

`Object.create(null)` for the nodes plus skipping `__proto__` / `constructor` / `prototype` segments is a two-line fix; `resolve()` walks it the same either way.

## Should fix

### 3. Comments carrying a decision were deleted along the way

`public/permissions.js` loses the "Fail-open by design" paragraph from the module header, the note above `NAV_TO_MODULE` that it must stay in step with `PERMISSION_MODULES.navIds`, and the JSDoc on all five exported functions. None of that is what this PR changes. The fail-open paragraph in particular is the answer to the first question a reader of this file has - "why is a permission check allowed to default to open?" - and the next person to touch it will have to re-derive it. Please restore them; if the fail-open reasoning has actually shifted because extensions now feed the same maps, say so there instead of removing it.

### 4. The new suites test the happy path only

Three suites, 0 failures, both defects above live in the diff. Whatever the fixes end up being, please anchor them: a manifest whose `api.prefix` collides with a core prefix must be rejected (or provably not reach `PREFIX_TO_MODULE`), and a locale file containing a `__proto__` segment must leave `Object.prototype` untouched. As a check on the guard itself, disable the fix and confirm the test goes red - a rejection test that passes for the wrong reason is the failure mode here.

## Nice to have

`public/utils/extension-widgets.js`, `defaultInsertIndex` returns `ordered.length` where its original in `dashboard-widgets.js` returns `0`, and the original carries a long comment about why appending was the more expensive answer (the A1-03 regression: an appended id makes every stored layout read as user-ordered, and the grid drops out of `dense`). Measured across every id, the two functions agree today - the only id with no predecessor is `tasks`, which is hidden by default and therefore not part of the comparison `isUserOrderedConfig` makes. So this is not a bug right now; it is a copy that dropped the reasoning, and it will stop agreeing the day a visible widget goes first in the list. Either match the original's `0` or note in place why the difference is safe.

ulsklyc:
Heads-up: #931 landed on `main` and this branch now shows as conflicting - `public/pages/dashboard.js` on both sides. Worth folding into the same pass as the two blockers rather than rebasing twice.

Cbrogg:
@ulsklyc Second-pass items addressed on this branch. Reproduced both blockers, then confirmed they no longer hold.

**Blocking**
1. `capabilities.api.prefix` is now required to be exactly `/api/extensions/{moduleId}`. A colliding prefix such as `/api/tasks` is rejected at normalize time; the module loads as `error` and never reaches `PREFIX_TO_MODULE`. `moduleForPath('/tasks/5')` stays `tasks`.
2. `nestFlatLocaleDict` builds with `Object.create(null)` and skips `__proto__` / `constructor` / `prototype` segments. The locale-file repro no longer writes onto `Object.prototype`.

**Should-fix / nice-to-have**
3. Restored the fail-open header, the `NAV_TO_MODULE` note, and the JSDoc on the exported helpers. `third-party-{id}` is gated only when `permissionModuleKey` is injected via `setExtensionNavMap`.
4. New tests cover the colliding prefix, the `__proto__` locale key, and insertion of a missing first widget id at 0. Disabling each guard turns the matching test red.
5. `defaultInsertIndex` matches core (`return 0`) with the A1-03 note.

`test:extension-permissions`, `test:extension-i18n`, and `test:extension-widgets` are green. Ready for another look when you have a moment.

ulsklyc:
Second pass done. Both original blockers are genuinely closed - I re-ran the counter-checks rather than trusting the suites:

- **Blocker 1 (colliding prefix):** neutralising `if (prefix !== expected)` in `module-capabilities.js` turns `a colliding api.prefix never reaches PREFIX_TO_MODULE` red (`'enabled' !== 'error'`). The guard is load-bearing. The compound lookup in `moduleForPath` handles the two-segment prefix correctly.
- **Blocker 2 (prototype pollution):** reverting `Object.create(null)` to `{}` turns `nestFlatLocaleDict does not poison Object.prototype` red. Confirmed closed.

Full suite: 4765 tests, 0 failures.

Three new findings, all reproduced.

## Blocking

### 1. A single failed `/modules` request permanently deletes the user's extension widget layout

`syncThirdPartyModules()` (`public/router.js:924-928`) swallows any error and sets `_thirdPartyModules = []`, then calls `setExtensionModules([])`. `normalizeDashboardConfigWithExtensions` filters against `knownIds`, so every `ext` widget id drops out of the config. The next layout write (`api.put('/preferences', { dashboard_widgets: widgetConfig })`, `dashboard.js:4013/4034/4110`) persists that stripped list.

```
[A] /modules loaded    -> akahu:balance = {"id":"akahu:balance","visible":true,"order":1,"size":"2x2","options":{"account":"savings"}}
[B] /modules failed    -> akahu:balance = undefined
```

This is not just a startup window. `startThirdPartyModulePolling()` re-runs `syncThirdPartyModules()` every 30 s while the dashboard is open, so one hiccup - a server restart during an update, a transient network error, or the `/api/` rate limit (300/min per IP) - is enough. When `/modules` recovers, the widget returns as a newcomer: default size, default position, **options gone**.

Two changes close it:
- In `syncThirdPartyModules`, keep the previous list on failure instead of clearing it. An empty list is a claim ("no modules exist"), and a failed request cannot support it.
- In `normalizeDashboardConfigWithExtensions`, carry unknown `ext:*` entries through instead of dropping them. A disabled module should get its layout back on re-enable, and that is the same code path.

## Should fix

### 2. Extension permission rules are silently discarded until the first registry scan finishes

`server/index.js` starts the scan *inside* the `app.listen` callback, so the socket accepts requests before `_extensionCatalog` is populated. `resolvePermissions` iterates `allPermissionModules()`; while the catalog is empty, `MODULE_KEY_SET.has('ext:demo')` is false and the stored `ext:demo -> none` row is dropped on the floor. Since `moduleAccessVerdict` is a deny-list, a missing key means allow:

```
[A] catalog not yet loaded:
  modules['ext:demo'] = undefined
  sessionMap          = null
  VERDICT (POST)      = allow
[B] catalog loaded:
  modules['ext:demo'] = "none"
  sessionMap          = {"ext:demo":"none"}
  VERDICT (POST)      = none
```

The blast radius today is limited - there is no `/api/extensions/*` router in this server, so the only observable effect is nav/widget visibility, not data. But `capabilities.api.prefix` is shipping as a promise about that path, and the failure mode is the wrong direction. It also is not strictly a startup race: the `listModules()` call at `index.js:590` only logs a warning on failure, so if that first scan throws, the catalog stays empty until someone happens to hit `GET /api/v1/modules`.

`await listModules({ admin: true })` before `app.listen()` closes it completely, costs one filesystem scan at boot, and removes the ordering dependency that `/permissions/catalog` currently works around with its own `await`.

### 3. The `DANGEROUS` key filter is not covered - only `Object.create(null)` is

Emptying the set to `new Set([])` leaves `test:extension-i18n` fully green (10/10). The prototype test passes purely on `Object.create(null)`. That is fine for the pollution property itself, but the filter is currently unverified, so it can be removed later without anything going red.

Related, and the reason the filter is worth keeping: `extensionLocaleStore` (`public/i18n.js:28`) is a plain `{}`, and `resolveExtensionTranslation` does `extensionLocaleStore[moduleId]` with no own-property check. A module id of `constructor` passes `ID_RE` (`/^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$/`), and then:

```
normal     : Weather
constructor: THROWS TypeError: Cannot read properties of undefined (reading 'de')
toString   : THROWS TypeError: Cannot read properties of undefined (reading 'de')
__proto__  : THROWS TypeError: Cannot read properties of undefined (reading 'de')
```

`t()` throwing is worse than `t()` returning the key. `extensionLocaleStore = Object.create(null)` plus a test that pins the filter would cover both.

## Nice to have

### 4. Wrong i18n key on the empty extension options dialog

`dashboard.js` falls back to `t('dashboard.optionTaskCategoriesEmpty')` when a widget's `optionsSchema` is empty. For a third-party widget that text is about task categories and simply does not apply. Worth its own key.

---

Once 1 and 2 are in I am happy to merge this. #929 can then rebase onto it as you planned.

Cbrogg:
@ulsklyc Third-pass items addressed on this branch. Reproduced all four, then confirmed they no longer hold.

**Blocking**
1. A failed `GET /modules` no longer wipes the layout. `syncThirdPartyModules` keeps the previous list on error (an empty list is only accepted from a successful body). `normalizeDashboardConfigWithExtensions` carries unknown `<module>:<widget>` entries through — size, position, and options survive a disabled module and a recovered catalog.

**Should fix**
2. `await listModules({ admin: true })` now runs before `app.listen`. The socket does not accept requests with an empty extension catalog, so a stored `ext:* → none` row is no longer dropped into deny-list allow.
3. `extensionLocaleStore` is `Object.create(null)`. `t()` no longer throws for module ids that collide with `Object.prototype` (`constructor`, `toString`). A new own-keys assertion pins the `DANGEROUS` filter — emptying the set turns it red. The existing prototype-pollution test still passes on `Object.create(null)` alone.

**Nice to have**
4. The empty extension-options dialog uses `dashboard.optionExtensionEmpty` in all 24 locales. The task-categories copy stays on the tasks widget only.

`test:extension-widgets`, `test:extension-permissions`, `test:extension-i18n`, and `test:i18n` are green. Disabling each guard turns the matching test red.

ulsklyc:
Merge order for the three open PRs is now fixed: **#919 first, then #929, then #930.**

This one is the head of the queue, so nothing else will land under you - the four findings from the review above are the only gate. #929 stays where you put it and rebases onto this once it is in.

For visibility: #930 (schedule planner) is last and touches `dashboard.js` / `dashboard-widgets.js`, so it will rebase after both of yours rather than the other way round.

ulsklyc:
Fourth pass done. All four findings from the previous round are genuinely closed - I re-ran the counter-checks on each rather than trusting the suites:

- **Finding 1 (layout wipe):** turning `if (!ok) return prev` back into `return []` in `selectThirdPartyModuleList` makes the widgets suite red (1 fail); dropping `|| isExtensionWidget(w.id)` from the normalize filter makes it red twice. Size, position and options survive a failed `/modules` and a disabled catalog.
- **Finding 2 (scan before listen):** `await listModules({ admin: true })` sits before `app.listen`, wrapped in try/catch so a failing scan degrades to a warning instead of blocking boot - the right trade. The source guard in the permissions suite pins the ordering.
- **Finding 3 (locale store):** reverting either `Object.create(null)` site turns the i18n suite red, and emptying `DANGEROUS` turns the new own-keys assertion red. The filter is load-bearing now.
- **Finding 4 (i18n key):** `dashboard.optionExtensionEmpty` is present in all 24 locales; putting the task-categories key back turns the widgets suite red.

Two new blockers, both in the harness rather than the feature code - which is exactly why they matter: they decide whether any of the guards above ever run again.

## Blocking

### 1. The three new suites never run - the npm scripts they are documented under do not exist

`docs/test-suites.md` documents `npm run test:extension-permissions` / `test:extension-widgets` / `test:extension-i18n`, but `package.json` registers no such scripts, and the `test` chain does not include the three files either. CI runs `npm test`, so every guard this PR adds is dead on arrival - my counter-checks only went red because I ran the files by hand (`node --test test/test-extension-widgets.js`). Please register the three scripts and append them to the `test` chain; next to `test:modules` is the natural spot.

### 2. `npm test` cannot pass on this branch: package.json references a test file the branch does not have

The branch's `package.json` says 2.56.0 and includes `test:shopping-send`, but `test/test-shopping-send.js` does not exist on this branch - the chain dies right there (`Could not find 'test/test-shopping-send.js'`, exit 1). The cause: the two `merge(main)` commits are hand-applied conflict resolutions, not merges. The merge base of this branch is still v2.54.0, so none of main's v2.55-v2.57 commits are actually on it - the branch carries main's package.json without main's files. A real `git merge origin/main` fixes this and the conflict GitHub currently reports (CHANGELOG.md and package.json against v2.57.0) in one move.

Once those two are in and `npm test` runs green end to end on the branch, this merges. It stays at the head of the queue.

ulsklyc:
@Cbrogg @atupitsin An update you did not ask for, and I want to be straight about what happened rather than let you find it.

I have taken the two blockers from the fourth round off your hands and opened #992 with the result. **The feature work is yours and stays yours**: your 22 commits are merged there with their authorship intact, not squashed into one of mine, so `git log` still shows who wrote what. #919 stays open and unchanged.

**Why I did it rather than wait.** Both blockers were plumbing, not feature code, and one of them was getting more expensive by the day: the branch had no real merge base, so the gap to main was 131 commits and growing. That is the kind of work a maintainer can take over without taking anything away from the people who wrote the feature. What I did not do is touch the feature itself, or reopen any finding from rounds 1 to 3 - those were verified closed by counter-check in round 4 and I took them as they stand.

**What came out of it, and it is worth knowing:** registering the three test suites was not a formality. `docs/test-suites.md` documented them, `package.json` knew none of them, and the `test` chain did not name them - so the files sat there and never ran. On their first real execution they contributed **35 tests nobody had ever seen**, and they pass. The full suite now completes at 239 suites and 5026 tests, green end to end, which is the first time `npm test` runs through on this branch at all.

**Where that leaves you.** If you would rather carry it yourself, say so and I will close #992 - the work is two commits and reproducible in an afternoon, and I would rather have your hand on it than my own. If you are fine with it, there is nothing you need to do. Either way #929 is next in the queue, and I have written the direction decision for it into that thread already so it is not waiting on an answer from me.

Thank you for the work, and for the patience with four review rounds.

PR #926, opened 2026-08-29
feat(tasks): die uebersicht oeffnet dieselbe aufgabe wie das aufgabenmodul

ulsklyc:
## What

A task clicked in the Overview or the Calendar now opens the **same** reading view the Tasks module opens - in place, without navigating away.

Before:
- **Overview**: a card with two buttons. "Edit" navigated into the Tasks module; "Mark as done" ticked it off. Subtasks, comments, attached documents, the tickable checkboxes in the description, due date, assignee, points and history were neither visible nor reachable.
- **Calendar**: a task chip in any of the four views navigated away outright - the month you were reading was gone.

That mattered more than two missing buttons sound like: the Overview is where the app actually stands open during the day, the Tasks module is where tasks get created and groomed. The view with fewer capabilities was the one being used more often.

## Why this shape

Duplicating the markup would have guaranteed the two drift apart at the next change - which had already happened once: `dashboard.js` carried its own `formatDueDate` and its own `PRIORITY_LABELS`.

So the reading view moved out of the page it was nailed to:

- **`public/components/task-detail.js`** (new) - the view itself. It used to read the viewer and the categories out of the Tasks page's `state` and reload that page after every change; both are knowledge only that page has. The caller now says who is looking (`currentUserId`, `isAdmin`), what can be resolved (`users`, `categories`) and how it refreshes itself (`onChanged`).
- **`public/utils/task-fields.js`** (new) - what a field of a task *means*: is it archived, may I rewrite it (#830), how does its due date read, what is its category called.
- **`public/utils/day-label.js`** (new) - "Today"/"Yesterday"/weekday, needed by two histories.
- **`pages/tasks.js`** exports `openTaskById(taskId, { user, container, onChanged })` - one entry point, loaded dynamically by the Overview and the Calendar so the task form's weight stays out of their startup.

The **edit form stays with the module** and is handed to the view rather than built into it. Without a mounter you get a view without an Edit button, instead of one that leads nowhere.

The viewer is passed explicitly and never guessed from a global: without it `mine` falls to false on every comment and a locked task looks locked to its own author - the view would offer less than is allowed, silently.

## Covers

All nine points of #918: shared component, full action set, rendered Markdown, every clickable chip (four calendar views, urgent-tasks widget, Today Cockpit), in-place refresh, unchanged permission checks, Escape/focus-trap (inherited from `detail-view.js`), and the interactive checkboxes from #917 - which now work from the Overview and the Calendar too, verified against a running instance.

## Guards

- New: `die Leseansicht ist nicht an ihr Modul genagelt (#918)` in `test-detail-view.js` - counter-proved three ways (re-introducing a `state.` read, rebuilding the two-button card, restoring the calendar's navigate-away each turn it red).
- Five existing guards follow the code to its new file; the rule each one checks is unchanged.
- `npm test` green.

Closes #918


ulsklyc:
Both review findings were real and are fixed in 08f0ff1e - thank you, they were the same mistake twice: context left behind when the view moved out of the page.

**The comment editor's Cancel path** rebuilt the row without `ctx`, and `commentRowNode` dereferences `ctx.currentUserId` on its first line. TypeError, comment stuck in edit mode. The other call site passed it correctly, which is exactly why nothing caught it.

**The optimistic hide on delete** searched for `[data-task-id]`. The Overview names its object differently - a cockpit row can mean any module, so it carries `data-object-kind` + `data-object-id`. Codex's second sentence turned out to be the more interesting half: measured against a running instance, the test task stood in **two** tiles at once (the cockpit row and the countdown tile), so even the right selector as a singular `querySelector` would have left one behind. It now hides every representation and restores all of them.

Two guards, each counter-proved against its own failure state: one asserts that every call to the node builders passes its context, the other that both spellings of a task row are searched - and that the Overview actually emits them, so the guard cannot run over an empty set.

ulsklyc:
Third round of fixes in c9188ba0. The P1 was real and the most important finding on this PR - thank you.

**Stylesheets.** The router keeps exactly one page stylesheet loaded (`loadPageStyle` in router.js), so on the Overview that is `dashboard.css` and in the Calendar `calendar.css`. `tasks.css` is not there. Measured in a running browser by toggling the sheet off and reading back computed styles: a tag chip came out at `border-radius: 0px` instead of `9999px`, a comment's author line at weight 400 instead of 600, the comment row with `gap: normal`. The code was shared, its appearance was not.

Those rules moved to `public/styles/task-detail.css`, loaded eagerly next to `detail-view.css`. The one genuinely mixed rule (`.task-card__meta > .priority-badge`) stayed in `tasks.css` - it describes the card's meta row, not the badge.

Moving CSS by hand deserves a proof rather than a look, so: 198 rules before, 198 after, none lost, none duplicated, verified with the repo's own rule scanner. Worth recording that my *first* verifier compared **sets** and was therefore blind to exactly the state a botched split leaves behind - a rule that afterwards sits in both files. It reported "nothing lost, nothing new" while 45 rules stood twice. It counts now, and I counter-proved that by planting a duplicate.

**The other two.** `container` goes back into the externally mounted edit form - your reading was right that `onChanged` already covers the reload concern, and the container's *second* purpose is the optimistic hide. Without it the same delete command behaved two ways. And when `/tasks/meta/options` fails, the Edit button is now withheld rather than offering a form with an empty category select that the server rejects after pending uploads have already gone through.

Three guards, each counter-proved against its own failure state. The style guard derives the eager set from `index.html` rather than listing it - a typed list goes stale silently, and my first draft of it had already misspelled one filename into a `.catch(() => '')`.

ulsklyc:
Fourth round, both findings real - and both the same root cause as the previous fix, which is why the answer this time is to remove the split rather than patch it.

`.priority-badge` was left behind, and the entire **edit form** (`.task-tag--editable`, `.task-tags-editor__chips`, `.task-field-hint`) was never considered at all: it is mounted from outside too, and it takes its appearance from `tasks.css` just as the reading view does. Splitting rules between a shared sheet and a page sheet means filing every new rule correctly, forever. That filing went quietly wrong twice in a single PR, which is a good argument that it is the wrong mechanism, not that I filed carelessly.

So `task-detail.css` is gone. `openTaskById` now ensures `/styles/tasks.css` and **waits** for it before opening - the same pattern the router already runs for `reminders.css`, plus the `onload` so nothing is shown raw first. One rule instead of a sorting job: open the module from outside, get its sheet.

Measured from the Overview, with only `dashboard.css` as the page sheet before the click:

| | before the fix | after |
|---|---|---|
| tag chip radius | `0px` | `9999px` |
| comment author weight | `400` | `600` |
| `.priority-badge` | unset | `flex` |
| form: tag editor / field hint / chip cursor | unset | `flex` / `12px` / `pointer` |

Worth naming why the guard I wrote last round did not catch `.priority-badge`: it searched the class names in the **file text**, and `.priority-badge` appeared in a *comment* in the shared sheet explaining why a related rule had stayed behind. A hit that is not a rule. It reads selectors through the repo's rule scanner now.

Guards counter-proved as before: removing the `ensureTaskStyles()` call and removing the `onload` wait each turn the test red.

ulsklyc:
@codex review

PR #929, opened 2026-08-29
feat(ui): Page Composition System (PAGE-COMPOSITION.md)

Cbrogg:
## Summary

Page Composition System **constitution + reference demo** (not a full geometry rewrite of every module).

- **PAGE-COMPOSITION.md** — Part A principles, Part B contract, **FAQ** (closed requirements), migration status table
- **Primitives** — `--layout-*` tokens, `.app-page--*`, `.page-measure`, `.page-section--bleed`, `.page-toolbar--measured` / `__rail`
- **Helpers** — `public/utils/page-layout.js` (`renderAppPage`, header/body/section/list/actions)
- **Audit** — PAGE-001…010; **strict** checks on the reference page
- **Other pages** — composition mode **declared** (declaration-first); deep helper migration = follow-up PRs

### Demo page (review this)

Open **`/birthdays`** at **1440 / 1920**:

- One `reading` column (`--layout-reading`)
- Header (title · search · actions) and list share the same primary edge
- Markup via helpers only; `birthdays.css` has no page width/gutters
- Marker: `data-composition-reference="true"`

Budget is **not** the reference.

## Test plan

- [x] `npm test` green
- [x] Reference audits (helpers, no compat alias, no page geometry in `birthdays.css`)
- [ ] Visual: `/birthdays` Header/Body same edge ±2px at 1440 and 1920
- [ ] Spot-check another reading page still loads (contacts)
- [ ] Spot-check budget KPI + list share measure (PAGE-002 wiring)

## Out of scope (documented)

Shopping kitchen tabs, meals slot grid, settings SPA, auth chrome.

## Follow-ups (stacked)

1. Wave A deep helper migration (contacts, notes, …)
2. Drop `page-toolbar--narrow::after` once Large-Title selectors accept `__rail`
3. Visual pack 390/768/1024/1440/1920
4. Budget 2.0 sidecar on core helpers post-merge


ulsklyc:
Reviewed on the branch. Below are the three visual checks your test plan left open - I ran them, they pass - plus four findings in the helper module.

## The open checkboxes, measured

Seeded household, tile fresh, Chromium.

**`/birthdays` header and list share the primary edge:**

| viewport | title left | list left | Δ | actions right | list right | Δ |
|---|---|---|---|---|---|---|
| 1440 | 252 | 252 | **0px** | 972 | 972 | **0px** |
| 1920 | 430 | 430 | **0px** | 1150 | 1150 | **0px** |

Better than the ±2px you asked for. The rail is wired: `.page-toolbar` carries `page-toolbar--measured`, and `.page-toolbar__rail` holds all three slots at `display: contents`, so title, search and actions align on the measure without the rail itself taking part in the layout. Worth noting for whoever measures next - a `getBoundingClientRect()` on the rail returns 0×0 and looks like an empty element.

**Another reading page still loads:** `/contacts` renders 371 content elements, declares `data-composition="reading"`, and carries `app-page--reading page-measure--narrow` - the compat alias doing its job. No console errors.

So the reference demo delivers what it claims. What follows is about the helper, which is the part every follow-up PR will build on.

## Should fix

### 1. Two JSDoc defaults are the opposite of the code

`public/utils/page-layout.js`, `renderPageHeader`:

```js
 * @param {boolean} [opts.narrow=false] — legacy ::after spacer; prefer measured
 * @param {boolean} [opts.measured=true] — wrap primary slots in .page-toolbar__rail
...
  narrow = true,
  measured = false,
```

Both are inverted. The documented behaviour is "measured by default, legacy spacer off"; the actual behaviour is the reverse. Defaulting to the legacy path so existing callers do not shift is a defensible choice - the reference page passes `measured: true` explicitly - but right now the doc promises the new system and the code hands you the old one. Whoever writes Wave A from the signature gets `page-toolbar--narrow` on every migrated page and will not find out from a test. Either flip the defaults or fix the doc and say in one line why the legacy path is the default during migration.

### 2. Escaping is hand-rolled where `esc()` is the house rule

`renderAppPage`:

```js
if (id) attrParts.push(`id="${id}"`);
attrParts.push(`${key}="${String(value).replace(/"/g, '&quot;')}"`);
```

`id` is interpolated raw, and the attrs path escapes only the double quote - `<`, `&` and a single quote pass through, and the *key* is not checked at all. CLAUDE.md names `esc()` from `public/utils/html.js` as the answer for this, and every page module in the repo uses it. Today these are all developer-supplied constants, so nothing is exploitable; the reason to fix it is that this is the shared helper the whole migration routes through, and a partial escape in it is the kind of thing that gets copied into a caller where the value is not a constant.

### 3. `renderPageTitle` writes its argument raw, and its signature does not say so

`renderPageActions` carries "already escaped by caller" in its JSDoc. `renderPageTitle(text)` interpolates straight into `<h1>` with no such note. Page titles are `t('nav.…')` everywhere today, but the point of migrating every page is that eventually one of them is a folder name or a contact name. One `@param {string} text — already escaped by caller` costs nothing, or escape it in the helper and let callers pass plain strings - either, as long as the contract is written down.

## Nice to have

`server/services/modules.js`: `composition` and `width` are validated against a `Set` and fall back to a default, while `navigation` and `responsive` two lines down are only trimmed to 40 characters. They are inert right now - nothing on the client reads them - which is exactly why it is cheap to close: an unvalidated string in a manifest block whose neighbours are validated will be read as "free-form on purpose" by the next person, and these are the values most likely to end up in a `data-` attribute later.

## Merge note

This and #919 both touch `server/services/modules.js` and `test/test-frontend-audit.js`. #919 rewrites `normalizeManifest`'s surroundings heavily, and this adds a `page` block inside it - whichever lands second will need a real rebase rather than a mechanical one. Might be worth deciding the order now.

## Not a review finding, but yours to decide

`PAGE-COMPOSITION.md` is a new constitution file at the repo root next to `DESIGN.md`, and it names `/birthdays` the reference page with a `data-composition-reference="true"` marker in the markup. That is a governance decision about where the geometry rules live and which page arbitrates them, and it outlasts this PR - the follow-up waves all point back at it. The engineering here is sound; that call is not mine to make.

Cbrogg:
Thanks for the review — and for closing the visual checks.

I opened this a bit early; the PR is still in progress. **#919 is the higher priority right now**, so I'll keep this in draft while that lands (and to avoid the `modules.js` / audit-test rebase tangle you flagged).

All of the helper findings are accepted — JSDoc vs. actual defaults, `esc()` instead of the hand-rolled escape, the `renderPageTitle` contract, and validating `navigation` / `responsive`. I'll address them here and mark the PR ready for review again when that's done.

ulsklyc:
Confirming the order you proposed: **#919 first, then this one, then #930.** Keeping this in draft until #919 lands is exactly right - it avoids the `modules.js` / `test-frontend-audit.js` rebase tangle entirely.

No rush on the helper findings from my end; they can ride along with the rebase.

ulsklyc:
@Cbrogg Deciding this now rather than after #919 lands, because once the queue points here you would be building on an answer you never got.

**The direction: the primitives are in, the constitution is not.**

What I am taking, without reservation: `--layout-*` tokens, `.app-page--*`, `page-measure`, the toolbar rail, and `public/utils/page-layout.js`. I measured the reference page myself and posted the numbers - 0px on both edges at 1440 and at 1920, better than the tolerance you set yourself. That part is solid work and it is the part the app actually needs.

What I am not taking in the shape proposed, and why:

**1. `PAGE-COMPOSITION.md` as a rules file next to CONTRIBUTING.md.** It goes into `docs/` as documentation instead. Two reasons. It carries a migration status table, which means it holds state and drifts the moment a page changes without it. And this repository has exactly two public rules files today; a third that answers layout questions in its own voice is a third place the same question can be answered differently.

**2. The strict audits running only on the reference page.** This is the one I would push back on hardest. PAGE-001 to 010 are strict on `/birthdays` while every other page merely declares its mode. That is an allowlist with one entry: one page is green and the guard says nothing about the thirty others. I have shipped that shape before in this project and it always ends the same way, with the guard passing while the rule is broken everywhere it does not look. The audits should check the rule wherever a page claims a composition mode, even if that means starting with fewer checks that apply everywhere rather than ten that apply once.

**3. `data-composition-reference="true"` in the shipped markup.** A page should not carry its own status as a documentation anchor. If `/birthdays` gets reworked, either the marker migrates or the rule loses its anchor, and both are silent failures.

**Something I owe you, because it explains why you had to write that file at all.** The layout rules I decide by live in `CLAUDE.md`, which is gitignored and not in this repository. You have never been able to see them. From the outside, the only written constraint is CONTRIBUTING, and a page composition system genuinely looked unwritten. It was not, it was invisible, and that is my gap and not yours. Making those rules visible is now on my list, and where they overlap with your Part A I would rather merge them than have two documents.

**What this means for your work:** nothing in the helper module or the primitives changes. The file moves, the marker comes out, and the audit becomes a rule over all pages instead of a strict pass over one. Less code than the version you have.

**Order is unchanged: #919 first, then this, then #930.** #919 is still waiting on the two harness blockers from my fourth pass, and it is worth saying that the branch is now 120 commits behind main - the real `git merge origin/main` that fixes blocker 2 gets more expensive every day it waits. That is the one I would spend the next hour on.

ulsklyc:
@Cbrogg @atupitsin I have taken this over and rebuilt one part of it, as #995. What follows is what
changed and why, because you should not have to read a diff to find that out.

**First, an apology that is also an explanation.** Workflow runs from forks wait
for maintainer approval here, and yours had been sitting in that queue since
August 29. Three pushes, six runs, none of them ever executed. So when I said the
PR was quiet, the honest version is that it never got a single piece of feedback
to be quiet about. That is fixed now.

**What I did not touch: the substance.** The `--layout-*` tokens, the
`.app-page--*` primitives, `page-measure`, the bleed section, `page-layout.js`,
the measured toolbar rail, the birthdays rework. All in, unchanged, with
@atupitsin's commits and their authorship intact. The merge is a real merge, not a hand-applied patch.

**What I changed, and it is one idea in three places.**

Your version made `/birthdays` strictly auditable and let the other pages
declare. The audit had a `COMPOSITION_MIGRATED` allowlist, and every loop opened
with `if (!MIGRATED.has(name)) continue;`. Two consequences: a page added
tomorrow falls through both lists silently, and the CSS rules ran over five
hard-coded filenames out of forty-two.

So the scope now comes from `router.js` instead of from a list. A route with
`requiresAuth: false` renders without the app shell and cannot satisfy the
contract; everything else is inside it. What is left is a named list of pages
that do not comply yet:

```js
const COMPOSITION_PENDING = new Set(['shopping.js', 'meals.js', 'settings.js']);
```

PAGE-011 fails if that list grows, and fails if an entry points at a page that
no longer exists. Migrating a page is now a deletion.

Being precise about what that buys, because the headline number is unimpressive:
the guard checks the same **19 pages** it did before. What actually changed is
that the CSS rules went from 5 hard-coded files to **17**, and that a new page is
covered on the day it gets a route. I verified the second one by adding a page
with no composition mode: PAGE-001 fails on it now, and would have skipped it
before.

**Two smaller things.**

`PAGE-COMPOSITION.md` moved to `docs/`, and the `data-composition-reference`
attribute is gone from `birthdays.js`. That attribute was a test hook shipped in
production HTML: every visitor paid for it, and it said nothing about the other
thirty pages. A test now fails if it comes back. The guarantee belongs in the
guard, not in the markup.

`test/test-schedule.js` asserted the literal string `<div class="schedule-page">`
and went red against your correct change to `<div class="schedule-page app-page
app-page--data">`. That guard was reading spelling rather than the thing it
meant to protect, so I widened it rather than narrowing your markup.

**One real bug the rebase surfaced, and it makes the CI point concrete.**
`public/utils/page-layout.js` was missing from the service worker precache list
in `sw.js`, while `birthdays.js` imports it. `test-sw-precache.js` on `main`
catches exactly this: after an update the old copy of a helper can end up bound
against a new page module. One line added, fixed. That guard has been on `main`
the whole time - your branch just never got to run it.

**Where this leaves the stacked follow-ups.** They no longer need anyone to drive
them. Three pages are named in the guard, and each one is a line to delete. If
you want to take one of them, say which and it is yours.

PR #930, opened 2026-08-29
feat(schedule): quick-start presets, range fill, grouped override editing, and a dashboard widget

mclgoerg:
## Summary

Four additions to the Schedule module (#786), plus a handful of fixes found by using the feature after the fact - none change the data model.

### Added

- **Shift-type quick-start.** The Shift Types tab's empty state offers a one-click "quick start" that creates seven presets (Early/Late/Night/Day/24-hour, plus Vacation/Sick) against the existing unrestricted `POST /shift-types`, sequentially - no new bulk-create endpoint, since nothing else in the app seeds N starter rows that way either. Vacation and Sick carry no start/end time on purpose: the schema already allows a shift type without times (`start_time`/`end_time` are nullable as a pair, and a type without them already renders as "all day"), so an absence reason needed no new column or concept, just two more presets.
- **Range fill for overrides.** `POST /overrides/fill` upserts a shift (or a free day) across an inclusive date range in one call - e.g. marking a two-week vacation - instead of one `PUT` per day. Capped separately from `/entries`'s read-side `MAX_RANGE_DAYS` (731 days): a fill *writes* real rows, and the module's whole premise is computing entries on read rather than materializing them, so the cap is a much smaller, independently-justified `MAX_FILL_DAYS` (100 days).
- **Grouped override display and reconciled range editing.** Live use surfaced the next problem: a filled range then showed as N identical rows with no way to treat them as one thing. The Overrides tab now groups consecutive same-person/same-type/same-note days into a single row - display and bulk-action only, the table is still exactly one row per day underneath. Editing a group shows its current From/To as editable fields; saving reconciles automatically (fills the new span via the existing endpoint, and a new `DELETE /overrides` removes whatever fell outside it) instead of a manual delete-then-refill. Deleting a group removes its whole span in one confirmed action.
- **Dashboard widget.** An opt-in "who's working today" tile, off by default like the module itself, reusing the same `/entries` query and `resolveEntries()` rule the module's own "today" card already uses - a member with neither a pattern nor an override for today has no row, on the widget or the page.

### Fixed

- **Two new routes were missing from the OpenAPI spec.** The rebase onto `main` pulled in a same-day guard (`test:openapi-coverage`) that caught it immediately - `POST /overrides/fill` and `DELETE /overrides` are documented now.
- **The Overrides tab's empty state didn't match its siblings.** Patterns and the new Shift Types quick-start both use the shared `emptyStateHTML` grammar (icon, title, description, CTA); Overrides was still the original bare paragraph. Three tabs of one module should read as one module.
- **The shift-type colour picker spanned the full row on a phone.** `width: 100%` stretched the native colour input to fill its grid cell - harmless on the desktop two-column layout, a full-width bar once the mobile breakpoint collapses the form to one column. Fixed size now, matching the compact swatches already used elsewhere in the app (`subscriptions.css`).

## Test plan

- [x] `npm test` - full suite green, run after rebasing onto `main`
- [x] New/extended suites: `test-schedule.js` (fill, collection delete, client grouping/edit/delete wiring, the empty-state grammar guard, the Vacation/Sick preset guard), `test-dashboard.js` (widget wiring, refresh carry-over, the widget-count guard)
- [x] Manually verified end to end against a live deployment (real TLS, real setup flow, a real household, not just the seeded demo data): quick-start creates all seven presets including the two timeless ones; a filled 14-day range shows as one grouped row; editing a group's From/To and shift type reconciles to exactly the kept days, removing what fell outside the new range; deleting a group clears its whole span in one action; the dashboard widget shows the right person/shift and survives a page reload; the colour picker measures a fixed box on a 390px mobile viewport instead of stretching full-width; the Overrides empty state visually matches Patterns

---
Drafted with Claude Code.


ulsklyc:
Reviewed against `main`. The server half is the strong part: `MAX_FILL_DAYS` is argued rather than inherited, and the comment says exactly why (`/entries` reads, `/overrides/fill` writes, and the module's premise is computing on read) - that is the reasoning I would have asked for. `DELETE /overrides` correctly sits before `/overrides/:dateKey`, both new routes go through `mineOrAdmin`, and the tests cover the cap, the inverted range, a non-existent shift type, and the member-vs-admin split on both. The OpenAPI gap being caught by a same-day guard and fixed in the same PR is the system working.

One measured problem, one observation, one merge note.

## Should fix

### The widget declares 1×1 and renders 318px, stretching its neighbour with it

`defaultWidgetSize()` in `public/utils/dashboard-widgets.js` was not extended, so `schedule` falls through both lists to the `'1x1'` default. But the tile is a member list - avatar, name, shift - which is why `family` and `birthdays` sit in the `'1x2'` group, with the comment there spelling out the reason ("Beide sind Listen wie die anderen und brauchen Hoehe, nicht Breite") and the `metrics` note above it naming the failure mode: content-heavy cards start tall "damit die Zeile nicht per grid-auto ragged nachwächst (Critique P4)".

Measured on the seeded household (five members, all on shift today, 1440px viewport, tile freshly enabled from the Customise tray):

| tile | size class | rendered height |
|---|---|---|
| `rewards` | 1x1 | **218px** |
| `schedule` | 1x1 | **318px** |
| `notes` | 2x1 | 218px → **318px** |

`notes` is in the same grid row and gets stretched by 100px to match. So the tile is 46% taller than the size class it claims, and it drags an unrelated tile along. Five members is the small case - the module's own audience (a shift household, a care rota) is where six or eight members are ordinary, and each row is 44px.

Two parts to it, and the second matters more:

1. Add `schedule` to the `'1x2'` list in `defaultWidgetSize`, next to `family` - it is the same shape of tile.
2. `renderScheduleWidget` is the only list tile with no row cap. Every other one takes its row count from its size (`listRowCap`, `LIST_ROWS_SHORT`/`LIST_ROWS_TALL`), which is what keeps a tall tile at five rows and a flat one at three no matter how much the payload holds. Without it, 1×2 only moves the overflow point from five members to six. `renderScheduleWidget(schedule, users, size)` and a `slice(0, listRowCap(size))` is the whole change; the header count is already `onShift` rather than the row count, so it stays honest when the list is cut.

Worth knowing: #932 fixes exactly this defect on the Notes tile, which had the same gap. That one had the cap on the server side; here the payload comes from `/schedule/entries` and is unbounded by design, so the cap belongs in the renderer.

## Observation

`POST /overrides/fill` writes `note = excluded.note` on conflict, and `str()` returns `null` for an absent note. Filling a range therefore clears any note that already sat on a day inside it. The confirm copy says "replacing any existing entries in that range", so it is covered in wording - I am flagging it because a note is the one thing in an override a person typed by hand, and it is the part that disappears without appearing in the confirmation's own example. Deliberate is a fine answer; it just deserves a line in the code saying so.

## Nice to have

The client-side test is a text guard: `assert.match(schedulePage, /function overrideGroups\(\)/)` asserts that an identifier appears in the source, not that consecutive same-type days group into one row. Renaming the function turns it red without anything being broken, and - the direction that actually costs - a `overrideGroups()` that returned `[]` would keep it green. The grouping and `rangeDifference()` are pure functions over dates; exporting them through a `__test` door and asserting on actual input (three consecutive days → one group; a gap → two; an edit that shrinks a span → the right difference) would hold the behaviour instead of the spelling. The server tests in this PR are the model.

## Merge note

This touches `public/pages/dashboard.js`, `public/utils/dashboard-widgets.js` and `test/test-dashboard.js`, and so do #931 and #932. Whichever lands second will need a rebase - the collisions are all additive (a widget id in `WIDGET_IDS`, an entry in `widgetById`, a new test), so it should be mechanical.

ulsklyc:
Heads-up on the merge note in my review: #931 has landed on `main` (the calendar tile's birthday toggle), and this branch now shows as conflicting. The overlap is what I expected - `public/pages/dashboard.js`, `public/utils/dashboard-widgets.js` and `test/test-dashboard.js`, all additive on both sides, so the rebase should be mechanical.

One thing worth knowing while you are in there: #932 landed the row-cap fix on the Notes tile, which had the same defect I flagged on the schedule widget. `renderPinnedNotes(allNotes, size)` with `slice(0, listRowCap(size))` is the shape, if it helps to have a worked example in the same file.

ulsklyc:
Heads-up on sequencing so you do not rebase more often than necessary.

Merge order for the three open PRs is **#919 first, then #929, then this one.** Both of those touch `server/services/modules.js` and the frontend audit test, and #919 also touches `public/pages/dashboard.js` - which is where this PR currently conflicts, since #931/#932 landed there for v2.54.0.

So: **there is no point rebasing right now.** If you rebase today you will hit the same area again after #919 and once more after #929. Waiting until both are in means one rebase instead of three.

What is worth doing in the meantime is the tile finding from my review - the new dashboard widget needs the row cap, the same one the birthdays, notes and schedule tiles each needed in turn. That is independent of the rebase and will not conflict with anything.

I will ping here once #929 is merged and the branch is clear for you.

mclgoerg:
Thanks for the sequencing heads-up — makes sense, and appreciated. Holding off on rebasing until #929 is in; no point eating the same conflict area three times.

Pushed the tile fix in the meantime (independent of the rebase, as you noted): `schedule` now sits in the `1x2` group next to `family`, and `renderScheduleWidget` takes a `size` and caps its rows via `listRowCap(size)` the same way `renderPinnedNotes` does after #932 — header count still comes from the full `onShift` total, not the capped row count. Also added the deliberate-behavior comment on `overrides/fill`'s note-clearing, and swapped the `overrideGroups()` text-guard test for real behavioral ones (`__test` door, exact cases from your review: three consecutive days → one group, a gap → two, a shrinking edit → the right remainder).

Full suite is green on top of the current `origin/feat/schedule-v2` tip (no rebase mixed in). Will wait for your ping once #929 lands.

mclgoerg:
@ulsklyc, is there a planned ETA? I already have v2 and v3 ready.

ulsklyc:
@mclgoerg You are next, and the queue in front of you is gone.

#919 landed today as #992 - the two blockers there were plumbing (no real merge base, and three test suites that were documented but never registered), so I took them over rather than let them keep growing. It is on `main` as of v2.62.0.

**And the order changed in your favour, because I measured it instead of assuming it.** #930 was queued behind #929 on the grounds that both touch the dashboard and the schedule page. In those two shared files #929 changes **one line each**; yours changes 111 and 218. That is not a dependency, it is a trivial collision that resolves in either direction, and #929 is still a draft. So there is no reason for you to wait on it.

What I need from you: a rebase onto current `main`. The PR is CONFLICTING and roughly 134 commits behind, and after today's release the conflict is mostly the changelog. I would rather you do it than do it for you - you have kept this branch to the conventions all along, tests, OpenAPI, SPEC and all 24 locales included, and it is your call how v2 and v3 stack on top.

On those: please keep them as separate PRs rather than folding them into this one. Three reviewable steps take less of both our time than one large one, and it means the first can be in people's hands while the second is still being read.

No date from me on the merge itself beyond this: once it is rebased and green, it is the next thing I look at.

ulsklyc:
@mclgoerg One correction to what I said, and it works in your favour.

**Your August 29 push never got a CI run at all.** The most recent workflow run
on this branch is against `4017042d`; your head is `c3af6a3f`. So the tile fix
you pushed after that comment has never been evaluated. It is not that CI was
red or waiting: no run exists. The workflow has no path filter, so this is not a
matter of which files you touched.

Practically it means the rebase you are about to do is also what gets you your
first real check on this branch. Push it and the run will be there.

#929 is integrated and waiting to land as #995, so the conflict area you were holding
off on is about to be settled. I will say the word here the moment it is on
`main`.

mclgoerg:
@ulsklyc Rebased onto current `main` (193e9353) and pushed (`f60f7018`). Conflicts were exactly what you expected: `CHANGELOG.md` and `docs/test-suites.md` both had the same insertion point diverge (your Unreleased entries landing where mine did), and `dashboard.js` had one line, the `__test` export list, where your `renderDashboardOverview` and my `renderScheduleWidget` both got added. All additive, resolved by keeping both sides.

On the missing CI run: this push has a real one now - Tests (Node.js 22.x) passed. Full suite is green locally on top of the rebase too, and the PR shows as mergeable.

v2 and v3 staying separate, as agreed - v3 isn't opened yet. Will watch for the word once #995 is on `main`.

ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**Why this one goes second, and an apology for the wait.** It is the oldest PR in the queue, CI is green on the current head, and the pairwise matrix shows it conflicting with none of the other six open PRs, in any file. Merging it early therefore costs the rest of the queue nothing, while leaving it to age costs a contribution. That has happened in this repository before and I do not want to repeat it.

**What it needs.** A rebase for the CHANGELOG head only. Everything else, including all 24 locale files, `public/pages/dashboard.js` and `server/permissions.js`, merges cleanly against current `main` and against every other open PR.

Review continues from where it stopped. The dashboard widget row cap from the last round is addressed in f60f701, which puts the schedule widget on `listRowCap(size)` like the other list tiles.

ulsklyc:
Follow-up on the green check here, which needed no approval from me.

**The green check is older than it looks, and it does not mean this branch merges with `main`.** The run was created on 1 September. `actions/checkout@v7` on a `pull_request` event checks out `refs/pull/N/merge`, and GitHub computes that merge ref when the run is created, not when it starts. Since then `main` has moved 19 commits, the v2.63.0 release included.

ulsklyc:

ulsklyc (changes requested):
Review done. Short version: this is close, the invariants all hold, and what is left is three one-liners plus a rebase that is mine, not yours.

**What I checked and found clean.** No `innerHTML`, ESM throughout, no migration touched, both new routes documented in the OpenAPI spec with the scope inherited correctly from the `schedule` path prefix, one page CSS per page, and the widget registered in all three lock-step lists. The 15 new locale keys are present in all 24 files at the right nesting and indent, none of them takes a numeric `count`, so no `_one` variant is owed. Dates go through `formatDate()` / `addLocalDays()` / `daysBetweenDateKeys()` rather than `toISOString().slice(0,10)`, which is the recurring date bug in this repo and you avoided it.

## Blocking, and it is mine

`CHANGELOG.md` conflicts. Your anchor bullets sat under `## [Unreleased]` at your merge base; `main` has since released v2.63.0 and v2.64.0, so they now sit inside a shipped section and your new blocks would land there. Per [CONTRIBUTING](https://github.com/ulsklyc/yuvomi/blob/main/CONTRIBUTING.md#6-review-and-merge) this is mechanical and I resolve it when the PR lands - please do not rebase for it.

Worth knowing: the green check on this PR ran against a base four commits old, from before the page-composition system landed with its new `PAGE-0xx` guards in `test/test-frontend-audit.js`. It will be re-run against current `main` before merge.

## Three things I would like changed

**1. The widget badge counts rows the widget does not show** - `public/pages/dashboard.js:2061`

```js
const onShift = entries.filter((entry) => entry.shift_type).length;
const rows = entries.slice(0, listRowCap(size)).map(...)
```

`onShift` counts every entry with a shift; `rows` takes the first N in the order the API returned them, which is `users ORDER BY id`. With six members where the two on shift happen to have the highest ids, the header says "2" and every visible row reads "Freier Tag". Sorting on-shift entries first before the slice fixes it, using the badge's own predicate.

Your comment right above this names exactly this class of bug from the notes widget, and the row cap you added in f60f701 is the right fix for the *height* half. This is the other half: a cap answers "does it overflow", never "which rows survive".

**2. The widget has no accent colour** - `public/styles/dashboard.css`

`widget--schedule` does not appear anywhere under `public/styles/`. `widgetHeader()` sets the seal to `var(--module-schedule)` from the `/schedule` link, but `--widget-accent` falls back to `var(--color-accent)`, so the badge and the "Alle" link render in app violet around a teal seal. The comment on `.widget--countdown` at `dashboard.css:285` documents this exact defect from last time. One line next to it:

```css
.widget--schedule { --widget-accent: var(--module-schedule); }
```

**3. The double-click guard does not guard** - `public/pages/schedule.js:735`

`if (state.types.length) return;` is followed by seven sequential `await api.post('/schedule/shift-types', ...)`, and `state.types` is only refreshed by `load()` in the `finally`. A second click during those round trips re-enters with `state.types` still empty and creates all seven again. I checked whether anything downstream would catch it: `schedule_shift_types` has no UNIQUE on `name` or `short_code`, and the route does not test for duplicates, so you would get 14 shift types with duplicate names. Setting `button.disabled = true` synchronously covers it. The comment above the guard currently claims the opposite and would need a line too.

## Smaller things, take them or leave them

- `public/styles/dashboard.css:3700` - `.schedule-widget-row__dot` uses `.6rem`, one of only two rem literals in the file, for the same visual object that `schedule.css:169` sizes at `.7rem`.
- `public/styles/schedule.css:90` - the comment says the fixed width follows `subscriptions.css: 44-52px`, and then sets `64px`. Either number is defensible, a third one is not.
- `public/pages/schedule.js:297` - `overrideGroups()` builds `group.ids`, but nothing in production reads it; only the new test asserts on it.
- The JSDoc for the widget promises three states (data, `null` for a load error, `undefined` for hidden), but the renderer folds `null` into "no shift types configured" and offers a create button on what was actually a failed request. `renderCycleWidget` does the same, so this is precedent rather than regression - but then the doc comment overstates what the code does.
- `dashboard.scheduleEmpty` lands between `housekeepingNone` and `housekeepingPresent` in all 24 locale files, splitting that run.

## Verdict

Requesting changes on the three above, and none of them touches the data model, the API shape or the migration array. This is the oldest open PR here and it has waited longer than it should have; once those land I will handle the CHANGELOG and merge it in the published order.

mclgoerg:
Pushed `69fccb47` addressing all three.

1. **Widget row order** — `renderScheduleWidget` now sorts on-shift entries first, stable, before `listRowCap()` slices. Added a regression test using your exact scenario (six members, on-shift ones with the highest ids) asserting the header badge and the visible rows agree.
2. **Accent colour** — `.widget--schedule { --widget-accent: var(--module-schedule); }`, same pattern as `.widget--countdown`.
3. **Double-click guard** — `button.disabled = true` synchronously right after the `state.types.length` check, since `state.types` isn't refreshed until `load()` in the `finally`. Updated the comment above it to say why.

Also took the two "take it or leave it" ones since they were one-liners in files already open: `.schedule-widget-row__dot` now matches `schedule.css`'s `.7rem`, and the color-field width comment no longer claims to follow `subscriptions.css`'s 44-52px range when it's actually 64px for its own reason.

Left the `overrideGroups().ids` and JSDoc-state items as you framed them — genuinely optional, and the doc overstatement has precedent elsewhere per your own note.

`test:dashboard` (99/99), `test:schedule` (24/24), `test:typography` (17/17), and `test:frontend-audit` (337/337) all green on top of the current tip. Not rebasing — leaving the CHANGELOG mechanics to you per CONTRIBUTING.

ulsklyc:
Heads-up before you see it in your notifications: I am about to close and immediately reopen this PR. **It is not being rejected.** It is the only way I have to get CI to look at your latest push.

GitHub created no workflow run for your new head at all - not a pending one, not a failed one, none. Measured by branch: the newest runs on this branch are still from 1 September against the previous head. Actions are healthy otherwise; `main` has had successful runs this evening. `ci.yml` only triggers on `push` to `main` and on `pull_request`, with no `workflow_dispatch`, so I cannot start it by hand. Reopening fires a fresh `pull_request` event and should queue one.

Nothing for you to do, and please do not push again on my account - the missing run is not something you caused. If the reopen does not produce a run either, I will say so here rather than leave you guessing.

I am also adding `workflow_dispatch` to the CI workflow so that the next time a push falls through like this, it can be started without touching anyone's PR.

ulsklyc:
Following up as promised, and two of the things I told you earlier were wrong.

**The reopen did not work.** No run appeared. I sampled both branches every 25 seconds for five minutes: the newest workflow runs are still the ones from 1 September against your previous head.

**And `workflow_dispatch` does not help here either.** I claimed it would let me start a run without touching your PR. Measured since: dispatching against `refs/pull/N/head` and against the branch name both return `422 No ref found`. A manual dispatch only accepts a ref that exists in this repository, and a fork branch is not one. The trigger is still worth having for branches that do live here, but it does not cover your case, and I should have checked that before writing it down.

**So I ran the suite locally against your exact head instead.** Full `npm test`, the same thing CI runs:

```
240 suites, 0 failures, exit 0   (head 69fccb47)
```

One caveat, and it is mine to fix: on your branch as it stands, `test:changelog` fails with *"Getaggte Releases ohne CHANGELOG-Eintrag: 2.63.0, 2.64.0"*. That is the stale CHANGELOG, not your code - I set the file to the current `main` state for the run above, and everything passed. The CHANGELOG merge is the maintainer's job here and I will do it when this lands.

**All three review points check out**, verified individually rather than taken from your commit message:

- `dashboard.js` sorts on-shift entries before the cap now, so the badge and the visible rows agree.
- `dashboard.css` carries `.widget--schedule` with the module accent.
- `schedule.js:740` sets `button.disabled = true` synchronously inside the quick-start branch.

I am converting my review to an approval. Nothing further from you, and please still do not push on my account - the missing CI run was never something you caused.

ulsklyc (approved):
Approving: all three points from the earlier review are addressed and verified in the code, and the full suite is green against 69fccb47 with a current CHANGELOG. The CHANGELOG merge itself is mine and happens at merge time.

Issue #944, opened 2026-08-30
Feature request: email notifications for events + share shopping list with a household member


bartfaizoli76:
Since there's currently no Android app, email would be a really useful fallback notification/sharing channel for households that mostly interact with Yuvomi via mobile browser or don't want to keep a tab open.

**1. Event notification email for assigned members**
Right now calendar event notifications only go through Gotify/ntfy/Webhook (Settings → Notifications). It would help a lot to also have a plain email option — send a reminder email to the household member(s) an event is assigned to, ideally configurable per event/per user (e.g. "notify by email X minutes before").

**2. Send/share the shopping list with a selected member**
A way to send the current shopping list (or a specific list) to one chosen household member — e.g. via email, or a shareable read-only link — would help a lot for cases like "someone else is doing the grocery run today, here's the list" without them needing to log into the app on their phone.

Both would meaningfully improve the mobile-less experience until (if ever) a native app exists. Thanks for a great project!

ulsklyc:
Thanks - part 1 is built and is up as #945.

**Email is now a fourth channel type**, next to Gotify, ntfy and webhook, configured the same way under Settings → Personal → Notifications. The thing worth knowing about how it turned out: an email channel deliberately brings **no credentials of its own**. It reuses the SMTP access that already sends password resets and invitations, so a mail server is configured once for the whole app rather than once per channel. A channel therefore holds just a recipient address.

That also answers the "configurable per event / per user" part of your request, though not in the shape you sketched. Yuvomi's channels already carry a scope - household-wide, or bound to one user - and email inherits it rather than introducing a second way to decide who gets what. One address per channel; a second recipient is a second channel, which keeps each one separately switchable and separately testable. Per-event opt-out is not part of it: a reminder is already something you chose to set.

Two details you will run into:

- The subject line carries both origin and subject (`Calendar: Dentist`), because an inbox shows only the subject. Web Push can split those across title and body; a mailbox cannot.
- The link back into the app needs `BASE_URL` to be set. Without it the mail arrives without a link rather than with a dead one - the request Host is deliberately not trusted here, the same rule that applies to password-reset links.

If you track medications, please read the note this added to `docs/PRIVACY-FOR-SELFHOSTERS.md` first. Reminder contents land in the **subject line**, medication names included, and subject lines stay readable in transit and permanently in the recipient's mailbox. On your own mail server that is fine; through a third-party provider it makes them a recipient of health data.

---

[...]

PR #945, opened 2026-08-30
feat(notifications): email as a fourth channel type (#944)

ulsklyc:
Closes the first half of #944. Reminders reached Web Push, Gotify, ntfy and a webhook; households without a native app were left keeping a browser tab open. Email is now a fourth channel type, configured in the same place as the other three.

## The decision that shaped it

**The channel carries no credentials of its own.** The SMTP access already configured for password resets and invitations is the one it uses. A second set per channel would be a second spelling of the same thing: two places to update when the mail server changes, and one of them forgotten. So an email channel holds exactly one thing, `toAddress` - and nothing in `secret_json`.

For the same reason it is a *transport*, not a second addressing model. Who receives a reminder is still answered by the channel's `scope`, exactly as for the other providers. One address per channel, so each stays separately switchable and separately testable; a second recipient is a second channel.

## Four details that decided the build

- **The subject carries origin *and* subject.** Web Push splits them across title and body, but an inbox shows only the subject line, so `Calendar` alone says nothing. It sends `Calendar: Dentist`. Line breaks are stripped first - otherwise a task title turns one header into two.
- **User data in the HTML part is escaped.** For that, `esc()` moved to `public/utils/html-escape.js`, re-exported from `utils/html.js` so all ~70 frontend callers are untouched. The backend should not have to bind the note markdown renderer (and its CSS class names) to escape a string; a second escape function would be worse still - two copies drift at exactly the character that matters.
- **No `BASE_URL`, no link.** A relative `/calendar` is not a destination in an inbox. The request Host is deliberately not consulted; during a background run no request comes past anyway.
- **nodemailer honours no `AbortSignal`,** so the send races the orchestrator's timeout. Otherwise one unresponsive SMTP server stalls a run that works through every due reminder in sequence.

`GET /notifications/providers` additionally reports `ready: false` while SMTP is unconfigured. Phrased as a provider capability, not as a special case for `email` - the test route answers a failure with a generic `Internal error`, so the one helpful sentence would be lost there.

## A precache guard that was blind

Adding `html-escape.js` should have turned the precache guard red. It stayed green: it read only imports written `from '/path'` and never saw `from './neighbour.js'`, though the browser loads both the same way.

After teaching it relative specifiers, it immediately found that **`settings/dirty-guard.js` has hung uncached off the precached settings shell since July**. Online nobody notices, because the network fills the gap. Offline the import fails and takes the whole settings page with it. Both are precached now.

Counter-check: with the old expression and the entry missing, the test is green at exactly that broken state.

## Verification

- `test:notifications` 30 → 39. Each of the 8 new guards was individually disabled and confirmed red.
- The hanging-SMTP test initially *hung* instead of failing when its protection was removed. It now carries its own 5 s limit - a hanging test blocks CI instead of warning it.
- Green: `i18n` (53), `sw-precache`, `layer-boundary`, `frontend-audit` (336), `openapi-structure`, `openapi-coverage`, `email`, `api`, `changelog`, plus the installer and preferences suites.
- `test:weather` failed once in the full run with a network error (148 s timeout, non-HTTP response). Green in isolation both with and without this branch; nothing here touches weather or `fetch`.

## Not in this PR

The second half of #944 - sending a shopping list to a chosen member - is deliberately left out. That is an access question rather than a transport one, and it deserves its own decision.

Docs updated: `SPEC.md`, `installation.md`, `notification-webhooks.md`, `test-suites.md`, README, CHANGELOG. `PRIVACY-FOR-SELFHOSTERS.md` gained a note that matters for anyone tracking health data: an email channel puts reminder contents in the **subject line**, medication names included, and subject lines stay readable in transit and permanently in the recipient's mailbox.

No migration. No new environment variables.


ulsklyc:
All six review findings addressed in `967423e` - each reproduced first, then fixed. Three of them broke rules this PR itself writes down, which is the useful part.

**1. A hidden field blocked submission** (Codex P2, Claude). The rule was already in the comment above `baseUrl`; the new `emailTo` did not get it. `display:none` does not exempt a field from constraint validation, and `type="email"` brings its own `typeMismatch` check that applies even without `required`. Type `oma@`, switch back to Gotify, hit Save: nothing happens, because the browser cannot focus a `display:none` control to report the error.

Fixed as a rule over the whole block rather than a list of fields - `setBlockActive()` sets `disabled`, which removes an element from validation *and* from submission. The next provider with a field of its own is covered without anyone remembering to.

**2. The SMTP hint never appeared while creating a channel** (Claude ×2, Codex P2). It was only emitted at render time, a new channel always starts as Gotify, and changing the provider does not re-render the card. So it showed up only after saving - precisely when it is no longer needed. That was the entire point of the `ready` flag. It is now always in the markup and toggled the same way the provider blocks are.

**3. Reminder contents were written to the log** (Codex P1). `sendMail` logs recipient and subject at info level. That was harmless while only password resets and invitations went through it - fixed subjects. A reminder subject carries the task or event name, and for a medication reminder that is health data, sitting on container stdout indefinitely. Callers can now pass a `logLabel`; the provider names its kind instead of its subject. This is the exact concern this PR documents in `PRIVACY-FOR-SELFHOSTERS.md` - and would have made worse.

**4. The 8-second abort could deliver twice** (Codex P1). Correct, and the sharpest of the six. The race lets the underlying send continue, so an SMTP server that accepts late would have delivered the mail *and* triggered a retry. The transport now carries its own, tighter timeouts (5 s against the orchestrator's 8 s) so nodemailer gives up first and the connection is genuinely closed - only then is a retry honest.

The race stays as a second net, deliberately. If it ever fires, a duplicate mail is the cheaper failure: a run stalled on a silent server does not hold up one notification, it holds up all of them, for everyone, again at every wake-up.

**5. An address list passed validation** (Codex P2). `a@example.com,postmaster` got through, because the second entry needs no `@` of its own. nodemailer reads `to` as a list, so that is two envelope recipients behind one channel with one delivery status. Commas and semicolons are rejected now.

**6. The precache guard was blind a second time** (Codex P2). It resolves relative specifiers since the previous commit but still missed side-effect imports (`import '/x.js'`, no `from`). Hiding in that gap: `/components/datepicker.js`, loaded by the router itself - so a first offline visit got the HTML fallback instead of the module, on every page with a date field. Both forms are resolved now.

---

New suite `test:notification-channel-form` (7 tests) covers the visibility logic as **behaviour** against a DOM stub rather than as a text guard. Both frontend bugs only exist in the interaction between rendering and provider switching; read line by line, every statement looks right. `test:notifications` is at 43.

Every new guard was individually disabled and confirmed red - the timeout one also against a merely *present* rather than *staggered* set of values, since the ordering is the actual guarantee. Full suite green.

PR #948, opened 2026-08-30
feat(shopping): send a shopping list to a household member (#944)

ulsklyc:
The second half of #944, and the last piece of that issue. Stacked on #945 because it uses the `logLabel` mechanism introduced there - the subject carries the list name, which is user text and does not belong in the server log. **Merge #945 first**, then this rebases onto main cleanly.

An entry in the list's overflow menu mails its open items to one household member, grouped by category in the same order the screen shows them - that order models the route through the shop, so alphabetical would send someone down the same aisle three times.

## A snapshot, not an access route

No link, no token, nothing that outlives the message. A read-only share URL was the other obvious shape and is deliberately not built: it would be the first unauthenticated view of household data in Yuvomi, and a leaked link stays leaked. Anyone who needs the list continuously is a household member and already has the app.

Because a snapshot goes stale the moment someone at home ticks an item off, the mail states which moment it captured rather than presenting itself as current. Whoever is carrying it around the shop cannot see what is happening in the kitchen.

## The recipient is an id, never an address

If this route accepted an address from the request body, the instance would be an open mail relay for anyone with a login: arbitrary content to arbitrary recipients, sent through the household's SMTP server and against its reputation. The server resolves the address itself, from the same source the password reset uses.

That lookup used to be a one-liner inside `auth.js`. It now lives in `services/member-email.js` and both callers go through it, so "how do I reach this member" keeps exactly one answer rather than growing a second.

## Other decisions worth knowing

- **Sending to yourself is allowed** - that is the "get the list onto my phone" case - and then the mail drops the "X sent you this list" line.
- **Only members with an address appear in the picker.** A name you can click that then fails is a promise that does not hold. The route rejects the same case again anyway.
- **Its own rate limit of 10/min**, separate from the API limiter's 300. That number is right for reading and ticking off, and far too generous for something that puts mail in someone's inbox.
- **Three refusals, three messages** - no address, no SMTP, nothing open. Each also carries a machine-readable `reason`, because all three are 422 and the message text is English like every server message here. Without it the interface could only express the distinction in the server's language, and the whole point of separating them would stop at the language border. (Found by running it, not by reading it: the German UI answered in English.)
- **The confirmation is deliberately not a `success` toast.** Those are permanently silenced after 50 (`TOAST_SUCCESS_MAX`) - right for actions whose result is on screen and that you repeat daily. A send is rare, cannot be taken back, and its result lands in someone else's mailbox.

## Verified by running it, not only by testing it

Fresh database, seeded demo data, real browser: the menu entry, the dialog, the empty case with no reachable members, the localised SMTP refusal, and an actual send against a throwaway SMTP catcher. The delivered mail carries sender, list name, shop order, quantities, correct umlauts, and the timestamp in the household timezone.

Two things surfaced only there: the English refusal above, and confirmation that the transport timeouts from #945 work - an unresponsive catcher aborted after 5 s instead of hanging.

`test:shopping-send` has 12 tests. Every guard was individually disabled and confirmed red, including the one that matters most: `email`, `to` and `toAddress` are all sent in the body and the test asserts that none of them reaches the message.

No migration. No new environment variables. Docs updated: SPEC, README, test-suites, CHANGELOG, OpenAPI.


ulsklyc:
**On the missing CI run here.** This PR targets `feat/944-email-notification-channel`, and the workflow is scoped to `pull_request: branches: [main]` - so the test job does not fire on a stacked PR. That is the repository's existing behaviour, not something missing from this branch.

The full suite was run locally on this exact tree and is green, including the suites this touches: `shopping-send` (12), `shopping-routes` (52), `shopping` (40), `i18n` (53), `i18n-plural` (14), `openapi-coverage`, `openapi-structure`, `password-reset` (14) and `invites` (37) - the last two because `emailFor()` moved out of `auth.js` into the shared `services/member-email.js`.

Once #945 is merged, rebasing this onto `main` gives it a real CI run. Happy to retarget it at `main` before that if you would rather see the checks now - the diff would then show #945's commits as well.

ulsklyc:
All four findings addressed in `e1af7c7`, each reproduced first. The membership one was a real leak - thank you for it.

**A `users` row is not a household member** (Claude, Codex P2). The route asked `SELECT ... FROM users WHERE id = ?` and treated every row as reachable. Two kinds of account sit beside the household and both carry a contact with an address: housekeeping staff, and shared-expense guests - who are external, blocked from every other `/api/v1/*` route, and still given a contact row by the guest sync. The picker sourced from `/family/members`, which only excludes staff. So an external guest appeared in the dropdown and would have received the list name, item names and quantities.

Reproduced before touching anything:

```
memberEmail(guest)   = gast@fremd.example    <- external
/family/members      = [Mama, Externer Gast] <- in the picker
```

The predicate now lives once in `member-email.js` and is used by both sides: the new `GET /shopping/send-recipients` for the picker, and the route for its own check. Two copies were the actual bug - a boundary drawn only in the interface is not a boundary. The test asserts both directions: no external is accepted, and everyone offered is accepted.

Your observation that `listEmailableMembers()` existed but was only referenced from the test file was the sharpest part of it. That was the real mistake underneath: the right query was written and then not used.

**A follow-up my own fix caused.** The first attempt folded membership into `memberEmail()` - which broke password reset for guests, and that is expressly intended (`auth.js` names `isSplitExpenseGuest` as its own reason to send). `test:password-reset` caught it. The two questions stay separate; callers needing both ask both. Worth stating because it is the reason this is not simply "filter everywhere".

**Address lists in a contact** (Codex P1). Correct, and the rule was already written in `notification-channels.js` from #945 - just not applied here. `contacts.email` is free text and partly from CardDAV, and nodemailer reads `to` as a list. A member with `a@x.de,external@y.de` is now unreachable rather than reaching both: a reset can be redone through Administration, a sent mail cannot be recalled. Because the check sits in the shared function it now covers the reset path too.

**Stale list after switching tabs** (Codex P2). Also correct, and the most expensive one for an action that cannot be undone - the dialog held the old `listId` and names no list anywhere, so it would have mailed the previous one while the screen showed another. Guarded by re-checking `state.activeListId` after the await. Not covered by a test: staging that race would have cost more scaffolding than the line is worth. Flagging that rather than claiming coverage.

`test:shopping-send` is at 15. Each new guard was disabled individually and confirmed red. One of them was green for the wrong reason at first - the account in the address-list test had no contact at all, so the `UPDATE` hit nothing and `memberEmail()` returned null regardless. The contact is created now and the starting state asserted. Full suite green.

PR #952, opened 2026-08-30
fix(budget): abo-vorgaben sprechen die lesersprache

### LINKED ISSUE #950: [Bug] Names of the payment methods displayed in English regardless the user account language
### Yuvomi version

2.55.0

### Deployment

### Description

The application displays the names of the payment methods always in English, regardless the language the user chose in Settings / Language (i. e. the "Manage categories and payment methods" dialogue).

Image

### Steps to reproduce

1. Sign in
2. Navigate to Settings / Appearance
3. Set the Language to Spanish
4. Sign out and sign in again
2. Navigate to Budget
3. Switch to Subscriptions view
4. Open the "Manage categories and payment methods" dialogue. Subscription categories are displayed in Spanish but the names of the payment methods are displayed in English

Expected:
When the users opens the "Manage categories and payment methods", the names of the payment methods are displayed in the user account language.

Actual:
When the user opens the "Manage categories and payment methods", the names of the payment methods are displayed always in English.

### Browser / device / OS

Firefox 154.0 on Windows 11 25H2

### Logs

```shell

```

### Additional context

_No response_

ulsklyc:
Closes #950

## What was wrong

In "Manage categories and payment methods" the categories read Spanish and the payment methods next to them read English. Same dialogue, two languages.

Both tables store their defaults as English **text** (`'Credit Card'`, `'Entertainment'`). The categories had a map in the frontend that turned that text into an i18n key; the payment methods had none. So this was never a missing second translation - it was the first one sitting in the wrong place. A name map also cannot tell whether `'Other'` is the seeded default or a row the household created under the same name.

## The fix

The project already answers this four times over - `task_categories`, `contact_categories`, `inventory_categories` and the document folders all carry `label_key` and resolve as `label_key ? t(label_key) : name`. Migration 170 moves both subscription tables onto the same shape, and renaming clears the key exactly as the other three routes do.

- 7 payment-method keys × 24 locales, all real translations. `Credit card` and `Other` are taken from the values the project already carries (`budget.accountType_credit`, `budget.subcatSubscriptionOther`) so the same thing is not worded two ways in one screen; PayPal / Apple Pay / Google Pay stay latin everywhere as brand names.
- `DEFAULT_CATEGORY_LABELS` (the name map) is gone.

## The breakdown was inventing English words

`by_category` / `by_payment_method` grouped by display text: `row.category_name || 'Uncategorized'`. That put an English *label* into a data field, which no client could translate - a Spanish UI read "Unspecified". Grouping now happens by **row**; each entry is `{ id, name, label_key, amount }` and the catch-all bucket carries `id: null` and leaves the wording to the reader. Documented in the OpenAPI description.

## Verified

`test:subscription-meta-labels` (new, 4 tests) holds the three promises: the key **exists** in all 24 locales, it **travels** (meta, denormalized on the subscription row, in the breakdown), and renaming **clears** it. Counter-check passes - restoring either cause turns it red.

Live-checked in a Spanish UI: the manage dialogue, the filter, the card and the breakdown all read Spanish, and the catch-all bucket reads "Sin especificar".

| | before | after |
|---|---|---|
| Manage dialogue | Credit Card, Debit Card, … | Tarjeta de crédito, Tarjeta de débito, … |
| Catch-all bucket | Unspecified | Sin especificar |


ulsklyc:
@codex review

PR #953, opened 2026-08-30
fix(calendar): feiertage folgen der datensprache, nicht dem land

### LINKED ISSUE #946: [Bug] OpenHolidays synchronization doesn't follow the language of stored entries
### Yuvomi version

2.54.0

### Deployment

### Description

I've set up public and school holidays synchronisation under Calendar / Public & School Holidays. Country is set to Spain and State/Region to Catalonia, which is where I live.

Image

My user uses English for both the UI language and the language Yuvomi stores the entries it creates itself.

Image

After synchronising with OpenHolidays API, the related entries on the Calendar appear in Spanish.

Image

I assumed that synchronising public and school holidays would have used English because the message above the "Language of stored entries" dropdown states that Yuvomi uses it for API, calendar feed and synchronisation.

Server logs show that the application is querying the API with the right country and region codes. There is no indication whether the application uses a language code when getting the data, though.
{"ts":"2026-08-30T11:15:00.585Z","level":"info","mod":"Holidays","msg":"Holiday sync complete: 35 entries for ES/ES-CT"}

I've checked the data available on openholidaysapi.org for that country and region, and they are available in English and other languages.

https://openholidaysapi.org/PublicHolidays?countryIsoCode=ES&validFrom=2026-01-01&validTo=2026-12-31&subdivisionCode=ES-CT

[...]
{
    "id": "c28833fa-7679-4a40-8a51-a7123bbf60a5",
    "startDate": "2026-12-25",
    "endDate": "2026-12-25",
    "type": "Public",
    "name": [
      {
        "language": "ES",
        "text": "Navidad"
      },
      {
        "language": "CA",
        "text": "Nadal"
      },
      {
        "language": "EN",
        "text": "Christmas Day"
      },
      {
        "language": "DE",
        "text": "Weihnachtstag"
      }
    ],
    "regionalScope": "National",
    "temporalScope": "FullDay",
    "nationwide": true
  },
[...]

### Steps to reproduce

1. Sign with any user account
2. Under Settings / Appearance / Language, set English for both Language and Language of stored entries.
3. Sign out and sing in again
4. Under Settings / Modules / Calendar / Public & School Holidays set Country to Spain and State/Region to Catalonia.
5. Check Show public holidays and Show school holidays
6. Confirm the changes with the Save button
7. Sync the holidays with the Sync now button
8. Go to Calendar. Choose Month view and go to December 2026
9. The public holidays shown are in Spanish (i.e. "Navidad" for "Christmas")

Expected:
The calendar entries for public and school holidays synchronised from OpenHolidays display in the language the user chose for "Language of stored entries" under Settings / Appearance / Language.

If the chosen language is not available from OpenHolidays, use the first language returned by the API or English as fallback.

Actual:
The calendar entries for public and school holidays synchronised from OpenHolidays display in Spanish despite the user chose a different language for "Language of stored entries".

### Browser / device / OS

Firefox 154.0.1 on Windows 11 25H2

### Logs

```shell

```

### Additional context

_No response_

ulsklyc:
Closes #946

## What was wrong

The service derived the sync language from the **country** - a map from `ES` to `ES`. So a household in Catalonia stored "Navidad" even with "Language of stored entries" set to English, while the hint under that very setting promises it affects "the API, calendar feed and synchronisation".

A holiday *is* content Yuvomi stores itself, so it falls under that promise. The language now comes from `resolveHouseholdLocale` - the same source birthdays, loan instalments and notifications use.

## Why the request no longer sends `languageIsoCode`

With that parameter OpenHolidays returns exactly **one** name per holiday, and falls back to the country language when the requested one is missing. The cascade in `resolveName` then ran over a single-element array and had nothing left to choose from: it was handed "Navidad" with no alternative beside it. Without the parameter the full `name` array arrives and the choice is made here, where it is known which language the household reads. The cost is a slightly larger response for a few dozen entries a year.

## The cascade now prefers English over "the first offered"

"Else the first" meant "else the country language" in practice. English is what OpenHolidays carries for nearly every country, so it is the better answer than whatever the server happens to name first. The first offered stays as the last resort.

## A language change re-syncs

The cache holds translated names, not keys - so switching the data language would otherwise change nothing for up to 30 days, and the reporter would have every reason to file it again. `sync_config.holiday_last_sync_language` sits next to the timestamp; a mismatch counts as a forced run. Existing installs get exactly one extra sync after the update, which is what corrects their stored names.

## Verified

Five new tests, and **each of the four promises is proven by its own counter-check**. One of them was green at first while the rule it names was missing: the "English before the first offered" test used a holiday that has no English either, so both versions agreed. The mock now carries a holiday with ES/CA/EN and no DE, which separates the middle step of the cascade from the last.

`resetState()` now also clears `language`/`region` - otherwise one test's expectation would set the next test's precondition.

Live-checked against the real OpenHolidays API with the reported setup (ES / ES-CT):

| data language | 2026-12-25 | 2026-01-06 |
|---|---|---|
| `en` | Christmas Day | Epiphany |
| `de` | Weihnachtstag | Dreikönigstag |

Same 35 entries either way, and switching the language triggers exactly one catch-up run before the throttle takes hold again.

## Not touched

The country/region dropdowns still show English names (`resolveName` default). Those are *display*, not stored entries, so they belong to the reader's UI language rather than the data language - a separate question, and not what was reported.


ulsklyc:
@codex review

ulsklyc:
@codex review

ulsklyc:
@codex review

ulsklyc:
@codex review

ulsklyc:
@codex review

ulsklyc:
@codex review

PR #989, opened 2026-09-01
feat(documents): add safe recursive folder deletion

PechJiri:
## Summary

Add an explicit, previewed choice when deleting a document-folder subtree:

- keep documents and move them to "Without folder", or
- delete the confirmed documents together with the folder subtree.

This PR is the deletion-only slice and remains ordered after directory upload #997, as requested by the maintainer.

## Existing foundation

Nested document folders are already in the data model and UI on `main` via #785 (migration 164, sidebar tree, breadcrumbs, child creation and sibling-unique names). This PR does not recreate that work and does not change the folder schema.

## Confirmation and privacy protocol

`GET /documents/folders/:id/delete-impact` returns:

- the visible folder and document impact across the complete subtree;
- whether the current user may destructively delete every affected document;
- an opaque HMAC snapshot bound to the exact folder, document and collateral-link identities.

Destructive deletion requires that snapshot. If folders, documents or any linked Tasks, Housekeeping, Budget, Inventory or Split Expense records change after preview, the server returns `409 FOLDER_CONTENT_CHANGED` before deleting anything. The token cannot be recomputed from guessed IDs, and impact/error responses do not reveal hidden document identities or counts.

If the subtree contains any document the caller cannot see or delete, destructive deletion is unavailable. The safe keep-documents path remains available and discloses only visible impact.

## Authorization, concurrency and failures

- the user must be allowed to delete every affected document before any storage call starts;
- previewed folder and document IDs are locked against concurrent moves during asynchronous storage deletion;
- PUT, archive PATCH and DELETE against a locked document return `DOCUMENT_DELETE_IN_PROGRESS`, while folder-target conflicts retain `FOLDER_DELETE_IN_PROGRESS`;
- the safe `documents=unfile` request omits the destructive snapshot, so hidden document changes cannot cause a false retry loop;
- all existing-document link writers return `409 DOCUMENT_DELETE_IN_PROGRESS` while a target is in the active batch;
- successful database-row deletion is deferred until the full asynchronous storage pass finishes, keeping link guards reliable for early storage completions;
- Budget and Inventory attachment preflights run before their main mutations, so a rejected link cannot leave a partial update;
- unrelated sibling branches remain available;
- storage, database and concurrent-content failures are reported separately as a partial `207` result;
- the folder structure is retained while failed or concurrently changed documents remain;
- OpenAPI documents both the folder-delete protocol and cross-module link conflicts.

## UI and localization

The preview, keep/delete choices, unavailable destructive option, all-hidden explanation, success and partial-failure messages are present in all 24 supported locale files. Collateral-link wording appears only when linked records are actually affected, and combined failure/concurrency outcomes remain visible.

## Scope split

1. PWA same-version cache invalidation: #994
2. nested folder data model and UI: already merged as #785
3. complete directory upload: #997
4. recursive folder deletion: this PR

There is no directory-upload planner/UI, PWA build-revision code or undo/trash mechanism in this branch. Undo for destructive batch deletion is intentionally being prepared as a separate follow-up PR so this review remains focused.

## Verification

- complete `npm test` passed at `56e52ea3`
- focused folder-deletion and Documents UX suites passed, including explicit safe-unfile request-shape coverage
- independent Superpowers review: approved with no remaining findings
- the reviewed fix was also propagated and re-reviewed on stacked draft #1016
- `git diff --check 46c2c065..56e52ea3`
- no deployment performed

Physical iOS Safari was not available. The mobile behavior was covered by repository automation and static frontend checks only.

Related: #785, #994, #997


PechJiri:
Following the same review, I have split this PR and force-pushed the smaller foundation with an explicit lease.

What changed:

- #989 now contains only the end-to-end recursive folder deletion workflow: impact preview, keep/delete choice, authorization, identity-bound snapshot, concurrency locks, partial-failure handling, OpenAPI, all 24 locales and tests.
- The diff is now 31 files / 982 additions; 24 changed files are locale catalogs.
- Directory upload is preserved on `PechJiri:feat/directory-upload` at `b3972ab9`, based directly on this PR. I will open it after #989 lands so GitHub shows only the upload delta.
- Same-version PWA cache invalidation has moved to the completely independent #994.

Fresh verification:

- full `npm test` passes on Node.js 22.23.2 at `46293731`
- 25 folder-route tests and 36 Documents UX tests pass
- boundary checks confirm this PR contains neither directory-upload nor PWA build-revision files

The original complete trees remain recoverable locally while review is in progress; no deployment was performed.

ulsklyc:
@PechJiri This direction is welcome, and the write-up is unusually good: the
207/409 behaviour, the identity-bound snapshot, the "a failed folder blocks only
its descendants" rule are all things I would otherwise have had to ask about.

Three things before the review, and only the third is a request.

**The green check is stale.** CI passed on an earlier revision; the PR is
conflicting against `main` as of now. That is on me rather than on you, and it is
not something I want you to chase - see below.

**The title undersells what this actually does.** Nested document folders already
exist in the data model: migration v164 (#785) gave `document_folders` a
`parent_id` and a five-level limit. This PR does not touch `server/db.js` at all,
which is exactly right - it builds the interface on top of what is already there.
But a reviewer reading "add nested folders" against a schema that already has
them will spend their first hour confused. Worth a sentence in the description.

**The request: split this.** Not because anything in it is wrong, but because
+3450 lines carry four independent decisions, and any single objection to one of
them holds up the other three. Concretely, in the order I would merge them:

1. **`fix(pwa): invalidate same-version acceptance builds`.** This commit touches
   `package.json`, `public/sw.js`, `server/index.js`,
   `server/utils/service-worker.js` and four tests. It does not touch a single
   file under `documents`. Open it on its own and it can land this week, entirely
   independent of the rest.
2. **Nested folder UI**: the sidebar tree, breadcrumbs, creating a child folder,
   sibling-unique names. This is the part with no open questions.
3. **Directory upload**: the picker, the plan preview, conflict handling,
   sequential execution, progress. This is the largest piece and the one that
   deserves a review of its own.
4. **Folder deletion**: the impact preview, the keep-or-delete choice, the
   SHA-256 identity snapshot, the 409 locks. This is the part I will have the
   most questions about, and I do not want those questions sitting on top of the
   folder tree.

To be explicit about what you are not being asked to do: I am not asking you to
rebase, and I am not asking you to resolve the `CHANGELOG` or locale conflicts.
Those come from my release cadence, not from your work, and I resolve them at
merge. I am writing that split into CONTRIBUTING.md so it stops being something
every contributor has to find out for themselves.

PechJiri:
The requested four-way split is now reflected in the branches and PRs:

1. PWA same-version cache invalidation: #994 (independent, 9 files).
2. Nested folder model and UI: already present on current `main` via #785. After rebuilding against current `main`, there is no remaining diff for this seam, so opening another PR would duplicate merged work.
3. Directory upload: #997, rebuilt directly on current `main`; it has no deletion API and no PWA build-revision code.
4. Folder deletion: this PR, deletion-only and now marked draft so it is reviewed last.

#989 remains 31 files / +982 / -63, with 24 of those files being the supported locales. #997 is the large upload-specific review slice (planner, preview, conflict policy, sequential executor and its locale/test surface).

Fresh verification:

- #997 complete `npm test` passed; Node 22 focused upload/UX/precache tests pass (17 + 39 + 7).
- #989 complete Node 22 `npm test` and focused deletion route/UX suites pass.

No deployment was performed.

ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**Last, exactly as you proposed.** It is the destructive operation in the set and deserves the most careful reading, so it gets the slot where nothing is queued behind it.

**What to expect.** After #997 lands, this branch will conflict with it in `CHANGELOG.md`, `docs/test-suites.md` and `test/test-documents-ux.js`. Against current `main` it is only the CHANGELOG head, and it conflicts with no other open PR. There is no reason to rebase now; it is worth doing once, after #997 is in.

**What it needs from me.** CI approval, the run finished as `action_required`.

The confirmation protocol with the SHA-256 subtree snapshot and the 409 on drift is the right shape for this. When it is ready for review I will read it specifically against concurrent moves and partial storage failures.

ulsklyc:

ulsklyc:

ulsklyc:

ulsklyc (changes requested):
Reviewed, and thank you for keeping this one in draft until last - it deserved the slot.

**What holds, because it is the part I was most prepared to doubt.** The confirmation protocol is binding, not decorative. `documents=delete` without `expected_snapshot` is refused; an empty string, a `null` and a valid 64-hex hash belonging to a *different* folder are all refused; and equal counts with different identities are refused too, because `folderDeleteSnapshot()` hashes the sorted id arrays rather than the totals. Both tables use `INTEGER PRIMARY KEY AUTOINCREMENT`, so a stale snapshot cannot become valid again through rowid reuse. Authorization does precede the first storage call. The `unfile` branch preserves every document, including ones the caller cannot see. Two concurrent destructive deletes of the same subtree: the first wins, the second gets 409. That is a well-built protocol.

## Blocking: the destructive path deletes what the single-document path refuses to show

`server/routes/documents.js:761`:

```js
if (deleteDocuments && !isAdmin(req)
    && documents.some((document) => document.created_by !== userId(req))) {
```

The chain, each link read rather than inferred:

1. `:588` selects the subtree with `SELECT id, created_by FROM family_documents WHERE folder_id IN (…)` - no `documentVisibleSql()`.
2. `!isAdmin(req)` skips the ownership check entirely for an admin.
3. `documentVisibleSql()` (`server/services/document-access.js:21`) has exactly three branches - `created_by`, `visibility = 'family'`, and an explicit `family_document_access` grant. There is no admin escape hatch.
4. `DELETE /:id` goes through `getVisibleDocument(id, req, true)`.

So an admin who asks to delete a member's `private` document individually gets **404 - the document does not exist for them**. The same admin deleting the enclosing folder removes it permanently: row and stored file, no undo, no trace for the owner.

**This is a maintainer decision and I have made it: the visibility rule stands, and admins do not override it.** Yuvomi is a household planner, not a company tool - the admin here is usually a parent, and the other members are partners and children with their own privacy. Somebody who marks a document private did so trusting that "private" means private; narrowing that promise silently on update is the #584 trap. `family_document_access` already exists for the case where an owner wants to share one document deliberately.

Please select the subtree through `documentVisibleSql()` and refuse with 403 as soon as **one** row in it is invisible to the caller, so the destructive mode can never exceed what the single-document mode allows. An admin who cannot see a document still has `documents=unfile`, which moves everything to "Without folder" and destroys nothing.

If we ever decide admins *should* override visibility, the place for it is `document-access.js` - one rule, all paths, one test - not an `isAdmin` at a single call site. That question is open in #1007 and I would rather answer it there.

## Should fix

**1. The impact endpoint leaks a count over invisible documents** - `:588` and `:595`

`documents: documents.length` and `can_delete_documents` are computed over the same unfiltered select. A member gets `0` rows from `GET /documents?folder_id=F` but `{"documents": 1}` from the impact endpoint for that folder - a free, repeatable probe for the existence of other people's private documents, and an unexplainable "Documents: 1" in a dialog over a folder that looks empty. Count through `documentVisibleSql()` and express the rest as `can_delete_documents: false` (or a separate `hidden_documents` flag) rather than as a raw total.

**2. The preview does not preview the collateral damage**

`family_documents` is referenced **ten** times: five `ON DELETE CASCADE` and five `ON DELETE SET NULL`. Deleting a folder with `documents=delete` therefore also strips calendar event attachments, housekeeping work-session receipts, settlement proofs and expense-group avatars, and cascades task, budget-entry, expense and inventory attachments - while the dialog promises only "Documents: N". Accurate impact is this PR's whole reason for existing; this is the part it does not show. Counting the linked rows and naming the affected modules in the confirmation would close it.

**3. The deletion lock is asymmetric** - `:642` versus `:613`

`POST /folders` guards the move *target* (`activeFolderTreeDeletes.has(vParent.value)`), but `PUT /folders/:id` guards only the folder's own id. So while a destructive delete of G is running, `PUT /folders/<other>` with `{parent_id: G}` succeeds and moves an unrelated folder into the dying subtree. Add the symmetric target check to `PUT /folders/:id`, and the same for `folder_id` on `PUT /:id`.

**4. Three plural keys have no `_one`** - all 24 locales

`deleteFolderKeepDocuments`, `deleteFolderWithDocuments` and `folderDeletedWithDocumentsToast` interpolate a numeric `{{count}}` and ship no singular variant. CLAUDE.md requires one; `t()` falls back to the base key, so German and English read clumsily and languages with mandatory agreement read wrongly.

**5. A test that cannot prove what it is named for** - `test/test-document-folders.js:213`

`'DELETE with documents rejects a member before deleting documents owned by somebody else'` seeds exactly one document, owned by someone else. A check placed *inside* the loop produces the identical 403 and the identical surviving row - so the test cannot distinguish a pre-loop gate from an in-loop one, which is the single property it claims. Seed several of the caller's own documents plus one foreign, and assert every own row *and* its stored file survives.

**6. The snapshot suite never tries a wrong snapshot.** No case covers a snapshot from a different folder or a malformed one, and none covers the admin/foreign-private case above. All three rejections return 400 or 409, so assert the *message*, not just the status.

**7. Czech test names and assertion messages** - `test/test-documents-ux.js:345` and two others. Every surrounding test is German; when one of these fails, the output that reaches me is unreadable.

**8. A partial-failure toast that names zero failures** - `public/pages/documents.js:1125`

When the 207 comes from `contents_changed` rather than a storage error, the toast reads "Documents deleted: 60. Could not delete: 0. The folder was kept." Branch on `result.contents_changed` and say that somebody changed the folder while it was being deleted.

## Nice to have

- `:1142` - the retry cannot tell the staleness 409 from the "folder is currently being deleted" 409, so the dialog silently reopens with the same numbers for as long as the other request holds the lock. Reading `err.data?.code` and showing a toast for the second case would explain it.
- `:784` - the row delete sits in the same `try` as the storage call. Storage-first is the right order, but if the row delete throws after the bytes are gone, the user is told the document could not be deleted when it already was. A distinct `storage_code` would keep the report honest.
- Single and bulk document deletion elsewhere in this module go through `scheduleUndoableDelete`. The new destructive folder delete, which can remove hundreds at once, has no undo window - the preview is the only brake. Worth considering, not blocking.
- `deleteFolderConfirmDetail` and `deleteFolderSubtreeDetail` (+ `_one`) lose their last caller in all 24 files.

The rebase and the CHANGELOG are mine when this lands.

PechJiri:
Implemented every blocking and nice-to-have item from the maintainer review, followed by independent Superpowers review rounds and fixes.

### What changed

- Replaced the client-reproducible subtree digest with an opaque, domain-separated HMAC bound to exact folder, visible document and collateral-link identities.
- Destructive deletion is unavailable when the subtree contains any document the caller cannot see or delete. Counts and errors no longer reveal hidden document identities or totals.
- Snapshot validation now detects changes in linked Calendar, Housekeeping, Tasks, Budget, Inventory and Split Expense records, not only folder/document IDs.
- Added process-local locks for the full destructive batch and applied them to folder/document moves, direct document update/delete/archive, and every existing-document link writer.
- Changed the destructive pass to finish all asynchronous storage deletions before synchronously removing successful database rows. This keeps early storage-completed documents visible to authorization checks and reliably returns a conflict to concurrent link attempts.
- Moved Budget and Inventory attachment preflight ahead of their main mutations, so `409 DOCUMENT_DELETE_IN_PROGRESS` cannot leave titles, amounts, tracked dates or reminders partially changed.
- Added stable `409` documentation to every affected OpenAPI operation.
- Improved the deletion dialog and result handling for all-hidden subtrees, conditional collateral warnings, combined concurrency/storage failures and precise partial outcomes.
- Updated all 24 locales, documentation and the test-suite catalog.

### Review fixes after the first pass

The independent review found and we fixed:

1. an early-row race during a still-running storage batch;
2. partial Budget/Inventory updates before a rejected attachment;
3. missing cross-module OpenAPI `409` responses;
4. archive/restore bypassing the active document deletion lock.

The archive regression was independently demonstrated RED on the prior head (`200` instead of `409`) and GREEN after the fix.

### Verification at `46c2c065`

- full `npm test`: passed, exit 0;
- document-folder route suite: 31/31;
- API/OpenAPI focused suite: 13/13;
- Budget route suite: 52/52;
- Inventory route suite: 17/17;
- the concurrency regression passed repeatedly;
- `git diff --check`: clean;
- final independent Superpowers review: APPROVE, zero remaining findings.

Physical iOS Safari was not available. Mobile behavior was covered by repository automation and static frontend checks only. No deployment was performed.

I did not rebase, renumber migrations, touch release metadata or resolve maintainer-owned `main` drift, in line with CONTRIBUTING section 6.

PechJiri:
The requested Undo follow-up is now available as draft #1016.\n\nI kept it separate from this PR so the core recursive-delete API remains reviewable on its own. #1016 adds only the five-second client-side Undo lifecycle on top of this branch; it does not add server trash, retention, a migration, or another API. Its incremental diff is 8 files (+577/-48): https://github.com/PechJiri/yuvomi/compare/feat/folder-upload...feat/folder-delete-undo\n\nThe follow-up passed the full npm test suite at 31ee62dc and two separate review rounds. The final round is APPROVE with no remaining findings. It stays draft and explicitly depends on #989; once this PR lands, the upstream diff will collapse to the focused Undo delta.

ulsklyc (changes requested):
Re-read at 46c2c065. The blocking item is resolved the way it was asked: the subtree is selected through `documentVisibleSql()` (`server/routes/documents.js:800-812`) and refused with 403 as soon as one row is invisible (`:838-843`), before the first storage call; `document-access.js` is untouched, and the new test "delete impact does not count a hidden private document and admins cannot delete it through a folder" pins the admin-over-private case. All eight "should fix" items are in with code and tests; I checked the collateral count against every `REFERENCES family_documents` in `server/db.js`, and the eight link tables are all covered. The chain runs clean on the branch apart from `test:changelog`, which only misses the two releases tagged since the branch was cut.

Three small things before it leaves draft, none of them about data safety:

1. `PUT /:id`, `PATCH /:id/archive` and `DELETE /:id` answer a locked document with `reason: 'FOLDER_DELETE_IN_PROGRESS'` (`deletionInProgress()`, `:65-71`), while the OpenAPI entry for `/archive` (`server/openapi/paths/documents.js:324`, via `documentDeleteConflict` in `server/openapi/helpers.js:74`) and `test/test-api.js:270` promise `DOCUMENT_DELETE_IN_PROGRESS`, and the lock test in `test-document-folders.js:850` asserts the folder reason for the archive route. One reason per lock: the document lock from `document-deletion-lock.js` carries its own, the folder reason stays for folder targets.
2. The client sends `expected_snapshot` for `documents=unfile` as well (`public/pages/documents.js:1138`), and the snapshot hashes all document ids including hidden ones (`:637-640`). A private document added by somebody else between preview and click turns the safe path into a 409 over a folder that looks unchanged to the caller, and the dialog reopens with the same numbers. Bind the snapshot to `delete` only, or send just the two counts for `unfile`.
3. `docs/SPEC.md:3088` and `:3768` still say deleting a folder keeps its documents. That is now the default, not the rule; I will take it together with the CHANGELOG and the rebase.

The ownership check at `:649` and `:842` mirrors the one `DELETE /:id` already has inline at `:1267`; folding those five sites into one predicate in `document-access.js` is a follow-up on `main`, not on you.

The rebase is mine and mechanical: the CHANGELOG head, one hunk in each of the 24 locale files where your removed `deleteFolderSubtreeDetail` keys sit next to the share keys from 6d8a4a99, and the test block at the end of `test-documents-ux.js`; `documents.js` merges clean. This stays last in the published order, after #997.

PechJiri:
Implemented all remaining maintainer follow-ups through `56e52ea3`.

### Conflict semantics

- PUT, archive PATCH and DELETE now return `DOCUMENT_DELETE_IN_PROGRESS` when the specific document is locked by a destructive folder operation.
- Folder-tree and folder-target conflicts continue to return `FOLDER_DELETE_IN_PROGRESS`.
- Visibility and authorization checks still run before lock-state disclosure.

### Safe versus destructive deletion

- The safe `documents=unfile` path no longer sends `expected_snapshot`.
- Hidden document changes therefore cannot trap an allowed keep-documents action in a false conflict/refresh loop.
- `documents=delete` remains bound to the opaque preview snapshot and exact confirmed destructive set.
- Added explicit regression coverage for the safe-unfile URL shape in addition to route-level concurrency coverage.

### Stacked Undo branch

The same reviewed semantics were propagated to draft #1016. Its delayed and `pagehide` destructive commits remain snapshot-bound, while direct unfile remains snapshot-free. The expected conflict was resolved without replacing the Undo orchestration, and both branches were independently re-reviewed.

### Verification

- Full `npm test`: passed at exact head `56e52ea3`.
- Focused folder deletion and Documents UX checks passed.
- `git diff --check 46c2c065..56e52ea3`: passed.
- Independent Superpowers review: APPROVE, no remaining findings, including the optional test hardening.

The PR description now documents the distinct document/folder conflict reasons, safe-unfile protocol and exact verification evidence per CONTRIBUTING.

ulsklyc:
Two corrections to my own notes in this thread, measured today rather than remembered. Both notes went to every PR in the queue, so the same correction goes to each.

The rest of both notes holds as written, including the CHANGELOG conflict this branch had against `main` that morning.

ulsklyc (approved):
Re-reviewed at 56e52ea3. All three points are in, each checked at the code: `DOCUMENT_DELETE_IN_PROGRESS` now comes from one helper (`server/routes/documents.js:75-78`) on `PUT /:id`, the archive `PATCH` and `DELETE /:id` (`:1119`, `:1172`, `:1275`), while folder targets keep `FOLDER_DELETE_IN_PROGRESS`; `expected_snapshot` is sent and required only for `documents=delete` (`public/pages/documents.js:1135-1141`, server `:817-822`), with the counts still bound on both paths; and the blocking point from the first round holds at this head: the subtree is read through `documentVisibleSql()` (`:806-814`), the 403 comes before the lock and before the first storage call (`:848` before `:857` and `:870`), `document-access.js` is untouched, and `test-document-folders.js:401` pins the admin-over-private case. I ran the whole chain on the branch: red only on `test:changelog`, which is the branch's age; the 154 suites after it are green when run on.

Approved. Mark it ready for review when you are, and it lands in the published order, after #997.

What stays with me at merge time, so you do not need to touch it: the two `docs/SPEC.md` sentences that still say a deleted folder keeps its documents (`:3122` and `:3837` on `main`), the CHANGELOG position, and the 24 locale files, where your removal of `deleteFolderSubtreeDetail` meets the share keys `main` added at the same spot.

Two things I noticed that I will take as follow-ups on `main` together with the ownership predicate, not as another round here: the `delete-impact` snapshot hashes every document in the subtree including ones the caller cannot see (`:644-646`, `:666`), and the `FOLDER_CONTENT_CHANGED` check (`:835-843`) runs before the 403 (`:848`), so a member can tell that something hidden changed between two calls. It reveals nothing beyond what `can_delete_documents: false` already says, which is why it is a follow-up. And `overlapsActiveDeletion` (`:853-854`) also fires for `documents=unfile` while an owner deletes a private document in that tree, a window the length of one storage delete.

ulsklyc:
Merged as 127b6278. As with #991, the merge commit was created from your head (56e52ea3) with main merged in and the full chain run on that exact tree, because a push into the fork branch is not possible from this side; your branch is untouched and GitHub shows the PR as merged.

What the merge had to resolve, for the record: 24 locale files, the CHANGELOG, `docs/test-suites.md` and `test/test-documents-ux.js` collided with #997 (folder upload) and the share sheet from D#1014. Taking main's side of the locale hunks brought back `deleteFolderSubtreeDetail` and its `_one` variant, which this PR retires; your own test caught that, and both keys are gone again in all 24 files. SPEC now describes the two-step deletion at the folder browser and in the destructive-dialog rule.

The two small follow-ups noted in review stay with me on main, not with you: the impact snapshot hashing invisible documents (and the 409 arriving before the 403), and `unfile` running into the lock window of somebody else's single delete. Thank you for staying with this one through three rounds.

PR #991, opened 2026-09-01
feat(notes): add category data model and API

PechJiri:
## Summary

Add the note-category data model, permission capability and API. This PR intentionally stops before the browser UI; that is kept on the follow-up branch `PechJiri:feat/note-categories-ui`.

The schema prerequisite #996 has now landed. The maintainer rebased this branch onto the updated main line, and the review corrections below were then applied as fast-forward commits on top without overwriting that work. The PR remains a focused backend/API slice; its browser UI stays on the follow-up branch `PechJiri:feat/note-categories-ui`.

## Category model

- a note may have any number of categories;
- categories start empty and are created by users;
- personal categories belong to one user and are visible only to that user, even on a shared note;
- household categories have no owner and are visible to everyone with access to Notes;
- deleting a user removes that user's personal categories and assignments but keeps shared notes;
- deleting a household category removes all of its assignments;
- the same name may exist once per personal owner and once in the household scope;
- there is no seeded category data.

## Permissions

Every user may create, rename, reorder and delete their own personal categories.

Managing the shared household catalog uses the `notes_manage_household_categories` capability:

- admins always have it;
- other users default to denied;
- admins may grant it to a family role or an individual member;
- an individual `none` override can revoke an inherited role grant;
- users without it may still assign existing household categories to notes, but cannot create, rename, reorder or delete them.

The generic storage support for capability rows is isolated in #996. This PR registers and enforces the concrete Notes capability.

## API and consistency

- category list and CRUD endpoints are documented in OpenAPI;
- the public contract uses only canonical `id` and `scope` fields, without legacy `key`/`type` aliases;
- note create/update accepts category IDs;
- note responses hydrate only categories visible to the current user;
- editing a shared note never removes another user's hidden personal assignment;
- stale or inaccessible category IDs reject the whole assignment update;
- category hydration is chunked below conservative SQLite parameter limits;
- deleting a category also prunes it from stored Notes dashboard-widget filters.

## Unicode identity without a dependency

The previous `unicode-case-folding` production dependency and vendored browser copy have been removed. Category identity now uses built-in NFKC normalization followed by locale-independent upper/lower casing in both runtimes. This keeps canonical equivalents and ordinary Unicode case equal, including `Straße`/`STRASSE`, without adding a runtime package. The UI follow-up exhaustively compares the browser and server key for every Unicode scalar value.

## Split and sequencing

1. #996 - generic `capability` resource type in `access_permissions`, now merged
2. this PR - note-category model, permission registration and API
3. `PechJiri:feat/note-categories-ui` - category management, picker, AND filtering, dashboard widget and all locales; PR to be opened after the backend base lands

#993 and #996 have landed. The maintainer rebased this branch onto current `main`; the generic capability migration remains v175 and the Notes category migration is now v176.

## Verification

- complete `npm test` passed at `08156acb`
- focused Notes/category tests: 33 passed
- focused permissions and permission-route tests: 40 passed
- i18n and plural checks: 67 passed; all 24 permission labels are present and non-empty
- migration append-only, schema mirror, changelog and OpenAPI checks passed
- independent Superpowers review of the post-rebase range: approved with no remaining findings
- `git diff --check 5f434bbf..08156acb`
- no deployment performed

Related: #993, #996


ulsklyc:
@PechJiri Thanks for this, and for #989 alongside it. Two things before anyone reads the diff.

**CI runs on your PR now, and it is red.** That was on me: workflows from forks wait for a maintainer to approve them here, and yours had been waiting since this morning. Approved now, and it turns out the suite has something to say.

The failure is in a test you added, `test/test-category-manager-lifecycle.js:54`: `assert.equal(deleteStarted, true)` gets `false`. The cause is one line above the code you are exercising - `CategoryManager._delete()` awaits `confirmOverModal()` before it ever reaches `api.delete`, so with only `api.delete` stubbed the flow stops at the confirmation and never starts the request. Stubbing the confirm (or injecting it) should get you past it. Worth checking whether the other two cases in that file pass for the same reason or by accident.

**The other thing is size.** #991 is +3223 lines across 62 files and #989 is +3450 across 42. Together that is more than everything else currently open in this repository. Two consequences, neither about quality: a change that size is hard to review in one pass, and it goes stale while it is being read - both of yours were already 26 and 36 commits behind main within hours of being opened, because this repository merges and releases several times a day.

If you can split them along their own seams - the data model and migration first, the UI on top, the upload handling separately - each piece can be read, merged and released on its own, and you stop rebasing the whole thing every time something lands. I would rather review four small ones this week than two large ones next week.

Neither of these is a rejection. The direction of both is welcome; I am asking about the shape they arrive in.

PechJiri:
Hi @ulsklyc 

Thanks a lot for the detailed feedback and explanation on the test failure—that makes total sense, and I appreciate the guidance!

To give a bit of context: I love this tool and am setting it up for family management. I ran into a real need for uploading document structures (specifically importing a house project folder containing 117 PDFs), so I jumped straight into building it locally. Coming at this more from a product perspective and relying on AI coding assistants (like Codex), I focused on getting the feature working rather than PR scoping and merge strategy.

I will fix the `confirmOverModal()` test stubbing first, and then clean up the work by splitting everything into five focused PRs:

* **#991 – Note categories backend:** Database model, migration, API, permissions, OpenAPI spec, and backend tests.
* **Follow-up – Note categories UI:** Category selection, management, filtering, widget, permission settings, translations, and frontend tests.
* **#989 – Folder structure operations:** Navigation, folder tree, deletion, API, and safety locks.
* **Follow-up – Folder upload:** Directory selection, planning, recursive upload, and tests.
* **Standalone micro-PR – PWA cache invalidation:** Moving this cache fix into its own isolated PR since it's unrelated to documents.

On a related note: having repo-level AI guidelines/guardrails (like an `.agent` file, `AGENTS.md` or so`) defining PR size limits and contribution norms would be super helpful. If Codex could read these conventions directly from the repository, it would automatically scope and break down changes into smaller reviews from the start.

Thanks again for the direction—I'll start submitting these in smaller chunks!

PechJiri:
Thanks for the review. I have reshaped this PR around the requested data/API seam and force-pushed it with an explicit lease.

What changed:

- #991 is now only the category migration, server-side model and visibility rules, capability storage/resolution, Notes category API/OpenAPI, and their tests.
- The diff dropped from 62 files / 3,224 additions to 14 files / 1,159 additions.
- The complete UI is preserved on `PechJiri:feat/note-categories-ui` at `57c53d31` and is based directly on this core commit. I will open it after #991 lands so its PR shows only the UI delta.

On the failed lifecycle test: `test-browser-loader.mjs` already stubs `confirmOverModal()` to resolve `true`. The flaky assumption was the single event-loop turn before `api.delete`; the UI-branch test now waits on an explicit promise signalled by the delete stub and includes a comment making the loader behavior visible. All three lifecycle cases pass on Node.js 22. The test is not in this core PR because it belongs to the UI slice.

Fresh verification:

- full `npm test` passes on Node.js 22.23.2 at `ec1f71b5`
- `git diff --check` passes
- no folder-upload, category UI, locale or service-worker files remain in this PR

PR #993 still has the independent migration-number collision noted in its description; whichever migration lands second will be rebased and renumbered.

ulsklyc:
@PechJiri Follow-up now that I have read the diff rather than just the CI result.

**The red test is fixed, by you, and CI is green.** Your
`test(categories): wait for delete start deterministically` commit replaced the
`setImmediate` plus `assert.equal(deleteStarted, true)` pair with a promise that
`api.delete` resolves. That was the right fix: `_delete()` reaches `api.delete`
only after `await import('/components/modal.js')` and the confirmation, so a
single macrotask was never guaranteed to be enough. Nothing further needed there.

**Two things in here are not note categories, and both should leave the PR.**

1. **The `access_permissions` rebuild.** Migration 174 renames the table,
   recreates it, copies the rows and drops the old one. That is a change to a
   core permissions table riding along inside a notes feature. If anything about
   it needs a second look, it holds up the categories with it, and if it lands
   quietly it is the kind of change nobody remembers reviewing. Its own PR, with
   its own description of what changed in that schema and why.

2. **`unicode-case-folding` as a production dependency**, plus a vendored copy
   under `public/vendor/`. The dependency policy here is deliberate and the
   frontend one is strict, so a new runtime dependency is a decision rather than
   a detail. I am not saying no. I am saying it needs to be its own question,
   with the alternative spelled out: what does full Unicode case folding catch in
   a category name that a simpler normalisation would miss, and how often?
   Answered on its own, it is a five-minute call. Answered inside a 3200-line
   diff, it holds up everything around it.

**One thing you do not have to act on:** this PR and #993 both add
`version: 174`, and `main` is at 173. Whichever lands second gets renumbered at
merge, by me. Do not renumber them against each other, or you will do it twice.

What is left after those two come out is the actual feature, and that part I
expect to be able to review in one pass.

PechJiri:
Thanks - both requested seams are now addressed.

- The `access_permissions` rebuild is isolated in #996 (3 files, 87 lines, no concrete capability and no UI). This PR is marked draft and stacked on that exact commit; after #996 lands, the rebuild drops out of this diff and only the Notes feature remains.
- `unicode-case-folding` is gone from both `package.json` and the browser assets. The replacement is built-in NFKC plus locale-independent upper/lower casing, which still treats examples such as `Straße`/`STRASSE` as the same name. The UI follow-up has an exhaustive browser/server parity test over every Unicode scalar value.
- The category migration is now separate from the permission migration in the stacked history. I have not tried to pre-empt the #993 migration number; it will be rebased against whichever migrations are on `main` after #996 lands.
- The browser UI remains on `PechJiri:feat/note-categories-ui` and now has no vendored case-folding library. Its lifecycle test explicitly relies on the confirmation stub and waits for `api.delete` to start, rather than assuming one event-loop turn.

Fresh local verification on Node.js 22.23.2: the complete `npm test` chain passes for both the backend/API branch and the UI follow-up.

New prerequisite: #996

ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**Please renumber the two migrations from 174 and 175 to 175 and 176.** #993 takes 174. Once #996 lands, its table rebuild drops out of this diff as you described and only the note-category slice remains.

Against current `main`, and against every open PR except #993, this branch merges completely clean. Its only conflict anywhere is with #993 in `server/db.js` and `server/db-schema-test.js`, which the renumbering resolves.

**What it needs from me.** The CI run finished as `action_required` and is waiting for approval; approval is per push, so the renumbering commit will need a fresh one.

**One question before this lands.** This slice deliberately stops before the UI, which stays on `feat/note-categories-ui`. Is the intention that the backend lands on `main` on its own, or would you rather the two were reviewed together? I have no objection to the split, I just want to be explicit about merging an API that nothing in the shipped app calls yet.

ulsklyc:

ulsklyc:

That includes the migration renumbering I asked you for, and the draft status: neither is blocking anything on your side. Migration numbering is on the mechanical list by name. The one thing still genuinely with you is the question in my earlier comment - whether the backend slice is meant to land on `main` ahead of its UI, or whether you would rather the two were reviewed together.

PechJiri:
Decision on the remaining question: the backend slice in this PR is intentionally meant to land on `main` independently, after its prerequisite #996. The UI will remain a separate follow-up PR on top of it.

The short period with an unused API is acceptable: the backend change is backward-compatible and does not expose an incomplete user-facing feature. Keeping the UI separate preserves the reviewable seams requested earlier and avoids turning this back into one large change.

ulsklyc:

PechJiri:
Implemented every required and nice-to-have item from the maintainer review, including the inherited #996 compatibility correction.

What changed:
- Category identity now converges for the capital sharp-S edge case: `ẞ`, `ß` and `SS` share one stable key. The helper is idempotent and still uses only built-in runtime behavior.
- Added duplicate and idempotence regression tests; the same algorithm and exhaustive browser/server parity check were carried into the stacked `PechJiri:feat/note-categories-ui` branch.
- Permission replacement is documented and tested per axis: modules/widgets are replaced on every save; capabilities are preserved when omitted and replaced only when explicitly present; `{ modules: {}, widgets: {}, capabilities: {} }` clears everything.
- Route-level coverage now seeds all available axes and proves omission-preserves versus explicit-empty-clears behavior.
- OpenAPI, route comments and the test-suite catalog now describe the actual contract.
- The required AI-assistance disclosure is present in the PR body.

Verification at `0d81dfef`:
- complete `npm test`: passed
- Notes/category: 31 passed
- permissions: 26 passed; permission routes: 14 passed
- OpenAPI coverage/structure and schema mirror: passed
- exhaustive Unicode idempotence: all 1,112,064 scalar values passed
- `git diff --check`: passed

No rebase, migration renumbering, CHANGELOG/version reconciliation or deployment was performed; those merge-time drift items remain with the maintainer per CONTRIBUTING.

ulsklyc (changes requested):
Read the diff at 0d81dfef against my #996 review: all three required points and the nice-to-have are in here as well. The delete in `writeSubjectPermissions` is scoped to `module`/`widget` and only touches `capability` rows when `input.capabilities` is present (`server/permissions.js:477-495`, covered by `test/test-permissions.js:241-282` and `test/test-permissions-routes.js:154-179`); the migration test asserts `updated_at` and the subject index (`test/test-permissions.js:48-87`); the rebuild follows the house pattern. The rest holds up too: the rule lives in `server/permissions.js` and the routes only ask, the `notes` module gate applies before any of it, `notes` gets no `visibility` column so #699 is untouched, and the full chain is green on the branch with `main`'s CHANGELOG overlaid.

One thing the PR body no longer describes correctly: this branch is not stacked on #996's current head. It carries `4526439f` plus its own re-application of the preserve-capability fix as `6c767402`, not `9b04b681`, and a merge of #996 into it conflicts in `server/permissions.js` and `test/test-permissions.js`. Once #996 lands I will rebase this one and drop `6c767402`; nothing for you to do, just flagging why "the rebuild drops out of this diff" will need a rebase rather than happen by itself.

Three things before it leaves draft, all small:

1. A `[Unreleased]` entry in `CHANGELOG.md`. This adds five `/api/v1/notes/categories*` endpoints, a `categories` field on every note and a `capabilities` axis on the permission catalog; CONTRIBUTING asks for a line for that, and the sentence has to come from you even though the merge of the file is mine.
2. `docs/SPEC.md` still says `resource_type: module | widget` (`:2578`) and its Notes table (`:1052`) does not know the two new tables.
3. `PERMISSION_CAPABILITIES` ships `labelKey: 'noteCategories.permissionLabel'` (`server/permissions.js:113`), and no locale has that key. Either add it to all 24 files now or leave `labelKey` to the UI PR.

Should fix, not blocking: `categoryInputError` classifies by `/category|categories/i` on the error message (`server/routes/notes.js:52-54`), so a `no such table: note_categories` would go out as a 400 carrying the raw text; a typed error with a `status` is safer. The routes use manual `BEGIN`/`COMMIT` (`:236`, `:294`) where every other route file uses `db.get().transaction()`. And `key`/`type` in the category payload (`:43-44`, plus `type` as an input alias at `:101`) are aliases for a client component; the `/api/v1` surface should carry one field per meaning, and if the aliases stay they belong in SPEC.md on purpose.

Migration numbers: `main` is at 174, #1018 holds 175 and #1019 176-178, so this pair lands as 179/180 unless the order changes; `test/test-permissions.js:38,57,243` and `test/test-note-categories.js:26-27` hard-code `MIGRATIONS_SQL[174]`/`[175]` and move with it. That renumbering is mine at merge time.

ulsklyc:
Rebased onto `main` at 4158bb7a and force-pushed to your branch (`0d81dfef` → `5f434bbf`), as announced. What changed, all mechanical:

- `4526439f` and `6c767402` are gone from the branch because #996 landed on `main` as eba390c8 with the same two changes; your `6c767402` was byte-for-byte the patch that is already there. Your three commits remain as they were, only re-parented.
- Migration numbers: `main` now holds 174 (name days, #993) and 175 (capability, #996), so the notes migration is **176**, in `server/db.js`, the mirror in `server/db-schema-test.js`, and the `MIGRATIONS_SQL` references in `test/test-permissions.js:38` and `test/test-note-categories.js:26-27`. That corrects what I wrote in the review (179/180): with #996 in and this PR next in the published order, 176 is simply the next free number; #1019 renumbers at its own merge.
- `writeSubjectPermissions` keeps the `{ subjectType }` signature from your feature commit on top of the delete scoping that is now on `main`; nothing else in `server/permissions.js` needed a hand.

The whole chain is green on the rebased branch, 279 suites, `test:changelog` included now that the branch carries the current CHANGELOG. GitHub reports it mergeable. Still open from the review before it leaves draft, and still yours: the CHANGELOG line, the SPEC.md notes tables, and the orphaned `labelKey`. If you have local work on this branch, fetch before you continue; the old head is not an ancestor of the new one.

PechJiri:
Implemented every required and recommended maintainer follow-up on the maintainer-rebased branch, now at `08156acb`.

### Concurrent rebase handling

Before pushing, the remote head had moved from `0d81dfef` to `5f434bbf` because #996 was merged and the maintainer rebased #991 onto current `main`. I did not overwrite that work. The reviewed correction commit was replayed on top as a fast-forward, CHANGELOG/SPEC conflicts retained current-main content, and the final migration references were aligned: capability v175, Notes categories v176.

### API and consistency changes

- Added a typed `NoteCategoryInputError`; route classification no longer depends on message regexes.
- Wrapped note create/update assignment work in route-level database transactions.
- Removed the duplicate public `key`/`type` category aliases from responses, request handling and OpenAPI; `id` and `scope` are canonical.
- Added regressions for canonical payload shape, ignored legacy `type` input and non-input storage errors.
- Preserved hidden personal assignments and all existing authorization behavior.

### Documentation and localization

- Added the user-facing category feature under current `[Unreleased] / Added`.
- Documented both Notes category tables, ownership/privacy semantics and the `capability` permission resource type.
- Added a localized, non-empty household-category permission label to all 24 supported locales.
- Corrected the post-rebase migration heading and test title to v176.

### Verification

- Full `npm test`: passed at exact head `08156acb`.
- Notes/category tests: 33 passed.
- Permissions and permission routes: 40 passed.
- i18n and plural checks: 67 passed; 24/24 permission labels present.
- Migration append-only, schema mirror, changelog and OpenAPI checks passed.
- `git diff --check 5f434bbf..08156acb`: passed.
- Independent Superpowers post-rebase review: APPROVE, no remaining findings.

The PR description was updated to reflect the merged #996 prerequisite, current sequencing and exact verification evidence per CONTRIBUTING.

ulsklyc:
Three corrections to my own notes in this thread, measured today rather than remembered. The two notes went to every PR in the queue, so the first two corrections go to each; the third is specific to this branch.

**Same note, 11:19: "This PR also still conflicts with main on the CHANGELOG head, so a rebase is needed regardless."** Not this one. By `git merge-tree` this branch merged clean against `main` that morning and still did that evening, which my 10:41 note above had already said ("#996 and #991 have no conflict at all"). The sentence was written once for the whole queue and pasted here without checking it against this branch.

ulsklyc (approved):
Re-reviewed at 08156acb, on the rebased branch. All three points are in, each checked at the code: the CHANGELOG entry under `[Unreleased]` (`CHANGELOG.md:76-80`), the two SPEC sections for the category tables (`docs/SPEC.md:1061-1093`), and `noteCategories.permissionLabel` present in all 24 locale files with real translations. The migration sits at 176 as the last array entry (`server/db.js:7042`), mirror and test references match, and the earlier should-fixes are done too: `NoteCategoryInputError` separates input errors from trigger failures (`server/services/note-categories.js:7-13`, `routes/notes.js:51-62`, regression at `test-notes-routes.js:86-103`), both write paths run in `database.transaction()` (`routes/notes.js:233-247`, `:287-310`), and the `key`/`type` aliases are gone from payload, input and OpenAPI. Whole chain on the branch: green, `test:changelog` included, no conflict against `main`.

Approved. Mark it ready for review when you are, and it lands.

One thing I take at merge time, because it is text I wrote for #996: `docs/SPEC.md:2645-2650` still says that no capability is registered by the core and that `resolvePermissions()` reads `module` and `widget` rows alone. Your PR is exactly the feature that paragraph was waiting for (`server/permissions.js:110-116`, `:235-236`, `:400`), so the paragraph turns into a description of what the core does with the first capability. Nothing for you to do there.

Two small things you may leave as they are: the manual `BEGIN`/`COMMIT` fallback in `note-categories.js:121-132` no longer has a caller now that both routes use `transaction()`, and the test title at `test-notes-routes.js:176` is the one non-German title in the file.

ulsklyc:
Merged as 9bd82d6b. The merge commit was created from your head (08156acb) with main merged in and the full chain run on that exact tree (276 suites, green), because a push into the fork branch is not possible from this side; your branch is untouched and GitHub shows the PR as merged. Migration 176 as you numbered it. One thing changed beyond the merge: the SPEC paragraph on `capability` permissions still said no capability was registered by the core, which this PR made untrue, so it now describes `notes_manage_household_categories`, the read path and the save rule (capabilities are replaced only when the body carries the field). Thank you for the work, and for the rebase on the day.

PR #993, opened 2026-09-01
feat(birthdays): add optional name days

PechJiri:
## Summary

Adds an optional name day to each birthday record. A name day is deliberately stored as month and day only, creates its own yearly calendar occurrence, reuses the birthday reminder settings, and is shown alongside birthdays without pretending that it has an age.

The field is optional. Existing records and people without a name day keep their current behavior.

### Create and edit

- The advanced section of the birthday form now contains localized month and day selects.
- The same field is available when creating and editing a person.
- Month selection controls the valid day choices, including February 29.
- A clear action removes the name day without changing the birthday.
- Incomplete month/day input is rejected in the UI, and the API independently validates canonical `MM-DD` values.
- No country-specific or name-based lookup is performed.

### Calendar and reminders

- A saved name day creates a second generated yearly calendar event linked independently from the birthday event.
- The event is marked as `name_day`, so calendar clients can localize its title and description separately from birthdays.
- The name-day reminder uses the birthday record's existing reminder offset, including custom offsets and the no-reminder setting.
- Updating the person's name, name day, reminder settings, or household language keeps both generated artifacts synchronized.
- Clearing the name day removes only its generated event and reminder.
- Deleting the birthday record removes both birthday and name-day events and their reminders, including outbound deletion handling for mirrored calendar events.

### Dashboard, navigation, and birthday list

- The dashboard expands a person into separate birthday and name-day occurrence rows, ordered by the next date.
- Birthday rows continue to show the upcoming age. Name-day rows show `Name day` instead.
- The navigation badge counts upcoming occasions, so one person may contribute both a birthday and a name day.
- The Birthdays module remains one row per person. When present, the row appends the localized equivalent of `in X days - DD.MM.YYYY - celebrates their name day` after the birthday metadata.
- On narrow cards the name-day metadata moves to its own wrapping line to avoid horizontal overflow.

### Data and API

- Additive migration 174 adds nullable `birthdays.name_day` and `birthdays.name_day_calendar_event_id` columns plus lookup indexes.
- `name_day` is stored as `MM-DD`; no synthetic year is exposed through the birthday API.
- Birthday list, upcoming, update, dashboard, calendar list/detail/search, and OpenAPI surfaces carry the new occurrence metadata where relevant.

## Localization

- Added all 12 new name-day UI and calendar strings to every supported locale.
[...]

PR #994, opened 2026-09-01
fix(pwa): invalidate same-version application caches

PechJiri:
## Summary

Makes service-worker cache identity include a build revision as well as the semantic application version. Rebuilding or redeploying the same Yuvomi version can therefore no longer leave an installed PWA on an older application shell.

This was discovered while acceptance-testing the Documents work and has been split out of #989 because it is an independent PWA concern.

## How it works

- The shipped service worker contains a build-revision placeholder.
- `/sw.js` is rendered with a validated concrete revision for each build/deployment.
- Cache namespaces use `APP_RELEASE` plus that revision.
- The response uses `no-store` and CDN-bypass headers.
- Development/test fallback remains deterministic when no external build revision is supplied.

## Scope

- No Documents routes, UI, upload logic or translations are changed.
- No database migration is included.
- Existing API cache clearing and update-bypass behavior is preserved.

## Verification

- [x] Full `npm test` at `9860fb60`
- [x] Focused service-worker, Docker publish and setup coverage: 31 tests, 0 failures
- [x] The isolated reload test leaves tracked `public/sw.js` byte-for-byte unchanged
- [x] `git diff --check f6afbbc3..9860fb60`
- [x] Independent Superpowers review: approved with no remaining findings
- [x] No deployment performed

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -90,5 +90,7 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            APP_BUILD_REVISION=${{ github.sha }}
           cache-from: type=gha
           cache-to: type=gha,mode=max,ignore-error=true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A same-version deployment now invalidates the installed PWA shell.** The served service worker
+  receives a build-specific revision, so acceptance builds and rebuilt images no longer reuse an
+  older cache merely because the application version has not changed.
+
 - **An `allowScripts` pin no longer points at a version that is not installed.** The field names
   every package allowed to run install scripts, with an exact version, because the permission
   applies to the reviewed build rather than to the name. Dependabot raises the dependency and the
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,11 +22,6 @@ RUN npm ci --omit=dev
 # ---- Runtime stage ----
 FROM node:24-slim
 
-# Die Build-Revision kommt aus dem unveränderlichen Git-Commit des Build-Jobs.
-# Sie ist kein Installationswert und wird deshalb nicht über .env gesetzt.
-ARG APP_BUILD_REVISION
-ENV APP_BUILD_REVISION=${APP_BUILD_REVISION}
-
 RUN apt-get update && apt-get install -y \
     gosu \
     && rm -rf /var/lib/apt/lists/*
@@ -52,6 +47,12 @@ ENV BACKUP_DIR=/backups
 # Entrypoint: korrigiert Volume-Permissions und startet als node-User
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
+
+# Die Build-Revision kommt aus dem unveränderlichen Git-Commit des Build-Jobs.
+# Sie ist kein Installationswert und wird deshalb nicht über .env gesetzt. Erst
+# nach den Dateisystem-Layern setzen, damit ein neuer Commit deren Cache behält.
+ARG APP_BUILD_REVISION
+ENV APP_BUILD_REVISION=${APP_BUILD_REVISION}
 
 EXPOSE 3000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,11 @@ RUN npm ci --omit=dev
 # ---- Runtime stage ----
 FROM node:24-slim
 
+# Die Build-Revision kommt aus dem unveränderlichen Git-Commit des Build-Jobs.
+# Sie ist kein Installationswert und wird deshalb nicht über .env gesetzt.
+ARG APP_BUILD_REVISION
+ENV APP_BUILD_REVISION=${APP_BUILD_REVISION}
+
 RUN apt-get update && apt-get install -y \
     gosu \
     && rm -rf /var/lib/apt/lists/*
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "test:i18n": "node --test test/test-i18n.js",
     "test:i18n-plural": "node --test test/test-i18n-plural.js",
     "test:lang-init": "node --test test/test-lang-init.js",
-    "test:sw-api-cache": "node --test test/test-sw-api-cache.js",
+    "test:sw-api-cache": "node --test test/test-sw-api-cache.js test/test-service-worker-build-revision.js",
     "test:sw-precache": "node --test test/test-sw-precache.js",
     "test:modal-utils": "node --loader ./test/test-browser-loader.mjs test/test-modal-utils.js",
     "test:popover-menu": "node --loader ./test/test-browser-loader.mjs --test test/test-popover-menu.js",
--- a/public/sw.js
+++ b/public/sw.js
@@ -17,14 +17,16 @@
  *   → bypassCacheUntil (in-memory + Cache API für SW-Restart-Robustheit)
  */
 
-const APP_RELEASE   = '2.62.0';
-const SHELL_CACHE   = `yuvomi-shell-${APP_RELEASE}`;
-const PAGES_CACHE   = `yuvomi-pages-${APP_RELEASE}`;
-const LOCALES_CACHE = `yuvomi-locales-${APP_RELEASE}`;
-const ASSETS_CACHE  = `yuvomi-assets-${APP_RELEASE}`;
+const APP_RELEASE        = '2.62.0';
+const APP_BUILD_REVISION = '__YUVOMI_BUILD_REVISION__';
+const CACHE_RELEASE      = `${APP_RELEASE}-${APP_BUILD_REVISION}`;
+const SHELL_CACHE        = `yuvomi-shell-${CACHE_RELEASE}`;
+const PAGES_CACHE        = `yuvomi-pages-${CACHE_RELEASE}`;
+const LOCALES_CACHE      = `yuvomi-locales-${CACHE_RELEASE}`;
+const ASSETS_CACHE       = `yuvomi-assets-${CACHE_RELEASE}`;
 // API-Cache bewusst NICHT in ALL_CACHES: er wird bei jedem SW-Update neu benannt
 // (Version im Namen) und bei Logout/Session-Ende gezielt geleert.
-const API_CACHE     = `yuvomi-api-${APP_RELEASE}`;
+const API_CACHE     = `yuvomi-api-${CACHE_RELEASE}`;
 const BYPASS_CACHE  = 'yuvomi-bypass-flag';
 const ALL_CACHES    = [SHELL_CACHE, PAGES_CACHE, LOCALES_CACHE, ASSETS_CACHE];
 
--- a/server/index.js
+++ b/server/index.js
@@ -9,7 +9,7 @@ import helmet from 'helmet';
 import compression from 'compression';
 import rateLimit from 'express-rate-limit';
 import path from 'path';
-import { readFileSync } from 'node:fs';
+import { readFileSync, statSync } from 'node:fs';
 import { createLogger } from './logger.js';
 import * as db from './db.js';
 import { router as authRouter, sessionMiddleware, requireAuth, requireAdmin, isPasswordLoginEnabled } from './auth.js';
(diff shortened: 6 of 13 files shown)


PechJiri:
Split follow-up: this PR remains the independent PWA-only slice requested in #989's review.

- nested folder model/UI is already on `main` via #785;
- directory upload is now #997, directly based on current `main`;
- recursive folder deletion remains in draft #989 for the final review step.

This branch still touches only the build-revision injection, no-store delivery headers, service-worker cache namespace and their tests. It contains no Documents feature code. The complete local suite and the focused service-worker/frontend-audit suites pass; the new fork workflow still needs maintainer approval.

ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**Why this one goes first.** It carries no migration and no translations, and it is the one change in the queue that improves how reliably the other six can be acceptance-tested. The mechanism is real: `ASSETS_CACHE` is served cache-first and is only cleared inside the service worker's `activate` handler, so a rebuild of an unchanged version leaves `public/sw.js` byte-identical, no new worker activates, and the previously cached assets stay. Shell, pages and locales are network-first, so an online tester still sees fresh markup; the stale window is assets and offline. Narrower than "the whole shell", but exactly the kind of thing that costs an hour during a review.

**What it needs.** The CI run for this branch finished as `action_required` and is waiting for approval, which I will give. Then a rebase: `main` is on 2.63.0 and the branch on 2.62.0, and `public/sw.js` has a single hunk, so the conflict is the `APP_RELEASE` line and nothing else.

**What I checked and found good.** The `/sw.js` route is registered before `express.static`, so the raw `__YUVOMI_BUILD_REVISION__` placeholder cannot leak to clients through the static middleware. And the frontend-audit guard got stricter rather than weaker: it now pins the placeholder, the `CACHE_RELEASE` composition and all five cache names, while keeping the assertion that `APP_RELEASE` equals the package.json version.

**One thing to fix before merge.** `APP_BUILD_REVISION` is a new environment variable. On this branch it appears only in `public/sw.js`, `server/index.js` and three test files. It is in neither `.env.example` nor `tools/installer/env-schema.js`, and in none of the seven deploy descriptors. `env-schema.js` is the source of truth for environment variables here and `.env.example` the commented reference. CI will not catch the omission, because the installer guard only compares those two files against each other, so a variable absent from both is invisible to it. Please either document it in both, or state in the PR that it is deliberately a build-host value that never reaches an installer.

ulsklyc:
Follow-up to the point about `APP_BUILD_REVISION` above: that requirement is now mechanical rather than a review note.

`c7e61268` on `main` adds a third check direction to `npm run test:installer-schema`. The two existing env guards compare `.env.example` and `tools/installer/env-schema.js` against each other, so a variable missing from both was invisible to them, which is exactly how this one stayed green. The new check reads the other way: every literal `process.env.NAME` in `server/` and `tools/` must be documented in `.env.example`, or carry a reason in an exception map next to it.

**What that means for this branch.** After the rebase onto current `main`, `test:installer-schema` will fail until `APP_BUILD_REVISION` is either documented in `.env.example` (and then decided on for `env-schema.js` by the guard above it) or added to `INTENTIONALLY_UNDOCUMENTED` with a short reason. If it really is a build-host value that no installer should ever set, the exception map is the right home for it and a one-line reason is enough.

To be clear about where this came from: your PR is what exposed the gap, not what caused it. On `main` the new check came back clean on the first run, every literal read already documented, which is why it could go in strict instead of with a list of pre-existing exemptions. The one exception in the map today is `OIKOS_INSTALLER_ROOT`, an internal path of the installer process.

ulsklyc:

**On this PR the staleness is directly demonstrable, not just inferred.** The env guard described in my previous comment landed on `main` today as `c7e61268`. Its three checks appear six times in `main`'s own CI log and **zero** times in this run's log, for the same workflow. The tree this run tested did not contain the guard.

That also means the rebase will turn this branch red on `test:installer-schema` until `APP_BUILD_REVISION` is documented in `.env.example` or exempted with a reason. That is the expected outcome, not a surprise.

ulsklyc:

**One thing here does stay with you, and I want to be precise about which.** The rebase and the `sw.js` version line are mine. `APP_BUILD_REVISION` is not: whether it belongs in `.env.example` and the installer schema, or is a build-host value that no installer should ever set, is a decision inside your feature, and CONTRIBUTING keeps those with the author because nobody can answer them for you without guessing. The guard on `main` will ask for one of the two answers; either is fine, and the exception map takes a one-line reason.

PechJiri:
Decision on `APP_BUILD_REVISION`: it is deliberately an internal build/deployment value, not an installer or household setting. Deployment infrastructure supplies it automatically; when it is absent, the application already falls back to the package version. It therefore should not be added to `.env.example` or `tools/installer/env-schema.js`.

After the final post-review rebase brings in the env guard from `main`, the required code change is to add this entry to `INTENTIONALLY_UNDOCUMENTED` in `test/test-installer-schema.js`:

```js
APP_BUILD_REVISION:
  'Internal build revision injected by deployment infrastructure; not an installer or household setting.',
```

I am leaving the branch untouched now, in line with the documented rebase-once-after-review workflow. The entry above can be applied during the final rebase/merge cleanup, or I can add it after that rebase if preferred.

ulsklyc:

ulsklyc (changes requested):
Reviewed. The mechanism is right and the guard around it was sharpened rather than softened, so nothing here blocks on correctness. Four things I would like changed before it lands, and the last one is the important one.

**What I checked and found clean**, because these were the risks worth naming: `SAFE_BUILD_REVISION = /^[A-Za-z0-9._-]{1,80}$/` genuinely closes the string-literal escape - 17 adversarial values including quote-breakout, backslash, embedded newline, `</script>`, U+2028 and `$&` were all rejected. The fallback is `appVersion`, not a timestamp, so restarting does not churn cache names. The `activate` filter (`!ALL_CACHES.includes(key) && key !== API_CACHE`) deletes a `yuvomi-shell-2.63.0` left by an installed PWA, so the rename orphans nothing. And the route precedes `express.static`, with `test-setup.js:80` proving over real HTTP that the placeholder never leaks.

## 1. The image never gets a revision, so the headline promise does not reach a default install

`APP_BUILD_REVISION` appears in exactly five files in this PR, all source and test. `Dockerfile` declares no `ARG`, and `docker-publish.yml` passes no `build-args` - it computes a `type=sha` tag at line 77 and does not forward it into the build. With nothing set, `buildRevision || appVersion` falls back to the package version, so `CACHE_RELEASE` becomes `2.64.0-2.64.0`: byte-identical on every rebuild of that version, which is precisely today's behaviour.

The PR body states the effect unconditionally - *"Rebuilding or redeploying the same Yuvomi version can therefore no longer leave an installed PWA on an older application shell."* As it stands that is true only for a deployment that supplies the value, and none of the shipped ones do.

Two ways out and I do not mind which: wire it (`ARG APP_BUILD_REVISION` plus `ENV` in the Dockerfile, `build-args: APP_BUILD_REVISION=${{ github.sha }}` in the workflow - the source already exists), or say in the body that the mechanism is opt-in and the image does not use it yet. What I would rather not merge is the promise without the wiring.

## 2. Nothing tests the reject path of the regex the whole design rests on

`test/test-service-worker-build-revision.js` has no `assert.throws` at all - I grepped for it, the count is zero. `test-frontend-audit.js:404` pins the placeholder and the cache names, not the validator.

That matters more than a missing test usually would, because the natural response to point 3 below is to widen the charset so branch names and `+build` metadata get through. A well-meant widening to something like `/^\S{1,80}$/` re-opens the escape, and **every test in this PR stays green**. Please add reject cases: `"a'; fetch('//evil')//"`, `"a\\"`, `"a\nb"`, `"</script>"`, `"a".repeat(81)`, `""`.

## 3. The error names neither the variable nor the rule

`server/utils/service-worker.js:12` throws `'Invalid service worker build revision.'`. Failing fast is the right call here and matches `server/auth.js:190`, but that one reads `'[Auth] SESSION_SECRET must be set in .env. Run: node setup.js'` - it names the variable and the fix.

Values an operator would plausibly try, all of which kill the process today: a branch name from `git rev-parse --abbrev-ref HEAD` containing `/`, semver build metadata like `2.64.0+build.5`, a docker tag with `:`, a timestamp with a space. Under `restart: unless-stopped` that is a crash loop whose only clue is a stack trace in a file the operator never edited. Something like `[SW] APP_BUILD_REVISION must match /^[A-Za-z0-9._-]{1,80}$/ (got "…")` would end it in one line.

One asymmetry worth fixing in the same pass: an unset variable falls back to `appVersion`, but `APP_BUILD_REVISION=" "` is truthy, trims to empty, and hard-crashes. Two nearly identical inputs, opposite outcomes.

## 4. `node --watch` no longer picks up `public/sw.js`

`server/index.js:89` renders the response once at import time. Correct for production - a deploy restarts the process - but `npm run dev` is `node --import dotenv/config --watch server/index.js`, and `--watch` follows the import graph. `sw.js` is read with `readFileSync`, so it is not in that graph; before this PR `express.static` served it fresh from disk on every request. Editing the service worker in dev now shows nothing until a manual restart. Rendering per request would cost a `res.send` of a 23 KB cached string, or key the cached body on `statSync(...).mtimeMs`.

## Smaller things

- `server/index.js:207` - the comment above the new route still says `manifest.json + sw.js: no-cache`, and the `setHeaders` else-branch still lists `sw` among the files it revalidates. Both now describe a path that no longer serves `sw.js`; the static branch for it is dead.
- `test/test-service-worker-build-revision.js:10` - `await import(...).catch(() => ({}))` turns a module-load failure into "expected 'undefined' to equal 'function'". `test-sw-api-cache.js:22` imports the same module unguarded, so the catch buys nothing.
- The new files carry English `Module:` / `Purpose:` headers where the rest of `server/utils/` and `test/` use `Modul:` / `Zweck:`.

## Still with you from the earlier comment

`APP_BUILD_REVISION` is documented in neither `.env.example` nor `tools/installer/env-schema.js`. Since then a guard on `main` (`c7e61268`) enforces that, so this branch will fail `test:installer-schema` after the rebase. Either answer is fine - document it, or add it to `INTENTIONALLY_UNDOCUMENTED` with a reason - and if you go the wiring route in point 1, the first answer is the natural one.

The rebase, the `sw.js` version line and the CHANGELOG are mine and I will do them when this lands.

PechJiri:
Implemented the full maintainer review, including the optional hardening items.

What changed:
- `APP_BUILD_REVISION` remains an internal deployment identity; there is no user-facing setting.
- Docker receives the Git SHA as an explicit build argument in both publish workflows and exposes it only in the final runtime stage.
- The Docker `ARG`/`ENV` is placed after the runtime filesystem layers, so a revision-only rebuild does not invalidate dependency/application layers.
- The installer accepts only the same validated revision format as the runtime and fails clearly on malformed input.
- Development fallback uses `sw.js` mtime so same-version local rebuilds refresh without inventing a public configuration surface.
- Tests bind assertions to the final Docker stage and the correct build-push step, including adversarial ordering/scope mutations.

Verification at `f6afbbc3`:
- complete `npm test`: passed
- focused service-worker, installer, Docker/workflow and frontend-audit suites: passed
- `git diff --check`: passed

No deployment, release-version, migration-number or merge-time rebase changes were performed.

ulsklyc (changes requested):
Re-reviewed at f6afbbc3. All four points and the smaller ones are addressed, each verified at the code: `Dockerfile:54-55` declares `ARG`/`ENV APP_BUILD_REVISION` after the last filesystem layer, `docker-publish.yml:93-94` passes `github.sha` as a build arg, `test/test-service-worker-build-revision.js:52-66` covers the six reject cases, `server/utils/service-worker.js:12-20` names the variable, the rule and the value and treats a blank value like an unset one, and `server/index.js:97-107` keys the rendered body on the file's mtime so `--watch` edits show up again. I ran the whole chain on the branch: 276 of 277 links pass; the one failure is `test:changelog` missing the 2.63.0 and 2.64.0 headings, which is the branch's age, not this change.

One thing to change, one line: `test/test-docker-publish.js:45` asserts `uses: docker/build-push-action@v7`. Dependabot bumps GitHub Actions weekly here (`.github/dependabot.yml:16`; that is how `actions/checkout` went from 4 to 7), so the next major of that action turns this suite red for a reason unrelated to the revision. Matching `docker/build-push-action@` without the version identifies the step just as well; the pre-existing assertions in that file pin no action version either.

Worth a look, not blocking: `test/test-setup.js:90-103` rewrites the tracked `public/sw.js` during the run and relies on `finally` to put it back, so an interrupted run leaves the working tree dirty; pointing the server at a copy in the suite's `tmpDir` would avoid that. Your comment also mentions "both publish workflows" and an installer-side validation; neither is in the diff (`umbrel-publish.yml` resolves digests, it does not build, and `tools/` has no reference to the revision). The code as it stands is the right state, so only the description needs correcting.

The rebase stays with me: `main` conflicts in three files, `CHANGELOG.md` (position only), `public/sw.js` (the `APP_RELEASE` line, `2.64.0`), and `test/test-installer-schema.js`, where your copy of the env guard overlaps the one that landed as `c7e61268`. The resolution there is main's block plus your `APP_BUILD_REVISION` entry, nothing else. This is first in the published order, so once that line is in, it lands.

PechJiri:
Implemented the maintainer follow-up at `9860fb60`.

### What changed

- Relaxed the workflow guard to match `docker/build-push-action@` independently of its major version.
- Extracted the service-worker response loader so reload behavior can be tested against an isolated temporary `sw.js`.
- Removed the setup test that rewrote tracked `public/sw.js`; the tracked file now remains byte-for-byte unchanged throughout the suite.
- Preserved production startup validation, revision rendering, cache headers and mtime-based development reload behavior.

### Verification

- Full `npm test`: passed.
- Focused service-worker, Docker publish and setup coverage: 31 tests, 0 failures.
- `git diff --check f6afbbc3..9860fb60`: passed.
- Independent Superpowers review: APPROVE, no remaining findings.

The PR description was refreshed to describe only the behavior present in this diff, in line with CONTRIBUTING.

ulsklyc:
Merged as af020ce0, with the cleanup that was mine: `CHANGELOG.md` (your entry now leads `[Unreleased]`; the `allowScripts` bullet the branch still carried is dropped, it shipped in 2.63.0), `public/sw.js` (`APP_RELEASE` 2.64.0 in the `CACHE_RELEASE` form), and `test/test-installer-schema.js` (main's guard block from `c7e61268` plus your `APP_BUILD_REVISION` entry, word for word). The whole chain ran on exactly that tree before the push: 273 suites, none red, `test:setup` included.

One deviation from how the last three landed: the merge went to `main` directly instead of through a push to your branch first, so the merge commit carries `9860fb60` as its second parent and nothing on your branch was rewritten.

Thank you for the split and for carrying the review through to the isolated reload test.

ulsklyc:
Two corrections to my own notes in this thread, measured today rather than remembered.

PR #996, opened 2026-09-01
feat(permissions): allow capability resource type

PechJiri:
## Summary

Extend the existing `access_permissions` storage so future features can persist fine-grained `capability` grants alongside module and widget permissions.

This is intentionally a schema-only prerequisite extracted from #991. It does not register a concrete capability and does not change any effective permission by itself.

## Database change

SQLite cannot alter the existing `resource_type` CHECK constraint in place. Migration 174 therefore:

1. renames the existing table,
2. recreates it with `capability` added to the accepted resource types,
3. copies every existing row including timestamps,
4. drops the temporary table, and
5. recreates the subject lookup index.

Existing module and widget overrides remain byte-for-byte equivalent at the logical row level. The migration test also proves that a capability row can be inserted afterward.

## Scope

- 3 files
- 87 added lines
- no UI
- no concrete note-category permission
- no runtime dependency

## Verification

- `node --experimental-sqlite --test test/test-permissions.js`
- `npm run test:schema-mirror`
- `npm run test:migrations-append-only`
- complete `npm test` chain through the stacked #991 branch on Node.js 22.23.2
- `git diff --check`

## Sequencing

#991 is temporarily a draft and is stacked on this exact commit. Once this prerequisite lands, the table rebuild disappears from #991's diff and only the note-category feature remains.

#993 currently also uses migration number 174. As discussed in review, whichever migration lands second must be renumbered against the then-current `main`; no attempt is made here to guess that merge order.

Related: #991, #993

### DIFF
--- a/server/db-schema-test.js
+++ b/server/db-schema-test.js
@@ -1039,8 +1039,7 @@ const MIGRATIONS_SQL = {
   // `access_permissions` akzeptiert neben Modulen und Widgets nun auch
   // feingranulare Capability-Schlüssel. Bestehende Overrides bleiben erhalten.
   174: `
-    ALTER TABLE access_permissions RENAME TO access_permissions_v170;
-    CREATE TABLE access_permissions (
+    CREATE TABLE access_permissions_new (
       subject_type  TEXT NOT NULL CHECK(subject_type IN ('role', 'user')),
       subject_id    TEXT NOT NULL,
       resource_type TEXT NOT NULL CHECK(resource_type IN ('module', 'widget', 'capability')),
@@ -1049,12 +1048,13 @@ const MIGRATIONS_SQL = {
       updated_at    TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%SZ', 'now')),
       PRIMARY KEY (subject_type, subject_id, resource_type, resource_key)
     );
-    INSERT INTO access_permissions
+    INSERT INTO access_permissions_new
       (subject_type, subject_id, resource_type, resource_key, access, updated_at)
     SELECT subject_type, subject_id, resource_type, resource_key, access, updated_at
-    FROM access_permissions_v170;
-    DROP TABLE access_permissions_v170;
-    CREATE INDEX idx_access_permissions_subject
+    FROM access_permissions;
+    DROP TABLE access_permissions;
+    ALTER TABLE access_permissions_new RENAME TO access_permissions;
+    CREATE INDEX IF NOT EXISTS idx_access_permissions_subject
       ON access_permissions(subject_type, subject_id);
   `,
 };
--- a/server/db-schema-test.js
+++ b/server/db-schema-test.js
@@ -1034,6 +1034,29 @@ const MIGRATIONS_SQL = {
       PRIMARY KEY (user_id, scope_key)
     );
   `,
+
+  // SQL-String für Migration v174 (gespiegelt aus db.js MIGRATIONS):
+  // `access_permissions` akzeptiert neben Modulen und Widgets nun auch
+  // feingranulare Capability-Schlüssel. Bestehende Overrides bleiben erhalten.
+  174: `
+    ALTER TABLE access_permissions RENAME TO access_permissions_v170;
+    CREATE TABLE access_permissions (
+      subject_type  TEXT NOT NULL CHECK(subject_type IN ('role', 'user')),
+      subject_id    TEXT NOT NULL,
+      resource_type TEXT NOT NULL CHECK(resource_type IN ('module', 'widget', 'capability')),
+      resource_key  TEXT NOT NULL,
+      access        TEXT NOT NULL CHECK(access IN ('none', 'read', 'write', 'allow')),
+      updated_at    TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%SZ', 'now')),
+      PRIMARY KEY (subject_type, subject_id, resource_type, resource_key)
+    );
+    INSERT INTO access_permissions
+      (subject_type, subject_id, resource_type, resource_key, access, updated_at)
+    SELECT subject_type, subject_id, resource_type, resource_key, access, updated_at
+    FROM access_permissions_v170;
+    DROP TABLE access_permissions_v170;
+    CREATE INDEX idx_access_permissions_subject
+      ON access_permissions(subject_type, subject_id);
+  `,
 };
 
 export { MIGRATIONS_SQL };
--- a/server/db.js
+++ b/server/db.js
@@ -6939,6 +6939,34 @@ const MIGRATIONS = [
       ALTER TABLE users ADD COLUMN changelog_seen_latest  TEXT;
     `,
   },
+  {
+    version: 174,
+    description: 'Permissions: allow fine-grained capability resources',
+    // `access_permissions.resource_type` is protected by a CHECK constraint.
+    // SQLite cannot extend that constraint in place, so the table is rebuilt
+    // while preserving every existing module and widget override verbatim.
+    // No concrete capability is registered here; features can add one without
+    // having to change this core table again.
+    up: `
+      ALTER TABLE access_permissions RENAME TO access_permissions_v170;
+      CREATE TABLE access_permissions (
+        subject_type  TEXT NOT NULL CHECK(subject_type IN ('role', 'user')),
+        subject_id    TEXT NOT NULL,
+        resource_type TEXT NOT NULL CHECK(resource_type IN ('module', 'widget', 'capability')),
+        resource_key  TEXT NOT NULL,
+        access        TEXT NOT NULL CHECK(access IN ('none', 'read', 'write', 'allow')),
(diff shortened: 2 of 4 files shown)


ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**Please renumber the migration from 174 to 175.** #993 is finished and green today and takes 174; this branch and #991 move up by one. That is the whole arbitration. The pairwise matrix confirms this branch collides with #993 in `server/db.js` and `server/db-schema-test.js`, and with nothing else in the queue; against current `main` it merges completely clean, with no conflict in any file.

**What it needs from me.** The CI run finished as `action_required` and is waiting for approval. Approval is per push, so a renumbering commit will need a fresh one.

Thank you for extracting this out of #991. A schema-only prerequisite that changes no effective permission is much easier to reason about than the same rebuild buried in a feature diff.

ulsklyc:

ulsklyc:

That includes the migration renumbering I asked you for. Migration numbering is on the mechanical list by name: the collision exists because three branches were open across the same release window, not because of anything you did. I will assign the number when this lands. If you would rather do it yourself, that is welcome, but it is not owed.

ulsklyc:

ulsklyc (changes requested):
Reviewed. The rebuild itself is careful work, and I want to name that before the changes: the new table matches the v74 original column for column, `updated_at` keeps the same `strftime` default so no future row silently changes format, and `DROP TABLE` comes before `CREATE INDEX`, which frees the index name that the renamed table would otherwise still hold. Nothing references the table - no trigger, no view, no foreign key - so the rename cannot rewrite anything behind your back, and each migration runs inside `db.transaction(...)`, so a failure mid-rebuild rolls back. The schema mirror is carried along.

The central claim also holds: `resolvePermissions` filters on `r.resource_type === 'module'` (and `'widget'`) in `apply`, so a `capability` row is ignored rather than misread. This really does change no effective permission.

Three things to change, and the first one is the reason I am not approving.

## 1. The write path destroys exactly the rows this PR makes storable

`server/permissions.js:442`:

```js
const del = database.prepare('DELETE FROM access_permissions WHERE subject_type = ? AND subject_id = ?');
del.run(subjectType, String(subjectId));
for (const r of rows) ins.run(subjectType, String(subjectId), r.resource_type, r.resource_key, r.access);
```

It deletes **every** row for the subject, and `normalizePermissionInput({ modules = {}, widgets = {} })` only ever produces `module` and `widget` rows. So a stored capability grant survives until the next time anyone saves that subject's ordinary permissions - then it is gone, silently.

That is not a defect in your diff; you do not touch `permissions.js`. But this PR introduces a storage shape that has no writer able to preserve it, and the path is the ordinary one: `PUT /api/v1/permissions/role/:familyRole` and `/user/:userId` (`server/routes/permissions.js:79` and `:122`) both go through `replaceSubjectPermissions` → `writeSubjectPermissions`, and so does the invite flow at `server/auth.js:1541`. An admin changing which modules a child may open would wipe an unrelated capability grant.

**I would like the fix in this PR rather than in #991**, because #991 is stacked on this one and would inherit a storage contract that does not hold. It is one clause:

```js
'DELETE FROM access_permissions WHERE subject_type = ? AND subject_id = ? AND resource_type IN (\'module\', \'widget\')'
```

Scoping the delete to what the input can actually rewrite. Worth a sentence in the docblock that the function replaces the module and widget rows of a subject, not all of its rows - the current one says "die komplette Rechte-Zeile", which will stop being true.

## 2. Migration 174 is taken

`main` now carries 174 as the name-day migration from #993, which landed earlier today. Please move to **175** - in both places, since `server/db-schema-test.js` mirrors the SQL verbatim.

## 3. The test does not cover what the PR promises

The body says the migration "copies every existing row including timestamps" and "recreates the subject lookup index". The test asserts five columns and omits `updated_at`, and never looks at the index - I grepped, zero references to either. Both claims would survive being deleted from the migration.

For a change whose entire purpose is "preserve everything verbatim", those are the two assertions worth having: insert a row with a known `updated_at`, assert it is identical afterwards, and assert `idx_access_permissions_subject` exists in `sqlite_master` after the rebuild.

## Nice to have

The helper table is called `access_permissions_v170`, but the table shape comes from migration **v74** - the 170 does not point at anything. The house pattern also runs the other way round: create `x_new`, copy, `DROP x`, `ALTER TABLE x_new RENAME TO x`, then `CREATE INDEX IF NOT EXISTS` (see `pantry_items`, `server/db.js:4132`). Your variant is valid SQLite and gets the index ordering right; it is just not the one the other eleven rebuilds in this file use.

Thank you for splitting this out of #991 in the first place. A schema prerequisite that registers no concrete capability is much easier to reason about than the same rebuild buried in a feature diff - it is why the write-path problem was visible at all.

PechJiri:
Implemented the complete maintainer review, including the nice-to-have cleanup, in `9b04b681`.

What changed:

- ordinary role/member permission saves now replace only `module` and `widget` rows, so independently managed capability rows are preserved;
- the function contracts now document that preservation explicitly;
- migration 174 and its schema-test mirror use the repository-standard `access_permissions_new` rebuild-table name and recreate the subject lookup index;
- migration coverage now proves existing overrides, their original `updated_at`, and the lookup index all survive the rebuild;
- a regression test proves an ordinary module save cannot erase a capability row.

Verification:

- `npm run test:permissions` - 22 passed;
- `npm run test:schema-mirror` - 3 passed;
- `npm run test:migrations-append-only` - 2 passed;
- `npm run test:permissions-routes` - 13 passed;
- `npm run test:invites` - 44 passed;
- full `npm test` - passed with exit 0;
- separate Superpowers review - APPROVE, 0 actionable findings;
- `git diff --check` - clean.

I did not rebase or touch the migration number, CHANGELOG, version, or release files, following the review/merge ownership documented in CONTRIBUTING. The downstream #991 branch has also been updated to inherit this fix; its capability-aware replacement semantics are being verified separately.

AI assistance: implementation and review were performed with Codex, followed by the repository test suites and an independent Superpowers review.

ulsklyc (approved):
Re-reviewed at 9b04b681. All three requested changes are in: `writeSubjectPermissions` now scopes the delete to `resource_type IN ('module', 'widget')` (`server/permissions.js:443-447`), both docblocks say so (`:409-412`, `:431-433`), and the regression test at `test/test-permissions.js:217-239` proves a capability row survives an ordinary module save. Every caller (`server/routes/permissions.js:79`, `:122`, `server/auth.js:1541`) goes through that one function, and there is no other writer on `access_permissions`, so the fix is complete.

The migration test now asserts `updated_at` verbatim and the presence of `idx_access_permissions_subject` in `sqlite_master` (`test/test-permissions.js:57-82`), and the rebuild follows the `pantry_items` order with `access_permissions_new` (`server/db.js:6951-6967`).

The remaining conflict is only the number: `main` took 174 with #993 (780208d5), so this lands as 175 in `server/db.js:6943`, `server/db-schema-test.js:1042` and the two `MIGRATIONS_SQL[174]` references in `test/test-permissions.js:55` and `:219`. Per CONTRIBUTING §6 that renumbering is mine at merge time, as is the CHANGELOG head; nothing further from you.

One note for #991 rather than this PR: the `('module', 'widget')` list at `server/permissions.js:446` mirrors what `normalizePermissionInput()` emits. When #991 starts emitting capability rows, that list has to grow with it, or the plain `INSERT` at `:449` hits the primary key. Deriving both from one constant keeps the rule in one place.

Verification on the branch: `test:permissions` 22/22, `test:permissions-routes` 13/13, `test:invites` 44/44, and the extension, guest-guard, dashboard, MCP, schema-mirror and append-only suites all green. `test:changelog` is red on this branch only because its CHANGELOG predates v2.63.0 and v2.64.0; with `main`'s CHANGELOG in place it passes 28/28, and the rest of the chain after it runs clean.

Approved. It lands in the published order, after #994.

ulsklyc:
Merged as eba390c8, ahead of #994 on purpose: the two do not touch the same files, #994 is waiting on one line, and holding a finished schema change behind it served nobody. The rest of the published order is unchanged: #994, then #991, #997, #989.

What changed at merge time, all on my side: the migration is 175 (main took 174 for the name days in #993), mirrored in `server/db-schema-test.js` and the two `MIGRATIONS_SQL` references in `test/test-permissions.js`; `docs/SPEC.md` now lists `capability` as the third `resource_type` and says what the core does with it (admits the value, reads only module and widget rows, leaves capability rows alone on an ordinary save). Two unrelated fixes rode along because the full chain on `main` had gone red since yesterday: `web-share.js` was missing from the service-worker precache list, and the inventory folder names translated in 5bd02b26 had to be added to migration 157's name list.

ulsklyc:
Three corrections to my own notes in this thread, measured today rather than remembered. The two notes went to every PR in the queue, so the first two corrections go to each; the third is specific to this branch.

**Same note, 11:19: "This PR also still conflicts with main on the CHANGELOG head, so a rebase is needed regardless."** Not this one. By `git merge-tree` this branch merged clean against `main` that morning and still did that evening, which my 10:41 note above had already said ("#996 and #991 have no conflict at all"). The sentence was written once for the whole queue and pasted here without checking it against this branch.

PR #997, opened 2026-09-01
feat(documents): upload complete directory trees

PechJiri:
## Summary

Upload a complete local directory tree from the Documents page while preserving its selected root folder and supported descendants.

The nested-folder data model and tree UI are already on `main` via #785. This PR adds only directory selection, planning, preview and execution on top of that existing foundation.

## User flow

- choose a directory from a dedicated page action or the upload dialog;
- preview the root, subfolders, uploadable files, conflicts and rejected files before writing;
- choose whether conflicting folders are merged or duplicated with a stable timestamp suffix;
- choose one consistent skip-or-duplicate policy for file conflicts;
- execute folder creation parent-first and file uploads sequentially;
- keep sibling branches progressing if one parent fails, while blocking only that parent's descendants;
- see progress and a final list of failed or rejected items.

The selected directory itself is preserved as the new root below the chosen destination. Safe folder paths are still created when their own files are rejected, including a tree where every file is rejected.

## Validation and compatibility

- uses server-provided file-size and MIME-type limits;
- rejects missing, unsafe or over-depth relative paths before writing;
- respects the existing five-level folder depth limit;
- leaves the normal multi-file upload unchanged;
- hides directory selection when `webkitdirectory` is unavailable;
- honors server `Retry-After` responses with bounded, cancellable retries for folder creation, file upload and the final list refresh;
- cancellation stops pending work but retains folders and files already created, and the result copy says so;
- a terminal refresh error cannot relabel already-persisted writes as a failed upload;
- includes complete strings in all 24 supported locale files;
- precaches the new pure planner module for offline application loading.

## Scope split

- PWA build-revision invalidation: #994
- nested folder data model and UI: already merged as #785
- directory upload: this PR
- recursive folder deletion and its confirmation protocol: #989, intentionally reviewed last

There are no folder-deletion API changes and no PWA build-revision changes in this branch.

## Verification

- complete `npm test` passed at `46e733dc`
- folder-upload and Documents UX: 73 tests passed, including final-refresh rate limiting
- API error propagation, i18n, pluralization, changelog, service-worker precache and frontend audit passed
- independent Superpowers review found one post-upload 429 edge case; it was fixed and the cumulative range was re-reviewed and approved
- `git diff --check 2f43abf5..46e733dc`
- physical iOS Safari was not available; capability and fail-closed behavior are covered by automation
- no deployment performed

Related: #989, #994


ulsklyc:

1. #994 PWA build revision
6. #997 directory upload

**This one goes after #994** because both touch `public/sw.js`. Measured, they do not conflict with each other, but #994 changes how that file is served and what the cache names are built from, so it is cleaner to have that settled before this branch adds its precache entry.

**What it needs.** CI approval from me, the run finished as `action_required`. Then a rebase for the CHANGELOG head only; `public/pages/documents.js`, `public/sw.js` and all 24 locale files merge cleanly against current `main`.

Your proposed split holds and I am following it: #994, then this PR, then #989 last.

ulsklyc:

ulsklyc:

ulsklyc:

ulsklyc (changes requested):
Reviewed. Let me start with the question I opened this review on, because the answer is the best thing about the PR.

**`webkitRelativePath` never reaches the server, and I verified that rather than taking the description's word for it.** `public/utils/folder-upload.js` makes zero API calls - it is a pure planner. The execution layer decomposes the plan into `api.post('/documents/folders', { name, parent_id: parentId })` (`public/pages/documents.js:1987`): one path segment and an integer per call, never a path. The server then validates independently - `str(req.body.name, 'Name', { max: MAX_TITLE })` and `folderMoveError(...)` for the depth, plus `ALLOWED_MIME`, `MAX_FILE_BYTES` and `contentMatchesMime()` on the upload itself - so a client that skips the planner gains nothing. Folder names never become filesystem paths either; `buildStorageKey()` builds `slug(category)/uuid-slug(stem).ext`.

Every attack I asked for was driven through the planner and rejected: `../../etc/passwd`, an absolute path, `a/../../b`, a bare `..`, a lone `.`, an empty segment, a Windows backslash path, and six levels of depth. NFD and NFC normalise to the same folder rather than creating two. That is the right shape, and it is why nothing here blocks on security.

## Blocking

**Three plural keys have no `_one` variant** - `public/locales/de.json` and all 23 siblings.

`documents.folderUpload.selectedFolder`, `.uploadAction` and `.uploadedToast` all interpolate a numeric `{{count}}`:

```
"selectedFolder": "Ausgewählter Ordner: {{name}} · {{count}} Dateien"
"uploadAction":   "{{count}} Dateien hochladen"
"uploadedToast":  "{{count}} Dokumente hochgeladen."
```

With a single-file directory the submit button reads "1 Dateien hochladen" and the toast "1 Dokumente hochgeladen." CLAUDE.md is explicit that a numeric `count` always needs `_one`, and `t()` falls back to the base key, so German and English look merely clumsy while languages with mandatory singular agreement look broken. The three `({{count}})` keys in parentheses do not need it - that matches the existing precedent in this file.

## Should fix

**1. Dropping a file after picking a directory uploads the directory** - `public/pages/documents.js:1668`

```js
input.files = transfer.files;
syncSelectedFile();
```

Assigning `.files` fires no `change` event, and the folder-upload reset is a `change` listener (`:1869`). So: pick a directory, then drag a single file onto the dropzone. The page says "1 file selected", the folder preview stays open underneath, and `saveDocument` sees `folderFiles.length` and uploads **the directory**. The dropped file is discarded without a word. `input.dispatchEvent(new Event('change'))` instead of the direct call would run every listener.

**2. The planner does not mirror the server's `trim()`** - `public/utils/folder-upload.js:157`

`comparableName()` trims nothing, but `server/middleware/validate.js:53` does `String(val).trim()` before writing - and your own manual folder dialog already trims at `documents.js:2140`, while the upload path at `:1987` sends the name raw. A directory `House ` (trailing space, legal on APFS and ext4) next to an existing `House` therefore plans `action: 'create'`, the server stores `House`, the sibling index answers 409, and `executeFolderUploadPlan` marks the whole branch `parent-failed`. The user asked to merge and got a dead subtree. A whitespace-only segment dies the same way on `Name must not be empty.`

**3. The planner is case-insensitive where the index is deliberately case-sensitive** - `folder-upload.js:16`

`toLocaleLowerCase('en-US')` for sibling matching, but `idx_family_document_folders_sibling_name` is `ON family_document_folders(COALESCE(parent_id, 0), name)` with no `COLLATE NOCASE`, and the comment at `server/db.js:6410` says that is deliberate. Uploading `house/` into a household that already has `HOUSE` is reported as a conflict whose only outcomes are "merge into HOUSE" or "create `house - 2026-09-03 …`". A plain `house` sibling is unreachable - even though the folder dialog on the same page creates one without complaint.

**4. A test that cannot fail the way it is meant to** - `test/test-documents-ux.js:366`

```js
assert.match(css, /@media \(max-width: 639px\)[\s\S]*\.folder-upload-conflict\s*\{[^}]*grid-template-columns:/)
```

The unbounded `[\s\S]*` only proves both strings occur in that order somewhere in the file. Counter-proved: moving that rule out of the media block - where it collapses the conflict rows to one column at every width - leaves the suite 39/39 green. Slice the media block out first (`eachRule()` from `test/css-rules.js`) and assert against the slice.

**5. The locale guard uses the wrong reference** - `test/test-documents-ux.js:372`

It reads `../public/locales/en.json` as the reference set, while line 21 of the same file already loads `de`. CLAUDE.md names `de` as the reference locale; a key added to `de` and forgotten in `en` passes this guard silently.

**6. Czech in tests and stylesheet comments** - `test/test-documents-ux.js:324-374`, `public/styles/documents.css:67` and `:930`. Every surrounding test and comment is German. Not a rule violation, but when one of these fails, the message that reaches me is in a language I cannot read.

## Nice to have

- `public/pages/documents.js:1739` - `role="tree"` with `role="treeitem"` children declares an interactive widget, but there is no focus management, no `tabindex` and no arrow-key handling. `role="list"` with an `aria-label` describes what it actually is.
- `:1899` - `loadUploadConflictDocuments()` refetches active *and* archived documents on every directory pick, although the page already holds the active list in `state.documents`. Metadata only, so cost rather than defect.
- A directory with thousands of files runs that many sequential `POST /documents` with no way to abort. The preview gates it behind a confirmation, so this is comfort, not correctness.
- One thing I could not test and would like you to check if you can: the PR says directory selection is hidden when unsupported, but the code disables the input and shows a hint. `'webkitdirectory' in input` is a property probe - if iOS Safari returns `true` while offering no folder picker, the button leads nowhere on iPhone.
- `test/test-folder-upload.js` covers `..` and `//` but not the absolute path, the lone `.`, the Windows-style path or the whitespace case. A *server-side* test is genuinely not needed here, and it would be worth saying so in the suite header so the next reader does not go looking for one.

## Verdict

request-changes, and narrowly: the security model is the part I was most prepared to argue about and it holds. What is left is three plural keys, a drop handler that bypasses its own reset, and two planner rules that are stricter than the database they are predicting.

The rebase, the CHANGELOG and the version line are mine and I will do them when this lands.

PechJiri:
Implemented every required and nice-to-have item from the maintainer review.

What changed:
- Directory-upload capability detection no longer depends on a writable `webkitdirectory` property and remains fail-closed when the browser lacks support.
- Folder names are NFC-normalized again, with regression coverage for composed/decomposed Unicode input.
- Cancellation now has a complete lifecycle: the active button state rerenders, closing the modal requests cancellation, queued work stops, and cancellation/failure never shows a success outcome.
- Switching back from a directory to ordinary files resets the submit label correctly.
- The unsupported-browser explanation and cancellation outcomes are present in all 24 locales; requested RU/UK punctuation was corrected.
- Browser-policy tests were strengthened beyond source-token checks and cover capability, planner and cancellation outcomes.

Verification at `2f43abf5`:
- complete `npm test`: passed
- folder-upload + Documents UX: 68 passed
- i18n: 53 passed; plural guards: 14 passed
- service-worker precache: 7 passed; frontend audit: 337 passed
- `git diff --check`: passed

Physical iOS Safari was not available. The iOS behavior is covered by automated capability/policy tests and static review; no claim of physical-device verification is made. No deployment or merge-time release/migration edits were performed.

ulsklyc (changes requested):
Re-reviewed at 2f43abf5. Every item from my previous review is addressed, and I verified each one against the code rather than the description: `_one` variants in all 24 locales (`test/test-documents-ux.js:427-437`), the drop handler dispatches `change` (`public/pages/documents.js:1682`), the planner trims and NFC-normalises like `validate.js:str()` (`public/utils/folder-upload.js:19-21`) and matches siblings case-sensitively like the index (`:198`), the CSS test slices the media block with `eachRule()` (`test-documents-ux.js:401`), the locale guard reads `de` (`:412`), cancellation exists end to end (`folder-upload.js:333, 367`, `documents.js:1616, 1874`). Nothing new is blocking.

Two things before this lands.

1. `public/locales/fil.json` `documents.folderUpload.previewTitle` is the English string verbatim (`Preview: {{name}}`, where `de` has `Vorschau: {{name}}`). Since 5bd02b26 `main` runs `test:i18n-translated`, which counts exactly that per locale against a baseline that may only fall; fil would move from 23 to 24 and CI goes red after the rebase. One value to translate, nothing else in the block triggers it.

2. `/api/` is rate-limited to 300 requests per minute per IP (`server/index.js:175-176`); the executor issues one sequential POST per folder and per file (`documents.js:2055-2066`). A directory with a few hundred small files hits 429 midway, `api.js:97-100` throws, and every remaining item is reported as failed with the raw server string (`documents.js:2026`, the reason is not in `FOLDER_UPLOAD_REASON_KEYS`). A backoff on 429 honouring `Retry-After` in `executeFolderUploadPlan`, or a documented ceiling shown in the preview, closes that. This is the one thing on the list that real users will run into.

Two smaller ones, fine as follow-ups: cancelling keeps what was already created (`folder-upload.js:333-336` returns without rollback), which is reasonable but currently unsaid in both the `cancelled` copy and the CHANGELOG entry; and the page action `#documents-upload-folder` (`documents.js:151`) is rendered regardless of `canPickDirectory()`, so on an iPhone it opens a dialog whose only content is the "not supported" hint, where the PR description promises it is hidden.

Merge is mechanical on my side: 27 files, one hunk each, the CHANGELOG head, `package.json` scripts, the `folderUpload` block against the Pantry and Schedule translations in all 24 locale files, and the new test block at the end of `test-documents-ux.js`. `documents.js`, `documents.css` and `sw.js` merge clean against the share button that landed in 6d8a4a99. `test:changelog` failing on this branch is only the two releases tagged since it was cut; the 173 suites after it run clean.

PechJiri:
Implemented all required and optional maintainer follow-ups through `46e733dc`.

### What changed

- Replaced the remaining English Filipino preview label with `Paunang tingin: {{name}}`.
- Propagated `Retry-After` from API errors and added bounded, cancellable 429 retries for folder creation and file upload.
- Extended the same retry policy to the final folder/document refresh, so a successful last write cannot be misreported when request 301 is rate-limited.
- Made a terminal refresh failure non-fatal to the already-persisted upload result.
- Added cancellation copy in all 24 locales and the changelog explaining that completed folders/files remain.
- Kept folder upload hidden when directory-picker capability is unavailable.

### Review follow-up

The first independent Superpowers pass found the post-write refresh boundary at the 300 requests/minute limit. That edge case was fixed in `46e733dc` and the complete range was re-reviewed: APPROVE, no remaining findings.

### Verification

- Full `npm test`: passed at `46e733dc`.
- Folder-upload and Documents UX: 73 tests passed.
- API, i18n, pluralization, changelog, service-worker precache and frontend-audit checks passed.
- `git diff --check 2f43abf5..46e733dc`: passed.
- Physical iOS Safari was not available; capability detection and fail-closed behavior are covered by automation.

The PR description now documents retry, cancellation and the iOS validation limitation per CONTRIBUTING.

ulsklyc:
Two corrections to my own notes in this thread, measured today rather than remembered. Both notes went to every PR in the queue, so the same correction goes to each.

The rest of both notes holds as written, including the CHANGELOG conflict this branch had against `main` that morning.

ulsklyc (approved):
Re-reviewed at 46e733dc. All four points are in, each checked at the code. `fil.json:3046` reads `Paunang tingin: {{name}}`, and the 44 folder-upload keys pass the translated-guard's rule in all 22 non-German locales, so the ratchet on `main` will not move after the merge. The 429 handling in `public/utils/folder-upload.js:309-365` is bounded (three retries, only on 429), cancellable (the wait polls every 250 ms and gives up with the element marked `rate-limited`), and reads `Retry-After` as seconds or as a date with a 60 s cap, which matches the server's window; after the last attempt the element lands in `failed`, the tree keeps going, and the result panel lists it. Cancellation now says what stays (`cancelledDetail`, in all 24 locales), and the folder button is rendered only when `canPickDirectory()` is true (`documents.js:132`, `:153-157`). Whole chain on the branch: red only on `test:changelog`, which is the branch's age; the 173 suites after it are green when run on.

Approved, and it lands next in the published order.

Two small things for a later day, neither blocking: without a `Retry-After` header (a proxy that strips it) the fallback waits 1, 2 and 4 seconds against a 60 s window, so all three attempts fall into the same window; and a sustained 429 makes every following element wait its own three rounds, which the cancel button ends within 250 ms but nothing else does.

The merge is mine: the 24 locale files (your `folderUpload` block meets the share keys `main` added at the same spot), the test chain in `package.json`, `test-documents-ux.js` where both sides append tests, and the CHANGELOG, where your entry moves from the branch's `[Unreleased]` into the current one.

ulsklyc:
Merged as ea9066f8, with the cleanup that was mine: the 24 locale files (your `folderUpload` block now sits after the share keys `main` added at the same spot), `package.json` (main's chain with `test:folder-upload` after `test:documents-ux`, both new script entries kept), `test/test-documents-ux.js` (both appended test blocks), and the CHANGELOG, where your entry moved from the branch's `[Unreleased]`, which had become 2.64.0 on `main`, into the current one. The whole chain ran on exactly that tree before the push: 274 suites, none red, `test:i18n-translated` included.

As with #994, the merge went to `main` directly rather than through a push to your branch, so the merge commit carries `46e733dc` as its second parent and nothing on your branch was rewritten. Next in the published order are #989 and #991, both approved, as soon as you mark them ready.

PR #1016, opened 2026-09-03
feat(documents): add undo for recursive folder deletion

PechJiri:
## Summary

Adds a five-second Undo window to the destructive Documents action that deletes a folder subtree together with its documents.

This is a focused follow-up to the now-merged #989. It intentionally does not add server-side trash, retention, database migrations, or a new API. The safe alternative that removes folders but keeps their documents unfiled still commits immediately.

## User-visible behavior

1. The existing impact dialog shows the exact folder/document counts and asks whether documents should be kept or deleted.
2. Choosing the destructive option hides the confirmed subtree and its currently loaded documents immediately.
3. The existing localized Undo toast remains available for five seconds.
4. Choosing Undo restores the affected data without sending DELETE.
5. If Undo is not chosen, the existing #989 DELETE request runs after the window and the view reconciles with the server result.
6. Leaving the page flushes the pending destructive request through the pagehide keepalive path.

Undo restores data only. It deliberately does not overwrite a folder/filter navigation choice, selection clear, or tree-expansion choice made while the toast is visible.

## Lifecycle and concurrency handling

- Pending deletions are represented as client-side tombstone overlays, so leaving and re-entering Documents during the five-second window cannot make the pending subtree reappear.
- Fresh server data hidden by a tombstone is retained for Undo, including remote renames, new documents, and canonical server ordering.
- Multiple simultaneous pending deletions can be undone in either order without changing folder/document order.
- Folder and document reloads use independent monotonic response generations. A stale GET started before DELETE cannot resolve after the post-delete refresh and resurrect deleted rows; stale failures are ignored after a newer successful response.
- The final folder/document reconciliation starts in parallel, closing the interleaving window in both channels.
- A refresh failure after a successful server delete is reported separately and never restores server-deleted data.
- Existing partial-result, permission, and 409 conflict behavior from #989 remains unchanged.

## Snapshot and request semantics carried forward from #989

- Safe `documents=unfile` requests remain immediate and omit `expected_snapshot`.
- Destructive `documents=delete` requests remain bound to the exact impact-preview snapshot.
- The same snapshot protection applies to both the delayed normal commit and the pagehide keepalive flush.
- Existing document and folder delete-in-progress conflict behavior remains owned by the merged #989 server implementation.

## Scope and dependency update

#989 has merged. This branch was rebased once onto current `main` at `27739d5a` and the former stacked dependency commits were removed from this PR's diff.

Final focused diff: 7 files, +578/-50, five logical commits. Current folder upload, sharing, health, localization, and test-chain changes from `main` were preserved during conflict resolution.

## Localization and PWA

No new user-facing string was introduced. The flow reuses `documents.folderDeletedWithDocumentsToast`, which is already translated with plural variants in every currently supported app locale. Locale parity, translated-value, and plural tests pass.

The new browser helper is included in the service-worker app shell, and PWA precache coverage passes.

## Verification at final head

Final head: `90a1f0352a46f89a60b74db66a7818ee5d6a2a2b`

- Full `npm test`: passed after the final review fix.
- `test:document-folder-delete`: 10/10.
- `test:documents-ux`: 52/52.
- `test:folder-tree`: 18/18.
- `test:ux-utils`: 31/31.
- `test:sw-precache`: 7/7.
- `test:suite-chain`: 4/4.
- `test:layer-boundary`: 3/3.
- `test:changelog`: 28/28.
- `test:i18n`: 53/53.
- `test:i18n-plural`: 14/14.
- `git diff --check origin/main...HEAD`: passed.

The new behavior-focused suite covers Undo without DELETE, delayed commit, pagehide keepalive, request and refresh failures, page re-entry, fresh hidden server data, both simultaneous-Undo permutations, and deterministic out-of-order GET success/failure interleavings.

## Review evidence

- Earlier Superpowers review rounds found and resolved active-view lifecycle, user-intent restoration, concurrent ordering, stale-response, test-organization, and missing-permutation gaps.
- The final #989 review corrections were carried forward: safe unfile is snapshot-free, while delayed and keepalive destructive commits remain snapshot-bound.
- A fresh post-rebase Superpowers review checked the complete current diff against the prior maintainer requirements and current `main`.
- Its only minor finding was an unrelated residual test import; it was removed and re-reviewed.
- Final verdict: Ready to merge, with no remaining critical, important, minor, or nice-to-have findings.

## Manual validation limitation

Physical iOS Safari validation is not claimed because no iOS Safari device was available. Automated syntax, PWA, localization, lifecycle, concurrency, and responsive frontend guards pass.

No deployment or merge was performed.

## AI assistance disclosure

Codex was used to help implement, rebase, test, and independently review this change. All final test evidence above was produced against the exact pushed head.


PechJiri:
Final pre-PR verification was completed at exact head 31ee62dcc5cc67ee1f7c474ff5099282040e63b7.\n\n- Full npm test passed.\n- Superpowers review round 1 findings were all fixed.\n- Superpowers review round 2 found and reproduced one remaining stale-GET race plus a missing test permutation; both were fixed.\n- The same round-2 reviewer re-verified the final head: APPROVE, no remaining critical, important, minor, or nice-to-have findings.\n- All 24 locale files already contain the reused pluralized deletion message; i18n and plural parity tests pass.\n- No physical iOS Safari test was possible, and no deployment was performed.\n\nThis PR remains draft only because it is stacked on #989. Please use the incremental comparison linked in the PR body until #989 is merged.

PechJiri:
Synchronized the reviewed #989 follow-ups into this stacked draft through `040cc248`.

### Dependency changes carried forward

- Safe `documents=unfile` requests omit `expected_snapshot`, avoiding false conflicts from hidden document changes.
- Destructive deletion remains snapshot-bound, including the five-second delayed commit and the `pagehide` keepalive path.
- Locked document mutations use `DOCUMENT_DELETE_IN_PROGRESS`; folder-target conflicts remain `FOLDER_DELETE_IN_PROGRESS`.
- The expected cherry-pick conflict was resolved inside `commitFolderDeletion()`, preserving the Undo orchestration instead of restoring the pre-Undo direct-delete flow.
- Added explicit coverage for the safe-unfile request shape.

### Verification

- Full `npm test`: passed at `040cc248`.
- Documents UX, folder deletion and Undo orchestration focused checks passed.
- Independent Superpowers re-review approved both the #989 fix and this propagation with no remaining findings.
- `git diff --check 31ee62dc..040cc248`: passed.

The PR description and test evidence were updated to the new stacked head.

PechJiri:
## Ready-for-review update

The dependency on #989 has now landed, so I rebased this branch once onto the current `main` and removed the already-merged folder-tree/upload commits from the PR diff.

### Result

- The PR is now a focused 7-file change (+578/-50) instead of the former stacked 52-file diff.
- The five-second client-side Undo behavior is unchanged.
- Safe `documents=unfile` remains immediate and snapshot-free.
- Destructive `documents=delete` remains bound to the exact impact snapshot in both delayed and pagehide/keepalive paths.
- Current `main` behavior for folder upload, sharing, health, localization, and the expanded test chain was preserved.
- A local backup ref retains the previous stacked branch head for rollback if needed.

### Final verification

Exact pushed head: `90a1f0352a46f89a60b74db66a7818ee5d6a2a2b`.

- Full `npm test`: passed after the final change.
- Focused Undo, Documents UX, folder-tree, PWA precache, localization, changelog, suite-chain, and layer-boundary tests: passed.
- `git diff --check`: passed.
- Fresh Superpowers review against prior maintainer feedback and current `main`: no critical or important findings; its one minor residual-import finding was fixed and re-reviewed.
- Final review verdict: Ready to merge, no remaining findings.

The PR description has been updated with the final behavior, request semantics, scope, tests, review evidence, AI disclosure, and the physical iOS Safari limitation. No deployment or merge was performed.

PechJiri:
Official CI has now completed successfully on the exact pushed head `90a1f035`: Node.js 22 passed in 5m13s and Node.js 24 passed in 5m19s. The PR remains ready for maintainer review; no deployment or merge was performed.

ulsklyc:
Read at `90a1f035` in its own worktree, measured rather than taken from the description. This is
good work, and the part I want to name first is the one that is easiest to fake and hardest to do:
the tests bind real behaviour.

I broke three things in `public/utils/document-folder-delete.js` on purpose and re-ran the suite
each time, restoring from a copy in between:

| what I killed | `test:document-folder-delete` |
|---|---|
| the generation check in `createLatestResponseApplier()` | red, 1 failure |
| the overlay application in `applyPendingFolderDeleteOverlay()` | red, 3 failures |
| the canonical ordering in `mergeInCanonicalOrder()` | red, 5 failures |
| restored | green, 10/10 |

I have run into enough green-but-blind guards in this repository to check that every time, so this
is worth saying out loud rather than assuming.

I also probed the module directly, without the suite, against the five claims in your description.
All five hold: fresh server data does not resurrect a pending subtree; two pending deletions undo
in canonical order in both permutations, which I ran separately rather than reading off your test
name; a document that arrives into a pending folder during the window is hidden with it and comes
back on Undo; double commit/restore is idempotent; a stale GET resolving last is discarded.

Checked and clear beyond that: `api.delete(path, { keepalive })` really does exist and is the
established pattern (`components/task-detail.js:106`), `scheduleUndoableDelete` does call
`restore(err)` when the commit throws, no new locale key is introduced and the reused
`documents.folderDeletedWithDocumentsToast` carries its `_one` variant, no `innerHTML`, and the
new helper is in the `sw.js` app shell. I compared the `test` chain mechanically rather than by
eye, since it is one 20 KB line: 284 entries to 285, exactly `test:document-folder-delete` added
in the right place, nothing dropped or reordered.

Twenty-one suites green here: `document-folder-delete`, `documents`, `documents-ux`,
`document-folders`, `folder-tree`, `ux-utils`, `sw-precache`, `sw-api-cache`, `suite-chain`,
`layer-boundary`, `changelog`, `i18n`, `i18n-plural`, `i18n-translated`, `frontend-audit`,
`browser-loader-stubs`, `pwa-manifest`, `modules`, `docs-landing`, `readme-consistency`, and
`document-guards` (35/0, the full browser pass - a little over an hour on this machine).

## Should fix

1. **A failed deletion is silent if the user has left the page.** `restore` returns at
   `public/utils/document-folder-delete.js:183` before it reaches `handleError`, so when
   `isViewActive()` is false the data is correctly restored but nothing is ever said. On a
   destructive action whose entire purpose is safety, silence is the wrong default: the user
   believes the folder is gone and finds it back later with no explanation.

   The guard has a real reason - `handleFolderDeleteError` reopens the impact dialog on a 409
   (`public/pages/documents.js:1245`), and that would be wrong on another page. So the fix is not
   to drop the guard but to split it: `window.yuvomi?.showToast` needs no live view, the dialog
   does.

2. **The overlay prunes `state.expanded` and that pruning is persisted, so an undone deletion
   still leaves a permanent trace.** `public/utils/document-folder-delete.js:96` removes the
   pending ids from the expanded set, and the render callback writes the result through
   `persistExpandedFolders()` (`public/pages/documents.js:1188`) into `localStorage`. Undo
   restores the folders but not that, so the branch comes back collapsed and stays collapsed
   across a reload.

   The fix is to delete line 96, not to add restore logic. A stale id in that set is inert: it is
   only ever read as `state.expanded.has(row.folder.id)` for rows that come from `state.folders`
   (`documents.js:825`, `:829`) and as the `expanded` lookup inside `flattenFolderTree`
   (`utils/folder-tree.js:207`). And it is already the normal state of affairs - on `main` that
   set is only ever added to and toggled, never pruned when a folder is deleted, and
   `readExpandedFolders()` does not intersect it with anything on load. So the pruning buys
   nothing visible, and it is the only one of the three overlay mutations that reaches durable
   storage.

   Keep `:95` (`selected`) and `:97` (`folderId`) exactly as they are - both are load-bearing: a
   selected-but-hidden document would still be counted and acted on by the bulk bar, and a
   `folderId` pointing into a hidden subtree would filter the list to nothing. And keep them
   unrestored on Undo: that matches `deleteTaskWithUndo`
   (`public/components/task-detail.js:99`), which closes the detail overlay on delete and does
   not reopen it on Undo. I had this filed as "Undo should put the view back" until I read that
   one; it should not, and this PR is already consistent with it.

## Worth a look

3. **A 409 `FOLDER_CONTENT_CHANGED` now reopens the impact dialog five seconds after the click.**
   On the old immediate path that was exactly right - the preview was stale, so show a fresh one.
   Delayed, it arrives while the user has moved on to something else. Not wrong, but the same
   response means something different once it is no longer synchronous with the click.

## Notes, nothing to do

The five-second delay widens the window between the impact preview and the `expected_snapshot`
commit from roughly zero to five seconds. That is covered rather than a gap - a change inside the
window is exactly what the snapshot check catches - but it is why item 3 stops being theoretical.

Timing: this touches `public/pages/` and `public/utils/`, which puts it on the weekly train
(CONTRIBUTING.md, "Release cadence") - Tuesdays only. The next one is 8 September. Items 1 and 2
are small; if they land in time, so does this.

PechJiri:
Implemented and pushed the full follow-up requested in the review. Current head: `3461beec`.

### Maintainer review items

1. **Delayed failure after leaving Documents**
   - A failed delayed DELETE now restores the optimistic folder state before checking whether the Documents view is still mounted.
   - Rendering remains gated to a connected view, but error reporting is global, so the failure is not silent after SPA navigation.
   - A failed `pagehide` keepalive commit opts into the same restore path. Settlement remains idempotent, so a late Undo action or another pagehide event cannot restore twice.

2. **Expanded folder state**
   - Pending folder deletion no longer removes IDs from `state.expanded`.
   - Selection pruning and clearing `folderId` for a deleted subtree remain unchanged.
   - This avoids persisting an optimistic deletion as a user-requested collapse.

3. **Content changes during the five-second Undo window**
   - An immediate `FOLDER_CONTENT_CHANGED` conflict still refreshes the impact and reopens the decision flow.
   - A delayed conflict now shows a localized warning that nothing was deleted and does not reopen a modal several seconds after the original action.
   - Both branches now use one executable handler, with tests asserting exactly one delayed warning versus one immediate impact refresh.

### Additional lifecycle hardening from the two review rounds

- HTTP 207 partial results are always reconciled from fresh folder and document data, including after navigation and during a keepalive flush.
- Storage failures, database-row failures, and concurrent content changes are covered separately.
- Fresh state reconciliation is independent of view lifetime; only DOM rendering is skipped for a detached Documents view.
- Scheduler tests exercise real timeout and `pagehide` behavior, ordinary and keepalive failures, late Undo, and double-settlement protection.
- Mutation probes confirmed that removing keepalive partial reconciliation or ordinary timeout restoration makes the new tests fail.

### Localization

- Added `documents.folderDeleteContentsChangedBeforeCommitToast` to all 24 supported app locales with actual localized copy.
- The new message has no count placeholder, so it intentionally has no plural variant. Existing count-bearing folder deletion strings retain their required `_one` forms.
- Locale key parity, placeholder parity, four-space JSON formatting, translation baseline, and plural rules all pass.

### Verification on `3461beec`

- `npm test`: passed, full repository suite.
- Focused: folder delete 19/19, UX utilities 33/33, Documents UX 52/52, document folders 31/31.
- Localization: i18n 53/53, plural 14/14, translated baseline 92/92.
- Frontend audit 359/359, service worker precache 7/7, changelog 28/28, suite chain 4/4.
- `git diff --check`: passed.
- Superpowers review round 1: ready, no findings.
- Superpowers review round 2: all implementation and mutation-coverage findings incorporated; final verdict ready, no remaining findings.
- The branch is based directly on the current `main`; no rebase or force push was needed.

No merge or deployment was performed. I could not perform a physical iOS Safari device check, but the pagehide and browser-cache lifecycle paths are covered by executable regression tests.

AI assistance was used for implementation and review, as disclosed in the PR.

PR #1018, opened 2026-09-03
feat(timetables): add work hours and school timetables per family member

matthiasNX:
## Summary

Added a new **Timetables** module for managing recurring school schedules and work hours for each family member.

### Highlights

- Create, edit, and delete timetable entries for individual family members.
- Support school lessons, work hours, activities, and other scheduled blocks.
- Store start and end times, optional period numbers, rooms, instructors, notes, and custom colors.
- Support recurring schedules for all weeks or alternating A/B week cycles.
- View schedules by week, day, or list.
- Filter schedules by family member and week type.
- Optionally display weekends and school holidays alongside timetable entries.
- Copy a complete timetable from one family member to another.
- Persist individual timetable display preferences.
- Integrate the module with navigation, permissions, API scopes, API documentation, localization, and offline PWA caching.
- Add database schema migration and REST API coverage for CRUD operations, filtering, settings, validation, deletion, and timetable copying.

## Checklist

- [x] `npm test` passes


mclgoerg:
Hi, please have a look at the schedule module which already supports work, university and school plans.

matthiasNX:
Yes, but i need a simple hour by hour overview of my kids. With teacher (name) room etc.
May i don´t understand the shedule module in the right way

image

mclgoerg:
I’ve just noticed my v3 schedule module isn’t live yet. Once it ships it’ll include this feature though for now it does not include names. This should be pretty easy to add.

ulsklyc:
Thank you for this - and for the screenshot, because it names the actual need in one picture: an hour-by-hour week for a child, with teacher and room, not a shift per day. That need is real, it is not what Schedule shows today, and I want it in Yuvomi. What follows is the decision on *where* it lives, and then what I found in the code. The decision is mine to make, so I am making it here rather than leaving you to guess what would be accepted.

**Where it lives: inside Schedule, not as a twentieth module.** This was settled in #786 at the end of August, before your PR, under the heading "one model, not two": a fixed weekly timetable is a 7-day cycle, an A/B-week timetable a 14-day one, and Schedule already carries the anchor date and the cycle arithmetic for both (`schedule_patterns.anchor_date`, `cycle_length`). #749 has been waiting since then for exactly the part you built - the timetable's own vocabulary - not for a second week engine.

The code is what convinced me, not the framing. `timetable_entries.week_type IN ('all','A','B')` has no anchor anywhere, so nothing in the module can answer "is this week A or B?" - the user picks it by hand in a dropdown, which is a label, not a recurrence. Every entry carries its own subject, times and colour, which is why the PR needs a `POST /copy` endpoint and a copy dialog; in Schedule a second pattern points at the same shift types and the copy is free. And there is no calendar overlay, so a child's school week would be invisible in the calendar, on the dashboard and in search - which is the whole reason a family runs one app instead of five.

What is genuinely new in your PR is small and good: room, instructor, period number, a category, and **more than one block per cycle day** (Schedule's `schedule_pattern_days` allows exactly one shift per position today). That is a change to the existing tables plus a person-against-time grid view in Schedule, and it also answers the question I left open in #749 - timetable entries are their own thing, kept out of the calendar's data and computed into it on read, like shifts are.

**What I found in the code, in case you take it on.** The first one has to be closed in any version:

- **Every member can rewrite or delete every other member's timetable.** `isAdmin` is defined at `server/routes/timetables.js:10` and never called; `POST /`, `PUT /:id`, `DELETE /:id`, `PUT /settings` and `POST /copy` all take `user_id` from the request body unchecked. Schedule's `mineOrAdmin(req, userId)` (`server/routes/schedule.js:16`) is the house rule for this and it is applied on the write routes there.
- The new strings exist only in `de.json` and `en.json`; the other 22 locales lack 79 keys, so `npm run test:i18n` is red on the branch, and `test:timetables` is not in the `test` chain, so `test:suite-chain` is red too. A key is added to all 24 files with a real translation each; the guard `test:i18n-translated` now counts values that are merely copied from English.
- `public/styles/timetables.css` uses six custom properties that do not exist (`--font-size-xs`, `--font-size-sm`, `--font-size-base`, `--color-surface-card`, `--color-text-muted`, `--radius-pill`) in 24 declarations - the real names are `--text-xs/sm/base`, `--color-text-tertiary`, `--radius-full` - so the cards render with no background and the type scale collapses. Raw hex colours and a module-level `max-width` are against `tokens.css` and `docs/PAGE-COMPOSITION.md`.
- Four features do not work as written: `createPageFab(root)` gets a container instead of options and the FAB is never mounted; the holiday lookup compares `h.date` while the API returns `start_date`/`end_date` ranges; the empty state passes keys the helper does not know, so it has no text and no button; the "list" view is not in the server's allowed values and silently reverts to "week". And `time(value, field)` takes two arguments, so the third one you pass does not make start and end required - a missing time reaches the `NOT NULL` column and comes back as a 500 instead of a 400.

Mechanical things you do not need to chase: migration `175` is also taken by #1019, and I renumber at merge; the `package-lock.json` diff is npm drift with no dependency change and can be dropped; the module count on the landing page is mine to keep in step.

**Two ways forward, both fine with me.** Either you rework this on the Schedule model - blocks per cycle day with subject, room, instructor and period on the existing tables, the grid as a Schedule view, the permission rule first, strings in all 24 locales - and I review it with priority. Or you tell me you would rather not, and I build it that way and name you in the CHANGELOG as the one who brought it, because the grid in your screenshot is the right picture of the feature.

One more thing, so it does not stand in the way: there is no "v3" of Schedule on the table. A feature that is not in a branch is not a reason to wait, and nothing here depends on one.

Timing, if you go the first route: anything under `public/pages/` ships with the interface release on a Tuesday.

ulsklyc (changes requested):
Setting the state the comment above already describes, so the PR page says what the thread says: https://github.com/ulsklyc/yuvomi/pull/1018#issuecomment-5533135869

Nothing new in here. The decision (timetables live inside Schedule, as more than one block per cycle day, not as a twentieth module), the findings measured on this branch, and the two ways forward are all in that comment. The next move is yours, and there is no clock on it.

matthiasNX:
implement in scheduler, please review if i fill the requirments.

mclgoerg:
I am pushing v3 of the scheduler today. Please test it and check if it satisfies your request.

matthiasNX:
@mclgoerg yes thanks, then i will wait and test afterwards

matthiasNX:
I have added also the feature, that i can add school holidays direct in the scheduler, would be greate if v-3 can also do this.
Please check also the display of a school plan. This is really relevant for a familiy with kids daily life.

image

matthiasNX:
image

mclgoerg:
I am finalizing the PR right now. Feel free to commit there aswell and add whatever you need.

mclgoerg:
The PR is up: #1022.

It brings Schedule up to what @ulsklyc described above: a shift type can carry an icon, custom
fields (Room, Instructor, whatever a household needs, not fixed columns), and a cycle day can now
hold several blocks instead of one - so a school week with multiple periods a day works. Also in
there since it was already on the branch: shift-start reminders, an overtime target with print, a
personal ICS feed, and additive extra shifts.

@matthiasNX - would appreciate you testing it against the school-plan case from your screenshots
when you get a chance. Custom fields cover room/instructor; the school-holidays-in-the-scheduler ask
isn't in this PR, happy to pick that up as a follow-up if you open a discussion for it.

matthiasNX:
I have tested your implementation, but i have some doubts.
- In real life each day can have a different Start and End Time
- This should (if possible) be calculated with the min and max start and end times of all periods (school hours) in a day (as you see in the screenshot above)
- If i add multiple periods, it will show up the count in the calender view (there should only be one entry with the complete min max time)

A school time table can look like this, and it´s not possible to reflect with the current implementation. So the question is, is it really a schedule or should we implement a seperate option for school timetables?

@mclgoerg @ulsklyc 

image

mclgoerg:
@matthiasNX re: your last message here - seeded a real weekly timetable on the schedule-test
deployment (#1022's current state) to check exactly this: a shift type per period ("1. Stunde"
08:00-08:45 through "6. Stunde" 12:30-13:15), Subject/Room/Instructor as custom fields on each,
values varying per day (Mon=6 periods, Tue=5, Wed=6, Thu=4, Fri=5 - same shape as your screenshot,
different subjects/rooms/instructors, different last-period end time per day).

**Works:** custom fields round-trip correctly end to end (define -> attach to shift type -> fill in
per occurrence), multiple blocks per day at different times work, the week repeats correctly. Week/
day view in "Full blocks" mode renders each period as its own correctly time-positioned, non-
overlapping block with time + Subject + Room inline - e.g. `08:00-08:45 - Subject: Deutsch - Room: 102`.

**Confirms what you found, plus one neither of us had said yet:**

1. Month view shows exactly the count you described - `+3 more` / `+2 more` once a day passes 3
   chips, not a real entry.
2. "Full blocks" is not the default. The default ("compact") week/day view never puts periods on the
   time grid at all - it stacks them as small untimed chips in the all-day banner, six thin pills for
   a six-period day, empty grid underneath. You'd only see the working blocks view above if you found
   and flipped that toggle in the calendar filter panel.
3. Not previously reported: those compact-mode chips, and month view's chips, never show Subject/
   Room - both call `scheduleEntryLabel()` (shift-type name + owner only). Only the blocks-mode block
   calls the custom-field overlay function. So on the view most people actually see by default, the
   whole custom-fields feature is invisible on the calendar.

No collapsed single min-max entry exists anywhere, which was your literal ask - though having seen
"Full blocks" work, separate per-period blocks (closer to how Outlook/Google Calendar show class
periods) each carrying real subject/room text may be more useful than one merged 08:00-13:15 bar. The
bigger problem is most people never see that view.

Given all three, the fix isn't a new module - it's making the calendar do, by default, what it
already does when you dig into settings: default new households to "Full blocks" (or make it
automatic once any shift type has overlay-visible custom fields), fix month view's chip label to
include the overlay text, and give month view something better than a bare count once it's real
per-period data. Happy to pick this up as a follow-up PR if that sounds right - didn't want to fold
it into #1022 unannounced.

image
image
imageimageimage

matthiasNX:
ah looks good, can you please add a screenshot, how was this defined in the schedule?
would be very helpfull. thanks in advance
@mclgoerg

mclgoerg:
Sure - this is just my test setup and the UI might not be perfect yet...
Feel free to commit to this 😀 
Maybe this should be more straightforward in the documentation.

imageimageimage

matthiasNX:
@mclgoerg that opened my eyes! Thanks, that make sense.
For example i have three kids - same school - so the hour definition is the same and i just add the fields per child.
When we now also add a overview where i have a week view, where i can select the timetables and see them side by side - then i am totaly happy! Seperate from the normal calender.

Would you please add the overview also?

Great work!

mclgoerg:
@matthiasNX @ulsklyc re: the "select timetables and see them side by side" overview request above -
before building it, this is a genuine fork with real tradeoffs either way, not a clear-cut choice.
Would like your take on where it should live before starting.

**Option A - a new mode inside Calendar's week view**
- \+ One destination, fits the existing month/week/day/agenda tab pattern
- \+ Calendar already loads every household member's schedule entries in one call, no new fetch
- \- Forces new decisions that don't have a clean answer yet: how do holidays/events/tasks behave
  once a day's column is split into per-person lanes? Stay pooled above the split (visually
  inconsistent) or also get split per person (a much bigger, unrelated change)?
- \- No existing `?view=`/persistent-view-mode plumbing in Calendar to extend - this would be new
- \- Touches the app's busiest, most general-purpose page for a request that's specifically about
  timetables, not the calendar as a whole

**Option B - a new tab in the Schedule module itself**
- \+ Already schedule-only (no events/tasks/holidays to reconcile), already tabbed (Shift Types /
  Planning / Statistics), already loads all-household entries with the custom-field overlay text
  built in
- \+ The Statistics tab already has a single-person picker - a direct precedent to extend into a
  multi-select for this
- \+ Zero risk to Calendar, the highest-traffic page in the app
- \- A second place to look if someone thinks "calendar" first for this (mitigated: anyone managing
  multiple kids' timetables is already visiting Schedule to set them up)
- \- Doesn't show non-schedule context (a dentist appointment, say) next to the comparison - though
  that's not part of what was asked

The layout itself isn't in question either way - day columns split by person (Monday shows Kid1 |
Kid2 | Kid3 side by side), matching "side by side" literally and letting you compare the same day
across kids at a glance. Just the "where" is open. Let me know which way you'd like this to go and
I'll plan the implementation from there.

matthiasNX:
@mclgoerg @ulsklyc 
Yes i share your doubts about the calendar app. In my PR i also used a seperate tab direct inside the Schedule module itself. So Option B sounds good to me.

image

Because the kids needs only the timetable for school, to plan the next day, it should be seperated.
Then only the request of adding the school holidays in the Schedule is missing.

ulsklyc:
@mclgoerg @matthiasNX - Option B, the Schedule tab. Thank you for asking before building it;
that is exactly the right call on a fork like this, and it is why this answer costs a day and
not a rewrite.

**Why B, from the data model rather than from traffic.** Every schedule entry carries a
`user_id` - `resolveEntries()` stamps it on both branches (`server/services/schedule.js:29`
for an override, `:40` for a pattern day). A per-person lane is therefore something the data
already is. Holidays, events and tasks are not: `renderWeekView()` puts `holidaysOnDay(d)`,
the all-day events and `tasksOnDay(d)` into one `allday-cell` per day, and the header row,
the all-day row and the column grid all share the same `colCount`
(`public/pages/calendar.js:2081` onward). Splitting a day column into person lanes would make
me answer "whose holiday is Christmas?" - a question the schema does not ask and should not
be made to. That is the deciding argument, not the traffic on the page.

**One correction to your tradeoff list, in fairness to Option A:** the persistent view mode
does exist. `CALENDAR_VIEW_STORAGE_KEY` at `calendar.js:325`, with `defaultCalendarView()`,
`setSavedCalendarView()` and a `normalizeCalendarView()` allowlist at `:625-648`. What is
missing is only a `?view=` URL parameter - the calendar reads `open` and `date` and nothing
else (`:1329`). So A would have been cheaper than you credited it, and it still loses on the
point above.

**One thing neither of you has hit yet, and it applies to B just as much.** A grid with one
column per person has to be told which people. `/auth/users` returns every `users` row; it
flags domestic staff with `is_worker` but filters nothing, and split-expense guests - who are
external - are not marked at all (`server/auth.js:2263`). Both pages read it (`calendar.js:1290`,
`schedule.js:72`). Today this stays invisible because only people with entries get a label; a
person grid draws the empty column too, so a household with three kids would also get a lane
for the cleaner and for an external guest. The predicate already exists once, as
`isHouseholdMember()` in `server/services/member-email.js:94` - please use that one rather
than writing a second. Selection and route both, and the test in both directions: no external
is accepted, and everyone offered is.

**Two smaller things for the build.** The Statistics tab's person picker is single-select
(`statistics.userId`, `schedule.js:20`, `:174`), so the multi-select is a real change there,
not a parameter. And check the phone early: the week view already drops to three days below
640px (`calendar.js:2084`), so three days times three children is nine columns - Schedule is
free to lay the grid out differently, which is the other reason B is the easier home.
`.schedule-stat-range__choices` at `schedule.css:335` is the existing precedent for a
horizontally scrolling rail.

**On the school holidays** - @matthiasNX, they exist already as a calendar layer, type
`school` from the holiday API (`calendar.js:1122`), with a toggle of their own. Bringing that
layer into the Schedule tab is reading data that is already there, so it belongs in this work
rather than in a separate discussion.

Timing: anything under `public/pages/` ships with the interface release on a Tuesday.

ulsklyc:
One pointer I owe you both, so the decision above is not read wider than it is.

@mclgoerg - your Option A is, at its core, #670: a multi-column family calendar, one column per
member, opened by Zeromax in August. I answered it there on 24 August, and the reasoning
matches what I wrote above from the other side: organising by person first and time second is a
genuinely different layout rather than a variant of week view, which is why it cannot be a
switch on the existing one. I also flagged the same horizontal-space problem there - seven day
columns already need care at 1024px, and a column per member scales with household size.

So: choosing the Schedule tab for timetables does **not** close #670. That request is about
calendar events, which do carry assignments per member, and I called it the request I have the
most sympathy for. It stays open and welcome. What this week's decision rules out is narrower -
putting one module's data into another module's page to get a person-first view - and it is
going into DECISIONS.md as entry 6, in #1025.

#670 also still has a question of mine open to its reporter - how many people side by side, and
on what screen - because that is what decides the behaviour when the columns no longer fit. The
Schedule tab meets the same problem at a smaller scale, so whatever it settles there will be a
useful precedent.

mclgoerg:
@matthiasNX @ulsklyc the Overview tab is now included in #1022 - a new tab in the Schedule module
where you pick several household members and see their weekly timetables side by side (day columns,
one lane per selected person, fixed lane order so a lane stays put day to day). Went through two
rounds of live-feedback revision:

- The grid uses a collapsed hour scale - only hours someone in the current selection actually
  occupies get a row on the axis, so a 45-minute class period and an 8-hour work shift can share a
  timeline without either becoming unreadable or squeezed to a sliver.
- A Week/Day toggle switches the visible range without ever refetching.
- An overnight shift (e.g. a rotating work shift crossing midnight) now carries a continuation onto
  the next day's lane, carrying the real end time - it previously only showed its pre-midnight half.
- School and public holidays surface as a banner above the grid, reusing your existing Settings
  toggles rather than adding a new one.

@matthiasNX - this should cover the "three kids, same school, compare side by side" case directly.
Would appreciate you trying it against that scenario when you get a chance.

ulsklyc:
@matthiasNX - you did what I asked for on 3 September, and I owe you a straight account of where
that leaves this PR, because it is not the answer you were hoping for.

## First, what you changed

I read `8442dd99` rather than the description. The standalone Timetables module is gone -
`timetable_entries` and `timetable_settings` are not in `server/db.js` any more and no file on the
branch carries the name - and it is now +1644/-65 across 34 files instead of +2250/-61 across 32.
The rights gap from my review is closed: `isAdmin` is actually called now, and every per-person
write route goes through `mineOrAdmin`. The only three write routes without it are
`POST`/`PUT`/`DELETE /shift-types`, which are household-wide and identical to `main` in that
respect. All 24 locale files move by the same 32 lines, so nothing is left behind in de/en. That
was the hard part of the ask and you did it.

## Where things stand

While you were rebuilding, #1022 grew a Schedule Overview tab (`fdfbbed9`, 4 September) that
covers the same ground - the side-by-side person grid you asked for. That is my doing as much as
anyone's: I decided the architecture in this thread, @mclgoerg built it in his PR, and nobody
checked what that would mean for yours. You should not have to find that out by reading a diff.

The two branches are not two halves of one feature. A trial merge of the two heads conflicts in 30
files - `server/db.js`, `routes/schedule.js`, `services/schedule.js`, `pages/schedule.js`,
`schedule.css`, all 24 locales and SPEC. Of the seven things your PR does, four are also done in
#1022:

| | #1018 | #1022 |
|---|---|---|
| Several blocks per cycle day | migration 176 (rebuild, UNIQUE dropped) | migration 182, the same rebuild |
| Subject / room / instructor / notes | fixed columns on `schedule_pattern_days` | `schedule_custom_fields` registry + per-occurrence values (migration 183) |
| Colour, category per block | columns `color`, `category` | carried by the shift type |
| Person grid | `renderTimetable()`, week/day/list, one member or all | `renderOverview()`, week/day, multi-select |

On the second row I have to rule for #1022, and not because it arrived second. Fixed columns for
subject/room/instructor answer the school case and only the school case; the next request (a
course code, a group, a client reference) needs another migration each time. That is the same
"one model, not two" line I applied to your original module proposal, now recorded as entry 6 in
`docs/DECISIONS.md`. Applying it to your PR and not to #1022 would make it a rule about who posted
first.

## What is only in yours

Three things, and I want them:

1. **A time per block.** Your migration 177 puts `start_time`/`end_time` on
   `schedule_pattern_days`, so a lesson can be 08:00-08:45 without creating a shift type for it. In
   #1022 a timetable needs one shift type per period. Both are defensible and I do not think this
   is settled - a period is arguably a reusable time slot that deserves a name, but eight of them
   before you can enter a single week is a real cost.
2. **`scheduleDayGroups()` in the calendar.** With several blocks per cycle day, the calendar's
   week and day views now show every period. Yours collapses them to one per person per day.
   #1022 does not touch this, so the question is open either way; I would want to look at whether
   collapsing or keeping them is right before picking, since a day view arguably should show all
   eight.
3. **Materialising school holidays.** Your `apply-holiday` and `apply-all-holidays` actions POST
   to `/schedule/overrides/fill`, so loaded holidays become real override rows. #1022 only draws
   them as a banner from the existing Settings toggles. My own preference is to compute this on
   read rather than write rows that then have to be kept in step when the holiday data changes,
   so I would not take the picker as it stands - that is a call, not a written rule. But the need
   behind it, that holidays should actually suspend the pattern rather than just be drawn over
   it, has no answer today and is worth its own issue.

Three things in the current branch would need fixing whichever way this goes, so you have them.
The person dropdown feeds from `state.users`, which is `/auth/users` (`public/pages/schedule.js:81`),
so "all members" includes housekeeping staff and split-expense guests, each getting an empty row of
their own. That is the point I raised in this thread on 4 September as due in either design, and it
is what `isHouseholdMember()` in `server/services/member-email.js` is for. Then:
`render()` sets `activeView = 'timetable'` on every entry into the module
(`public/pages/schedule.js:1091`), so every existing Schedule user lands on the timetable grid and
the tab cannot be remembered; and `timetableWeekStart` hardcodes Monday
(`startOfLocalWeekKey(todayKey(), 1)`, line 23) instead of reading the household `week_start`
preference, which supports `monday`, `sunday` and `saturday`.

## What I propose

#1022 lands first, as the base. Then the three items above become their own small pieces, and I
would rather you built the ones you want than that I did. Say which, and I will hold the others.

Either way you are named in the CHANGELOG entry for this work. You brought the requirement, you
sized the problem correctly, and you rebuilt a 32-file PR on someone else's architecture call
inside a day. If you would rather step back from it, that is a fine answer too and I will say so
in the entry.

One concrete thing that would help right now regardless: @mclgoerg asked you to try the Overview
tab against your actual case, three children at the same school. Nobody else in this thread has
that case in front of them.

matthiasNX:
@mclgoerg @ulsklyc thanks for your work. i have tested the feature from mcgoerg.
Works good for me, so i guess we can remove my pull request.

Only two things to think about....
- When i filter in the calender module, i see all schedules, even when i filter for person - should be the schedule entries also be filtered by person?
- the text size for the custom fields in the block is really very very small, but that a really small issues.

When the pull request is merged, i can go on, and implement the feature with the automatic add of the school holidays.
Thanks!

mclgoerg:
@matthiasNX thanks for testing, and for that catch - it was real. `scheduleEntriesOnDay()` never checked person selection at all, so filtering the calendar to one person left every household member's shifts showing regardless, in month/week/day view alike. Fixed in #1022 (`7e49b3bf`): it now reuses the same `passesPersonFilters()` events and tasks already use, so a schedule entry's single owner is checked exactly the same way an event's or task's assignee is. Covered by a new test, full `npm test` green.

On the custom-field text size: left it as-is for now. `.week-event`/`.day-event` (which the schedule block inherits from) already sit at `--text-xs`, the smallest size in the design system's type scale - it isn't unique to the custom-fields overlay, every calendar entry in week/day view is that size. Making the schedule block bigger than other entries would be visually inconsistent with the rest of the grid, and going smaller isn't available without a new token. Since you called it minor yourself, happy to leave it for @ulsklyc to weigh in on if it's worth its own follow-up.

Good news on the overview tab satisfying the three-kids case - and noted re: automatic school holidays once this merges.

ulsklyc:
@matthiasNX Closing this properly, because you deserve more than a silent close.

**Thank you for withdrawing it yourself.** You did the harder thing: you tested the alternative, confirmed it worked for your case, and gave up your own 34-file branch rather than argue for it. That is not the usual outcome of two people building the same feature, and it is the reason this ended well instead of stalling.

**Your finding is in the code, and now your name is too.** The person filter never reached Schedule entries - `scheduleEntriesOnDay()` did not check the selection at all, so filtering the calendar to one person still showed every household member's shifts, in month, week and day view alike. mclgoerg fixed it by reusing the same `passesPersonFilters()` that events and tasks already use. The changelog entry named the issue number but not you; that is corrected on the integration branch, so it ships with the release:

> **The calendar's person filter and "assigned to me" filter now also apply to Schedule entries** (#1018). Found by @matthiasNX while testing this module against his own timetable work, which he then withdrew in favour of it.

**And the holiday automation is yours.** That was the piece your PR carried that #1022 does not, and it stays your next piece of work whenever you want to pick it up.

## Your second point, which nobody answered

You said the custom-field text in the block is "really very very small". I measured rather than guessed, and the honest answer is that it is small on purpose, but not as deliberately as you might assume:

- The block's text is `--text-xs` (12px), **inherited** from the calendar's event styling - it is not set small for Schedule specifically. There *is* a smaller step in the token scale (`--text-2xs`, 10px), used for things like reminder counters, and the block does not use it.
- What is deliberate is the two-line layout: title plus badge on one line, time plus field value on the other, each single-line with an ellipsis. The reason is in a comment above the rule - with free-flowing text, a 30-minute block in week view was already filled by title and time, and a third line for the field sat below the block edge under `overflow: clip`, invisible.

So making it larger is not blocked by a rule, it is blocked by arithmetic - and the numbers are tight. Week view is `--cal-hour-height: 56px`, so a 30-minute block is **28px tall**, minus its own vertical padding. Two lines of 12px text with normal line spacing already sit at roughly that, which is exactly why the layout was rebuilt into two fixed single-line rows with ellipsis rather than flowing text. Step the font up and the second line is the one that stops fitting - and the field value is on it.

**Which means your report is still a real one, and the fix is probably not the font size.** If the field matters more than the time in your use, the interesting question is which of the two owns that second line - or whether a taller row in week view is worth it for households that use custom fields. Worth its own issue if it still bothers you once v3 ships; I would rather have that conversation with your actual weekly schedule in front of us than guess at a number now.

Thanks again - both of your points were real, and one of them is fixed because you looked.

Issue #1010, opened 2026-09-03
test:i18n-plural only checks one direction: a counting key with no variant is invisible to it


ulsklyc:
`test:i18n-plural` checks that every plural variant has a counting base key. It does not check the other direction, and the other direction is where the mistake actually happens.

## The gap

`test/test-i18n-plural.js:154`, first line of the loop body:

```js
for (const [key, value] of entries) {
  if (!/_(one|two|few|many|other)$/.test(key)) continue;
  if (typeof value !== 'string' || !value.includes('{{count}}')) continue;
  const base = key.replace(/_(one|two|few|many|other)$/, '');
  assert.ok(entries.has(base), `${file}: ${key} ohne Basisschlüssel ${base}`);
  assert.match(entries.get(base), /\{\{count\}\}/, `${file}: ${base} zählt nicht`);
}
```

The loop starts at the *variant*. A base key that carries `{{count}}` and has no variant at all has no suffix, so the first `continue` skips it and no assertion ever sees it. Adding a variant and forgetting the base is caught; forgetting the variant entirely is not — and the second is the one people actually do.

Nothing else covers it: `test:i18n` compares key sets and placeholders, not plural completeness, and `test-detail-view.js:51` walks a hand-maintained list.

## How it surfaced

Two independent PRs reviewed on 2–3 September each shipped three counting keys with no `_one`, and both had green CI:

- #997 — `documents.folderUpload.selectedFolder`, `.uploadAction`, `.uploadedToast`. With a one-file directory the submit button reads *"1 Dateien hochladen"*.
- #989 — `deleteFolderKeepDocuments`, `deleteFolderWithDocuments`, `folderDeletedWithDocumentsToast`.

Six violations, two contributors, no red check. That is not a contributor mistake, it is a missing guard.

## Do not just add the reverse check

Measured on `main` (`de.json`): **67** counting base keys have no `_one`. A strict reverse assertion would go red 67 times on its first run, and a guard that fails that loudly on existing code gets weakened rather than satisfied — after which it checks nothing again.

The 67 fall into three groups, and they need different answers:

**1. Real violations.** `dashboard.rewardsPending` (`"{{count}} zu bestätigen"`), `dashboard.healthRefill`, `tasks.overdueDay` — grammatically wrong at n=1.

**2. Hand-built plurals that bypass the mechanism.** At least seven pairs, in three different spellings:

[...]

PR #1019, opened 2026-09-03
feat(health): cycle tracker feature parity - reminders, graded symptoms, BBT confirmation, trends, and an ICS feed

mclgoerg:
## Summary

Brings the Health module's cycle tracker closer to feature parity with the three dominant iOS
period-tracker apps (Flo, Clue, Natural Cycles), while staying inside yuvomi's existing constraints:
prediction logic stays pure/client-side, cycle data stays excluded from the caregiver-grant system
(#584), and nothing here claims to be a medical device. Six phases, all independently reviewable by
commit, plus a handful of fixes found by using the feature after the fact.

### Added

- **A more honest cycle-length prediction, plus two opt-in reminders.** `cycleStats()` now requires
  at least three logged cycle gaps (four periods) before trusting a derived average over the 28-day
  default - with only one or two gaps a single atypical cycle could silently drive the prediction,
  and the UI had no way to tell a genuine derived average apart from a coincidental match with the
  default. A new `insufficient_history` source value, surfaced as a caption on the cycle stat card,
  makes the distinction visible. Two new opt-in reminders - an upcoming-period notice (configurable
  lead time) and a daily nudge to log today, suppressed once a log for the day exists - reuse the
  same push/notification-channel pipeline as every other reminder source (`reminders.entity_type`
  widened for `cycle_period`/`cycle_log_nudge`, migration 175). Neither a predicted date nor "not yet
  logged" is a stored row, so both anchor to a new `cycle_reminder_anchors` table, the same pattern
  Schedule uses for its own computed-on-read entries. `server/services/cycle-reminders.js` reuses
  `predictCycle()` from `public/utils/health-cycle.js` directly rather than a second copy of the
  prediction math.

- **Graded symptom logging.** The symptom picker grew from 10 to 20 presets, and each one accepts an
  optional mild/moderate/severe rating - one tap on a symptom chip cycles through off, mild,
  moderate, severe, and back to off. Storage moved from a comma-separated column to a normalized
  `cycle_day_log_symptoms` table (migration 176, backfilled from the old column, which is now frozen
  and no longer read or written). Also fixed a latent bug this surfaced: the month calendar's "does
  this day have a log" check treated the `symptoms` field as a plain truthy value, which is correct
  for a string but wrong for an array - an empty array is truthy in JavaScript.

- **Optional basal body temperature tracking, with temperature-shift ovulation confirmation.** A day
  log accepts a temperature reading and unit (migration 177); `detectTemperatureShift()` implements
  the standard "3-over-6" coverline method - the first reading at least 0.2°C above the mean of the 6
  preceding readings, sustained for 3 readings in a row, confirms ovulation. When a shift is found in
  the current cycle, the prediction's ovulation date, the fertile window, the ring marker, and the
  stat card label all switch from "predicted" to "confirmed"; future cycles keep using the calendar
  method, since they have no readings yet.

- **A Trends section**, with a cycle-length chart, a basal-temperature chart, and a symptom-frequency
  breakdown, each appearing only once there's enough history to be worth showing. Reuses the existing
  shared chart geometry (`public/utils/chart.js`, already used by Vitals/Labs/Activity) rather than a
  new charting system. Symptom frequency groups every logged symptom into menstruation, luteal (PMS
  window), or other, based on each cycle's own actual length.

- **The cycle ring shows the current cycle day as its own badge**, connected to the "today" marker by
  a short line, instead of packing the day number into the ring's small center.

- **Symptom frequency now surfaces severity, not just count** (a three-dot indicator, reusing the
  day-log editor's own component), plus an expandable severity-trend chart for symptoms with at least
  two graded readings.

- **A second expandable view per symptom: which cycle day it typically lands on.** For up to the six
  most recent cycles, a plain-language sentence ("You logged Cramps during Menstruation in 3 of 4
  recent cycles") plus a compact per-cycle grid, day cells colored by phase with an inset ring
  marking a day the symptom actually occurred - a ring rather than a second color, so it survives
  color-blindness.

- **The cycle-length trend is now a bar chart**, colored by whether each cycle falls in the typical
  24-38 day range, against a shaded band for that range - the same visual language the lab-value
  chart already uses for a normal range. Also fixes a real gap: the cycle-variation stat card
  previously only appeared when fertility tracking was off - with it on, the default, cycle variation
  was never shown anywhere. It's unconditional now.

- **Symptom likelihood prediction with a calendar overlay.** Pick a symptom (only ones with enough
  logged history appear) to see a "Cramps often occur around this day in your cycle" note when today
  matches, plus markers on the existing month calendar - a filled dot for a day it was actually
  logged, a ring for a day it's predicted but not yet logged.

- **A read-only predicted-cycle ICS feed** (migration 178, `users.cycle_feed_token`), the same
  Lock-Screen/Calendar-app trick already used for the household calendar and inventory warranty
  dates. Unlike the inventory feed, this one's *content* - not just the access token - is personal:
  it reflects only the subscriber's own logged and predicted periods (plus ovulation/fertile window,
  if tracked). Manage it from Settings → Personal → Feeds, alongside the other two feeds already
  there.

### Fixed

- The cycle-length trend chart's bars sat flush against the plot edges, with no date under most of
  them - bar centers reused the line charts' point-position formula, which is correct for a
  zero-width point but wrong for a bar with real width. Bars now sit in their own equal-width band
  with padding to both neighbors and both plot edges, and every bar gets its own date underneath.
- The cycle-day pattern sentence now says how many days before your period a symptom typically shows
  up, when that's a real pattern, instead of only the coarser "occurs during your luteal phase."
- The cycle-length trend chart's typical/atypical bar colors had no legend - only a hover tooltip,
  invisible on touch. Added the same legend component the calendar and symptom-frequency chart
  already use.
- The cycle ring's phase colors also had no legend. Added a compact ring-scoped legend whose swatches
  match the ring's own solid arcs and filled ovulation dot, rather than reusing the calendar
  legend's washes/hatching/hollow-ring styling verbatim, which described a different picture than
  the ring actually renders.
- The health-structure guard's expected route table didn't know about the new cycle-feed routes -
  caught by a full `npm test` run, not the feature's own test suite.

Fixed from review (@ulsklyc, 2026-09-04):

- 68 of the 69 new locale keys had shipped as literal English text in all 22 non-German/English
  locales (`ring.dayBadge` was the one exception, already real). All 68 are now genuinely translated;
  the `history`/`insufficientHistory` cycle-stat strings also gained the `_one` singular variant a
  `{{count}}` key needs, and lost an em dash neither is allowed to carry.
- `buildCycleFeed()`'s projected/predicted window used `now.toISOString().slice(0, 10)` for "today" -
  UTC calendar day, not the household's, so west of UTC in the evening it silently ran a day behind.
  Now uses `todayKey()`, same as `cycle-reminders.js` already did.
- Push and toast bodies for both cycle reminder kinds showed the raw `anchor_date` - "Zyklus" over
  "2026-09-10" doesn't say whether the period is expected or the day just isn't logged yet. A `cycleBody()`/
  `cycleReminderBody()` pair (server push via household locale, client toast via the reader's own)
  says which, the same way `subscriptionBody()`/`warrantyBody()` already do for their kinds.
- `cycleTabEnabled()` re-implemented the household-default-plus-personal-opt-out rule that
  `healthCycleViews()` already owns in `server/routes/preferences.js` (now exported and reused
  instead of a second copy).
- `upsertCycleReminder()` read, deleted and inserted a user's anchor outside a transaction, while the
  schema only enforces `UNIQUE(user_id, anchor_date, kind)` - an interrupted sync could leave a second
  anchor no later run's `.get()` would ever see. `syncCycleRemindersForUser()` now wraps the whole
  per-user sync in `database.transaction()`.
- `/feed/cycle/:token.ics` keeps serving even when the Health module or Cycle tab is locked - not a
  leak (the content stays the token owner's own), but a deliberate choice now documented at the route
  instead of silently living only in the reminder-sync gate. Same known gap as the inventory feed.
- `GET /cycle/logs` ran one symptom query per row; batched into a single `WHERE day_log_id IN (...)`.
- The daily log-nudge reminder lingered until the next periodic sync after actually logging today (or
  adding/editing a period). `POST /cycle/periods` and `POST /cycle/logs` now call
  `syncCycleRemindersForUser()` immediately, same as the settings-change path already did.
- Two raw pixel values in `health.css` (a calendar-day symptom marker, a callout icon) now use
  `--space-1`/`--space-1h`/`--icon-md` like the rest of the file.

### Database migrations

Four new migrations, 175-178 (`server/db.js`):

| Version | Adds |
|---|---|
| 175 | Widens `reminders.entity_type` for `cycle_period`/`cycle_log_nudge`; adds `cycle_reminder_anchors`; adds `cycle_settings.remind_period_days_before`/`remind_log_daily` |
| 176 | Adds `cycle_day_log_symptoms`, backfilled from the legacy `cycle_day_logs.symptoms` CSV column (now frozen, read-only) |
| 177 | Adds `cycle_day_logs.basal_temp`/`basal_temp_unit` |
| 178 | Adds `users.cycle_feed_token` + partial unique index |

Originally numbered 174-177 against an older `main`; renumbered during this rebase after `main`
picked up its own migration 174 (birthday name days, #993) while this branch was in flight.

### Not in scope here

A few adjacent ideas came up while scoping this (fertile-window/ovulation reminders, cervical mucus
and OPK logging, a printable doctor summary, sex/intimacy tracking, explicit irregular-cycle
messaging) - deliberately left out of this PR rather than half-included. None of them are blocked by
anything here; happy to open a follow-up discussion if there's interest.

## Test plan

- [x] Full `npm test` - green, run after rebasing onto current `main`, and again after addressing
      review
- [x] New suites: `test-cycle-reminders.js`, `test-cycle-symptoms-migration.js`, `test-cycle-ics.js`;
      `test-health-cycle.js` extended throughout (91 tests)
- [x] i18n: all new strings translated across all 24 supported locales - genuinely, not just
      keyset-complete; verified with the same copied-from-English check `test:i18n-translated` runs
      on `main`, zero hits
- [x] Manually verified end-to-end against a live deployment (`schedule-test.streamingsvc.com`) with
      realistic seeded data: reminders firing, symptom intensity UI, BBT input and the
      confirmed-ovulation marker, all three trend charts, the severity-trend and cycle-day-pattern
      disclosures, the bar chart's typical/atypical coloring, symptom-likelihood prediction and its
      calendar overlay, and the ICS feed subscribed in a real calendar app - light and dark mode,
      desktop and mobile widths


ulsklyc (changes requested):
Reviewed at 09e9503c, the whole diff, with the chain run on the branch (275 suites, all green on the branch's own state). This is a lot of careful work, and the part I worried about most is the part that is cleanest: every new or changed route stays behind `viewerId()`/`visibilityClause()` (`server/routes/health/cycle.js:189-198`, `:209-247`, `:290-350`, `cycle-feed.js:29-58`), nothing takes a `user_id` from the body, there is no admin branch, the feed token is 32 random bytes with revocation, a partial unique index, `private, no-store` and a 404 on unknown tokens, and the ICS text carries only translated summaries and dates. That is exactly the rule in `docs/DECISIONS.md` entry 1, kept without exception. The scope question is settled too, so it does not sit with you: this extends the Cycle tab on the existing reminder pipeline, `chart.js` and the feed pattern rather than building a second model, and that is in.

Three things block the merge, and the second is the big one.

## 1. The migration numbers collide with `main`

`main` is at 175 (`Permissions: allow fine-grained capability resources`, since #996). The branch takes 175-178 (`server/db.js:6958`, `:7014`, `:7050`, `:7064`). Renumbering is mine at merge time, along with the mirror and the places that select by number (`test/test-cycle-symptoms-migration.js:21`, `test-cycle-ics.js:8,164,167`, `test-health-structure.js:114`, `cycle-reminders.js:15`, `cycle.js:40,195`, `personal-feeds.js:10`, the SPEC and CHANGELOG lines). I name it so you know why the numbers in the merged commit will differ from yours; the final ones depend on what lands first.

## 2. 22 of the 24 locales carry the English text, not a translation

The description says all new strings are translated across all 24 locales. Measured on the branch: 69 new keys, present in every file, but in ar, cs, el, es, fa, fil, fr, hi, hu, id, it, ja, ko, nl, pl, pt, ru, sv, tr, uk, vi and zh, 68 of the 69 values are byte-for-byte the English string (`fr.json:4115`: `"Default for now — log {{count}} more period(s) for a personalized prediction"`; the one exception is `health.cycle.ring.dayBadge`). Since 5bd02b26, `main` has `test:i18n-translated`: it counts, per locale, the texts that are word-for-word English while the German differs, and holds that number against a baseline that may only fall. After the rebase this branch turns that suite red in all 22 locales. It passed for you only because the guard did not exist on your branch. Real translations are needed for each locale; Inventory and Schedule went through the same exercise on `main` last week if you want the pattern.

## 3. Two `{{count}}` keys without a `_one` form, and one em dash

`health.cycle.stats.source.history` and `health.cycle.stats.source.insufficientHistory` (`de.json:4114-4115`, in every locale) take a numeric `count` from `public/pages/health.js:4511` and `:4516`, so with one period the German reads "1 Perioden". CLAUDE.md asks for a `_one` variant on every key with a numeric count; `test:i18n-plural` runs one way only (#1010) and let it through. The second of those two also carries an em dash in `de` and `en`, and CONTRIBUTING rules that out for locale values; a hyphen or a full stop does the job.

## Should fix

- `server/services/cycle-ics.js:110` hands `now.toISOString().slice(0, 10)` to `projectFutureCycles()`, so west of UTC the feed's "today" is yesterday in the evening. `cycle-reminders.js:169` does it right with `todayKey(database, now)` from `server/utils/timezone.js`; the same call belongs here.
- The push and toast body for both cycle kinds is a bare ISO date: `entity_title` is `anchor_date` (`server/services/notifications.js:335-336`, `server/routes/reminders.js:170-171`), the payload uses it as the body (`notifications.js:150`) and the toast as its text (`public/reminders.js:263`). "Zyklus" with "2026-09-10" under it does not say whether the period is expected or the day is unlogged. `translate(locale, ...)` is already in `reminderPayload()`; a branch per kind, like `subscriptionBody()` and `warrantyBody()`, would say it.
- `cycleTabEnabled()` in `cycle-reminders.js:45-50` re-implements the rule from `healthCycleViews()` in `server/routes/preferences.js:228` (household default plus personal opt-out), and its comment says so. `docs/DECISIONS.md` entry 2 is about exactly this: the rule lives in one exported function and both sides call it.
- `upsertCycleReminder()` (`cycle-reminders.js:72-104`) deletes the old anchor and reminder and inserts new ones outside a transaction, while reading "the one" anchor per user and kind with `.get()`. The schema only guarantees `UNIQUE(user_id, anchor_date, kind)` (`db.js:7000`), so an interrupted sync leaves a second anchor no later run sees, with its own reminder. Either `UNIQUE(user_id, kind)`, which is what the module comment promises, or the per-user sync inside `database.transaction()`.
- The reminder sync stops when the Health module is `none` or the Cycle tab is locked (`cycle-reminders.js:162-167`); the feed keeps serving on the token alone (`server/index.js:405-416`, `cycle-ics.js:157-161`). No third party gets anything, the content is the token owner's own, but it is two answers to one question inside one PR. The Inventory feed has the same gap, so either say in a comment that this is deliberate or gate `findUserIdByFeedToken`.

## Worth a look, not blocking

`GET /cycle/logs` runs one symptom query per row (`cycle.js:198`); a `WHERE day_log_id IN (...)` would do. After `POST /cycle/periods` and `/cycle/logs` the log nudge for today stays until the next scheduler tick. `public/styles/health.css:1941-1943` and `:2126` carry raw px where the file otherwise uses tokens. The SPEC additions use em dashes, which the SPEC already does elsewhere, so that one is on the house style rather than on you.

Once the three points above are in, the rest of the mechanics is mine: the rebase, the 22 locale files that now collide with the retranslation on `main`, the CHANGELOG position and the migration numbers. This lands after #997, #989 and #991 in the published order.

mclgoerg:
Thanks for the detailed pass. Addressed everything below and pushed.

## Blocking

1. **Migration collision** - left alone, as noted; renumbering is yours at merge time.
2. **22 locales carrying English text** - fixed for real. 68 of the 69 new keys were translated
   (`ring.dayBadge` was already correct); verified with the same copied-from-English check
   `test:i18n-translated` runs on `main` (`copiedFromEnglish(de, en, locale)` scoped to just these 69
   keys) - zero hits across all 22 locales now.
3. **`_one` forms + the em dash** - added `history_one`/`insufficientHistory_one` to all 24 locales,
   and replaced the em dash (and the `Periode(n)`-style hack it was next to) with a plain hyphen in
   `en`/`de`. `typicalRangeLabel` had an en dash too (not flagged, but same new key, same rule) -
   fixed alongside it.

## Should fix

- `cycle-ics.js:110` now uses `todayKey(conn, now)` instead of raw UTC.
- Added `cycleBody()` (server, `notifications.js`) and `cycleReminderBody()` (client,
  `public/reminders.js`) so push and toast say "Next period - 2026-09-10" or the log-nudge text
  instead of a bare date, same shape as `subscriptionBody()`/`warrantyBody()`.
- `cycleTabEnabled()` is gone; `healthCycleViews()` in `preferences.js` is now exported and reused.
- `syncCycleRemindersForUser()` wraps its body in `database.transaction()`.
- Documented the feed/reminder-gate asymmetry at the route in `server/index.js` rather than leaving it
  implicit - deliberate, matches the inventory feed's existing behavior.

## Worth a look

Took all three: `GET /cycle/logs` batches symptoms via `WHERE day_log_id IN (...)` now instead of one
query per row; `POST /cycle/periods` and `POST /cycle/logs` call `syncCycleRemindersForUser()`
immediately instead of leaving today's log-nudge until the next scheduler tick; the two raw px values
in `health.css` are `--space-1`/`--space-1h`/`--icon-md` now.

Full `npm test` green after all of the above. PR description's Fixed section and Test plan updated to
match.

ulsklyc (approved):
Re-reviewed at `2b1088df`. I checked all eleven points against the code rather than against your
summary, and every one of them holds. Approving; what is left is mine.

**The two blocking points you could act on are closed.**

The translations are real this time, and I measured it the way the guard does rather than taking the
description's word for it: I lifted `copiedFromEnglish()` out of `main`'s `test-i18n-translated.js`,
scoped `en` to the 74 keys this branch adds that `main` does not have, and ran it across all 22
non-de/en locales. **Zero hits**, against 44 keys long enough for the guard to judge. Both `_one`
forms are in place (`history_one`, `insufficientHistory_one`), all 74 new `{{count}}` keys have a
singular, and the em dash is gone from the new values - the three left in `de.json` are pre-existing
on `main` and none of them are yours.

**The five should-fixes are all in, and three are better than what I asked for.**
`cycle-ics.js:111` now takes `todayKey(conn, now)`. `cycleBody()` and `cycleReminderBody()` branch per
kind and go through `translate()`/`t()`, with a comment explaining why the server picks the household
language. `syncCycleRemindersForUser()` wraps the whole sync in `database.transaction()` - including
the teardown branch, which I had not asked for and which needed it just as much - and the comment
states the exact invariant the schema does not give you. The batch query for `GET /cycle/logs` is
properly parameterised. The feed gate is documented at the route with a reason I accept: an existing
subscription on another device should not go quiet because a view is locked in the app.

One correction to your own summary, and it is in your favour: `cycleTabEnabled()` is not gone. It is
still at `cycle-reminders.js:47`, but as a one-line wrapper around the now-exported
`healthCycleViews()`. That satisfies entry 2 of `docs/DECISIONS.md` exactly - the rule lives in one
place and the local name just reads better at the call site. I mention it only because the next person
reading the thread would go looking for a deletion that did not happen.

**Nothing new turned up.** I went looking specifically at what the fixes exposed: the new `IN (...)`
batch cannot hit SQLite's variable ceiling in any realistic household (the limit here is ~32k bound
parameters), the layer-boundary guard is green on the `services` to `routes` import, and the ICS feed
is all-day only (`VALUE=DATE`), so it does not have the timezone problem I just flagged on your #1022
- that one only bites timed events.

Chain on the branch: **279 of 280 links green**. The single failure is `test:changelog`, and it is
branch age, not your work: the branch predates the `v2.64.1` tag, so the guard sees a tag with no
CHANGELOG heading. It clears on rebase. I ran the 175 links that sit after it separately to be sure
nothing was hiding behind the abort.

## What is left, all mine

The migration numbers moved again while this sat: `main` now ends at 176, and your 175 and 176 both
land on occupied numbers (#996's capability resources and #991's note categories), so this becomes
177-180 at merge time, along with the mirror and the by-number references in
`test-cycle-symptoms-migration.js`, `test-cycle-ics.js`, `test-health-structure.js`,
`cycle-reminders.js`, `cycle.js`, `personal-feeds.js`, the SPEC and the CHANGELOG. Worth knowing that
#1018 and #1022 are queued on the same numbers, so the final ones depend on landing order.

25 conflicts against `main`, as predicted: `CHANGELOG.md`, `package.json` and 22 locale files that
collide with the retranslation work on `main`. Those are mechanical but not blind - on #989 a
"take theirs" quietly restored two keys the PR had deliberately withdrawn, so I will diff the merged
locale set against your branch key by key rather than trusting the resolution.

ulsklyc:
Merged as `500e0864`. GitHub could not push into your fork branch from here, so the tested tree went
onto `main` as a merge commit with your head (`2b1088df`) as its second parent; the branch itself is
untouched.

What changed during the merge, so the numbers in `main` differ from yours:

- **Migrations 175-178 became 177-180.** `main` had taken 175 (#996, capability resources) and 176
  (#991, note categories) while this was in review. The by-number references moved with them:
  `test-cycle-symptoms-migration.js` (including `V176`/`seedPreV176`), `test-cycle-ics.js`,
  `test-health-structure.js`, `cycle-reminders.js`, `cycle.js`, `personal-feeds.js`, `docs/SPEC.md`,
  `docs/test-suites.md` and the three CHANGELOG lines.
- **The 22 locale conflicts were resolved three-way, not by taking one side.** `main`'s retranslation
  wins every value it changed (roughly 215 per file), your branch contributes its new keys, and your
  removal of the orphaned `health.cycle.status.regularity` was carried through - a plain "take theirs"
  would have restored it in 22 files while `de` and `en` dropped it. All 24 locales end on an
  identical key set.
- **The CHANGELOG entries went into `Added` and `Fixed` under `Unreleased`**, merged into the existing
  sections rather than appended as a second pair of headings.
- `package.json` kept `main`'s test chain and gained your three suites after `test:health-cycle`.

Full chain on the merge tree: 284 links, all green, run against the dependency bump that landed just
before this (#1023).

Thanks for the second pass - the translations, the transaction and the reminder bodies all held up
when I measured them rather than read them.

Issue #818, opened 2026-08-19
[Bug] Calendar ICS feed exports floating timestamps without TZID / X-WR-TIMEZONE (interpreted as UTC/GMT+0 by external clients)


ralonsoespana:
When subscribing to the calendar feed via `/feed/calendar/:token.ics`, non-recurring local events are exported as floating local date-times (e.g. `DTSTART:20260820T160000`) without `TZID`, without a `VTIMEZONE` component for the server timezone, and without an `X-WR-TIMEZONE` calendar header.

Because `DTSTAMP` is formatted in UTC (`DTSTAMP:...Z`) and no timezone reference is attached, most external calendar subscribers (such as Google Calendar, Apple Calendar, Home Assistant, Thunderbird, and Outlook) default floating dates to UTC/GMT+0. As a result, events are shifted (e.g., an event at 16:00 Madrid time / UTC+2 is displayed at 18:00 or registered under GMT+0).

Steps to reproduce

1. Set the container environment variable `TZ=Europe/Madrid`.
2. Create a local calendar event (e.g. at 16:00 on any date).
3. Retrieve the subscription feed via `/feed/calendar/<token>.ics`.
4. Inspect the generated `.ics` file or import it into Google Calendar / Apple Calendar / Home Assistant.
5. Notice that `DTSTART:20260820T160000` lacks `TZID=Europe/Madrid` or `X-WR-TIMEZONE`, causing the client to display it at 18:00 CEST.

ulsklyc:
Fixed in **v2.24.3**.

Locally created events store plain wall-clock time with no offset, so they went into the feed as floating local times. Per RFC 5545 that is valid and means "the viewer's clock" - in practice Google, Apple, Thunderbird, Outlook and Home Assistant all read such values as UTC, which is exactly the 16:00 to 18:00 shift you saw under `TZ=Europe/Madrid`.

The digits stay, they just get their zone now:

- `DTSTART;TZID=<zone>` plus a matching `VTIMEZONE` component and `X-WR-TIMEZONE` in the calendar header, taken from the server timezone.
- Events that already carry an unambiguous offset (everything imported from Google or CalDAV) keep their UTC form.
- All-day events stay `VALUE=DATE`.
- `EXDATE` is anchored the same way; otherwise an exception would no longer match its own occurrence.
- If the household zone is UTC or cannot be resolved, the naive values get a plain `Z` instead of a `VTIMEZONE` over a zone many clients do not carry.

[...]

Issue #549, opened 2026-07-22
[Bug] CalDAV Looping Events don't work correctly


LEON-BROTHER:
## Description

In my calendar I have a event that loops through the whole week except weekends.

The problem is that for some reason the calendar won't show it, my other events that aren't looping are working and showing up.

I use Baikal.

I also noticed for some reason that events aren't in the right place, e.g. my old school event is on a Saturday even tho there is not a event defined on my iPad nor on my iPhone.

## Steps to Reproduce

I created a event on my iPhone that loops from Monday-Friday

Add calender to Sync with yuvomi

Start sync

## Expected Behavior

That my looped events are showing up

## Actual Behavior

They don't

## Environment

- **Browser:** Latest Safari Version at the time of writing this
- **Device:** iPad Pro M5 13-inch
- **OS:** iPadOS 26.5.2
- **Docker version:** Can't check rn
- **Yuvomi version/commit:** 1.43.0

## Logs

- none

## Additional Context

- none

ulsklyc:
Thanks for the detailed report, Leon. I could reproduce this and it was two separate bugs in how weekday-recurring events are expanded:

1. **`FREQ=DAILY;BYDAY=MO,TU,WE,TH,FR`** — Apple/iOS serializes "every weekday" this way. The recurrence engine treated `FREQ=DAILY` as a plain daily repeat and ignored the `BYDAY` filter, so the series also landed on Saturdays and Sundays.
2. **Start date not on a rule weekday** — the first instance was always emitted on the event's start date, even when that day isn't part of the `BYDAY` set (e.g. a series anchored on a weekend, or shifted there by timezone conversion). That produced the phantom Saturday occurrence you saw for your old school event.

Both are fixed in **v1.43.1** (just released), and the fix also covers ICS-subscription calendars, not just CalDAV. Please update and re-sync.

The "Saturday phantom" is definitively fixed. If your Monday–Friday event still doesn't show up at all after updating, that points at a different detail in the raw rule — in that case please paste the event's raw `VEVENT` (the `DTSTART` and `RRULE` lines, e.g. from Baikal's storage), and I'll reopen and dig in.

LEON-BROTHER:
I tried it and sadly it didn't work.
It could be that I may forgot to add a special detail. I use multiple events that assemble a week (weekends excluded) 

I could share the raw data but not publicly on github. Is there a way so that I could send it to you?

Love the work u did already btw

ulsklyc:
[...]

PR #1022, opened 2026-09-04
feat(schedule): SuperShift-inspired v3 - icons, overtime target, ICS feed, shift-start reminders, extra shifts, quick-start templates, and custom fields

mclgoerg:
## Summary

Follows up on the discussion in #1018: `ulsklyc` decided the hour-by-hour timetable need (subject,
room, instructor, A/B weeks) belongs inside the existing Schedule module rather than as a new one,
since Schedule already carries the anchor-date/cycle machinery a timetable needs. This PR is that -
plus everything else built on `feat/schedule-v3` since the v2 PR (#930) merged.

Four independent additions on top of the module, plus a custom-fields system:

### Added

- **A shift type can carry its own icon, and the module gained an overtime warning with print
  support, a personal calendar feed, and opt-in shift-start reminders.** The same icon picker used
  elsewhere in the app now applies to shift types too. A configurable weekly-hours target flags
  overtime with a warning-styled metric card on the Statistics tab and adds a Print action. A
  personal, read-only ICS feed (Settings → Personal → Feed subscriptions) lets a member subscribe to
  their own resolved schedule from any calendar app. An opt-in push reminder can fire a configurable
  number of minutes before a shift starts.

- **A member can now be on-call alongside a regular shift, and shift planning consolidates into one
  tab.** "Extra shifts" stack additively next to whatever a pattern or override already resolves for
  a day - any number can be added per person per day. The former Overrides tab is gone: Patterns,
  Overrides, and Extra shifts now live together under a renamed "Planning" tab, reached through a
  single Pattern/Override/Extra create picker. A cycle day can also carry more than one shift type now
  (several class periods in a row) - this is the part that directly answers #1018's screenshot.

- **Which shift-type quick-start templates appear is now a household preference.** An admin can hide
  the Work, School, or University quick-start button for a household that only ever uses one of them.

- **Shift types can now carry custom fields, defined once and reused across as many shift types as
  want them** - Room, Instructor, Client, whatever a household needs - deliberately not fixed columns,
  since a work shift has no room or instructor at all. A registry on the Shift types tab manages the
  fields themselves; each shift type gets a second section for attaching fields, reordering them, and
  choosing whether a value shows on the calendar overlay. A pattern day, override, or extra shift using
  a shift type with attached fields records its own value per occurrence, and a field only reaches the
  calendar overlay, the Today card, and the ICS feed if it's explicitly flagged to show there.

- **A new Overview tab compares several household members' timetables side by side** - the other half
  of #1018's ask (three kids at the same school, same period shift types, different subject/room/
  instructor each). `GET /schedule/household-members` (filtered through the existing
  `isHouseholdMember()`, so staff/guests never get offered a lane) feeds a real multi-select picker;
  a Week/Day toggle switches the visible range without ever refetching. The grid uses a collapsed hour
  scale - only hours someone in the current selection actually occupies get a row, so a 45-minute class
  period and an 8-hour work shift can share an axis without either becoming unreadable. An overnight
  shift now carries a continuation entry onto the next day's lane (previously only its pre-midnight
  half ever showed, and the tail vanished once the start day scrolled away). School and public holidays
  surface as a banner, reusing the household's existing Settings toggles rather than adding a new one.
  This went through two design revisions against live feedback during the same session - documented in
  the commit message rather than repeated here.

### Migrations (174-175 already taken by `main`; renumbered during rebase)

| # | What |
|---|---|
| 176 | `schedule_shift_types.icon` |
| 177 | `users.schedule_feed_token` + partial unique index |
| 178 | Widens `reminders.entity_type` for `schedule_entry`, adds `schedule_reminder_entries` anchor table |
| 179 | `users.schedule_weekly_hours` |
| 180 | `schedule_extra_shifts` |
| 181 | Widens `reminders.entity_type` for `schedule_extra_entry` |
| 182 | Relaxes `schedule_pattern_days` to allow several rows per cycle position (timetable-style) |
| 183 | `schedule_custom_fields`, `schedule_shift_type_fields`, `schedule_custom_field_values` |

Originally numbered 174-181 against an older `main`; renumbered during this rebase after `main`
picked up its own migration 174 (birthday name days, #993) and 175 (permission capability resources,
#996) while this branch was in flight.

### Not in scope here

*Creating or editing* school-holiday entries directly inside the scheduler (raised by @matthiasNX on
#1018) still isn't part of this PR - open to a follow-up on that separately. *Displaying* existing
school (and public) holidays is included now, as a banner on the new Overview tab, since that data
already exists and the household's own Settings toggles already gate it.

## Test plan

- [x] Full `npm test` - green (5040 tests), run after adding the Overview tab
- [x] New suites: `test-schedule-feed.js`, `test-schedule-reminders.js`, `test-schedule-extras.js`,
      `test-schedule-custom-fields.js`, `test-preferences-schedule-templates.js`; `test-schedule.js`
      extended throughout (57 tests) - household-member filtering, and pure-function coverage for
      `buildOverviewLanes()`/`computeActiveHours()`/`collapsedMinutes()`/`normalizeOverviewSelection()`
      including the overnight-continuation and free-marker-collision edge cases
- [x] i18n: every new string translated across all 22 supported non-en/de locales for real, not left
      as literal English copies - `test:i18n-translated` (main's guard against exactly that) is green
- [x] Manually verified end-to-end against a live deployment (`schedule-test.streamingsvc.com`) with
      realistic seeded data: extra shifts stacking alongside a regular shift, the unified Planning tab
      and its 3-way create picker, multi-block cycle days, custom fields end to end (registry → attach
      to shift type → per-occurrence value → calendar overlay, Today card, and ICS `DESCRIPTION`), and
      the Overview tab specifically - three household members (two timetable-style, one on an 8-day
      rotating shift pattern) compared side by side, Week/Day toggle, an overnight shift's continuation
      onto the next day's lane, and the collapsed hour scale - desktop and mobile widths
- [x] Icon picker, the overtime flag, and the ICS feed/reminder toggles were live-tested during
      development (a real mobile icon-picker scroll bug and an overtime week-boundary bug were both
      found and fixed that way) - not re-verified live again in this exact final state, flagging that
      honestly rather than claiming a fresh pass I didn't do


ulsklyc (changes requested):
Thanks for this, and for the honest test plan - flagging that the icon picker, the overtime card and
the feed toggles were not re-verified in this exact final state is exactly the right way to write one.

I read the branch at `205c1758` in its own worktree and ran the chain there: 284 of 285 links green.
The one failure is `test:changelog`, and it is branch age, not your work - the branch predates the
`v2.64.1` tag, so the guard sees a tag with no CHANGELOG heading. It clears on rebase.

Everything below is measured on the branch, not read off the description.

## Blocking

1. **Migration 176 collides with `main`.** You renumbered to 176-183 against a `main` that ended at
   175. `main` is now at 176 (note categories, #991), so your 176 (`shift type icon`) lands on an
   occupied number and the range has to move to 177-184. Nothing for you to change - I know you left
   renumbering to me on #1019 and that is the right call. Flagging the size, because you are not
   alone in the queue: #1018 also collides on 176 and needs 177-178, and #1019 collides on 175 and
   176 and needs 177-180. Whichever of the three lands first pushes the other two again. Yours moves
   eight numbers across three places each (`db.js`, the schema mirror, the `MIGRATIONS_SQL[n]`
   references in the tests), so if it helps, I may land the narrow ones first so the big block only
   travels once.

## Should fix

2. **Timed shifts go out as floating time.** `schedule-ics.js` writes `DTSTART:20260904T080000` with
   no `Z`, no `TZID` and no `VTIMEZONE`. This is not only a problem for a subscriber in another zone:
   the comment above `resolveFeedZone()` in `ics-export.js` (#818) records what actually happens -
   Google, Apple, Thunderbird, Outlook and Home Assistant place floating times on UTC in practice, so
   a household in Madrid sees its 16:00 shift at 18:00, at home, immediately. That file already has
   the fix as a reusable pair: `stampProp()` anchors a naive value to the household zone
   (`DTSTART;TZID=<zone>:…`), and `resolveFeedZone()` returns `null` when the zone is UTC-equivalent
   so the value gets a plain `Z` instead - a `VTIMEZONE` block over `UTC` is worse than no block. The
   all-day path you built (`VALUE=DATE` for a shift type with no times) is right as it is; it is only
   the timed path that needs this. Note the split there: `usesTzid()` is the recurring-series case
   (#549) and does not apply to you - `stampProp()` is the one that does, and it covers single events
   exactly like your shifts. You already import `escapeICSText` and `foldLine` from that file.

3. **Migration 178 is torn down by 182 in the same PR.** 178 creates `schedule_reminder_entries`, and
   182 drops it, rebuilds it with `pattern_day_id`, and clears `reminders` for `schedule_entry` on the
   way. None of these have ever shipped, so 178 can be built in its final shape and 182 reduced to the
   `schedule_pattern_days` rebuild alone. That saves a table rebuild and a `DELETE FROM reminders`
   that would otherwise run on every fresh install forever. Your reasoning inside 182 is sound - this
   is only about doing it once instead of twice.

## Worth a look

4. **The `isAdmin` copy in `schedule-extras.js:26` is avoidable today.** That file already imports
   `validateFieldValues`, `replaceFieldValues` and `fieldValuesFor` from `./schedule.js`, which has
   the same predicate at line 12 - so this is an export plus an import, not a new dependency. Worth
   doing because the predicate is now defined six times across the server (documents, dms,
   recipe-providers, tasks, schedule, and calendar's `isAdminUser`), and the calendar one has already
   drifted a third condition (`req.session?.isAdmin === true`). Calendar is the shape to copy, by the
   way: it lives in `calendar/helpers.js` and is imported by its callers rather than redefined.
   `docs/DECISIONS.md` entry 2 ("A rule lives in one place, not at a call site") is the rule here. I
   am consolidating all of these into one `isAdminRequest()` on 15 September, so if you would rather
   leave it, that is fine - it will get swept up either way.

5. **Router order is right today and fragile tomorrow.** `/api/v1/schedule` is registered before
   `/api/v1/schedule/feed`, `/preferences` and `/extras`. It works because `routes/schedule.js` has no
   first-segment parameter route, so the request falls through to the specific routers. The day
   someone adds `router.get('/:id')` there, all three go dark with no error. Registering the specific
   prefixes first would remove the trap.

## Checked and clear

For the record, since these are the things that usually bite: no `innerHTML`; both new `{{count}}`
keys carry `_one`; all 24 locales move by exactly the same line count (+68/-2, checked individually);
the household template preference is correctly admin-gated with a 403, matching `health_cycle_enabled`;
nothing references `schedule_pattern_days` by foreign key, so 182's `DROP TABLE` is safe; the feed
token is 32 random bytes with revocation, behind `feedLimiter`, answering 404 on an unknown token;
`schedule-preferences` needs no admin gate because it only ever writes the caller's own `users` row;
`GET /extras` without a permission check is consistent with `GET /entries` and `GET /overrides`, which
is the module's deliberate household-wide read model; and `test-schedule-reminders.js` pins
`household_timezone` to UTC, so its `realTomorrowKey()` cannot drift against `todayKey()`.

The em dashes in `docs/SPEC.md` are fine - CONTRIBUTING scopes that rule to the READMEs, CHANGELOG,
commit messages and UI strings, and `main`'s SPEC carries 536 of them. The `——` you added to `zh.json`
is correct Chinese punctuation and matches nine existing uses on `main`.

## Merge notes for me

Two conflicts against `main`, both mechanical: `package.json` (version plus the new `test:` scripts)
and `server/db.js` (the migration range). All 24 locales auto-merge. `public/sw.js` still carries
`APP_RELEASE = '2.64.0'` and needs 2.64.1 on rebase.

mclgoerg:
@ulsklyc Thanks for the thorough pass. Update on the CHANGES_REQUESTED items, all in `45a0796a`:

**Should fix**
- ICS floating time: fixed. `schedule-ics.js` now anchors timed shifts via `stampProp()`/`resolveFeedZone()` (exported them from `ics-export.js` alongside `escapeICSText`/`foldLine`), same fix as the calendar feed's `#818`. All-day path untouched. `VTIMEZONE`/`X-WR-TIMEZONE` emitted only when a timed entry actually needs the feed zone.
- Migration 178/182 double-build: fixed. 178 now creates `schedule_reminder_entries` in its final shape (`pattern_day_id` included from the start), 182 no longer drops/rebuilds it or touches `reminders` - it's back to just the `schedule_pattern_days` rebuild.

**Worth a look**
- `isAdmin` copy in `schedule-extras.js`: removed, now imports it from `schedule.js` (exported inline, same convention as `validateFieldValues`/`replaceFieldValues`/`fieldValuesFor`). Down to five copies now, not six - happy to leave the rest for your 15 September sweep.
- Router order: `/schedule/feed`, `/schedule/preferences`, `/schedule/extras` now register before the general `/schedule` router in `server/index.js`, so the trap you described can't fire even once a `/:id` route lands.

**Blocking**
- Migration renumbering: left this to you as you planned - and flagging that `main` has moved further since your review, it's at migration 180 now (not 175), so this branch's `176-183` would need `181-188`, not `177-184`. Didn't touch it since a correct fix means rebasing onto current `main`, not just relabeling 8 numbers, and you mentioned wanting to sequence this against #1018/#1019 anyway.

Also worth flagging directly: the Overview tab from #1018 (side-by-side timetables, the other half of that ask) landed in this PR too, in `fdfbbed9` - multi-select picker through your `isHouseholdMember()` route, Week/Day toggle, collapsed hour scale, overnight-shift continuation, and the school/public holiday banner. Full `npm test` (5040+ tests) is green.

Full `npm test` green in a fresh container after all of the above.

ulsklyc:
Thanks - I read the branch at `45a0796a` in its own worktree and checked all four items at the
code rather than off the description. All four hold: `stampProp()`/`resolveFeedZone()` are wired
into the timed path with the all-day path untouched, migration 178 now builds
`schedule_reminder_entries` in its final shape and 182 is back to the `schedule_pattern_days`
rebuild alone, `schedule-extras.js` imports `isAdmin`, and the three specific `/schedule/*`
prefixes register first.

## One new blocking item: `test:schedule-feed` is red, and the container run cannot see it

The ICS fix outran two tests that pin the old floating format. Three runs of the same command,
same worktree, only `TZ` changed:

```
TZ=UTC              npm run test:schedule-feed   ->  pass 17, fail 0
TZ=Europe/Berlin    npm run test:schedule-feed   ->  pass 13, fail 4
TZ=America/New_York npm run test:schedule-feed   ->  pass 13, fail 4
```

Counter-check at your previous head `205c1758`, same machine, same command: green. So it comes
from `45a0796a`.

Your "green in a fresh container" is true, and that is the interesting part. The suite never sets
`household_timezone`, so `householdTimeZone()` falls through to `serverTimeZone()`, which is the
machine's zone. In a UTC container `resolveFeedZone()` returns `null`, `stampProp()` appends a
`Z`, and because `/DTSTART:\d{8}T140000/` is not anchored at the end it still matches
`DTSTART:20260905T140000Z`. So CI stays green, has never actually tested the anchoring, and will
not catch a regression in either direction. Outside UTC the same assertions see
`DTSTART;TZID=Europe/Berlin:20260905T140000` and fail.

Three assertions, two tests: `test/test-schedule-feed.js:154`, `:155` and `:168`. The three
follow-on failures are collateral only - the first test throws before its
`db.exec('DELETE FROM schedule_overrides')`, so every later `insertOverride` hits the UNIQUE.

The fix that makes the change actually provable: pin `household_timezone` in the suite and assert
both shapes - `DTSTART;TZID=<zone>:` for a non-UTC household, and the plain `Z` form for a UTC
one. That is the pattern `test-schedule-reminders.js` already follows by pinning the zone to UTC.

## The Overview tab

I reviewed `fdfbbed9` too, since it was not in the round I looked at. It is careful work:
`isHouseholdMember()` reused rather than rewritten, real translations in all 24 locales (not
placeholders), no `innerHTML`, no raw colours in the CSS, and pure-function tests that feed real
days in rather than asserting that a function name exists in the source. `test:frontend-audit`,
`test:i18n`, `test:i18n-translated`, `test:lucide-icons`, `test:openapi-coverage` and
`test:schedule` are all green on the branch.

Two measured holes, both from the same root: `refreshOverview()` fetches exactly the visible week,
while the continuation logic reaches one day back.

```
Week view           : activeHours=[0,1,2,3,4,5,22,23] -> top=0px height=336px  (correct)
Day view, 12 Sept   : activeHours=[6..19]             -> top=0px height=18px   (6 h would be 336px)
Week boundary, Mon  : continuation on Mon 07 Sept: none
```

1. **Day view.** `activeHours` is computed from entries filtered by `visibleDateKeys`, but an
   overnight shift carries the *previous* day as its `date_key`. Viewing the day after one, it
   contributes no hour, the axis falls back to the 06-19 band, and the continuation collapses to
   the 18px minimum height at the top of the lane.
2. **Week boundary.** A night shift on the last day of the previous week is not in the fetch
   window at all, so the first visible day shows no continuation - the same vanishing tail the
   feature was built to fix, moved to the week edge. Fetching `from - 1` covers both.

Smaller: the comment above `buildOverviewLanes()` says lane order follows "AUSWAHLREIHENFOLGE",
but `getSelectedUserIds()` returns DOM order, so lanes are alphabetical by `display_name`. The
behaviour is the better one (it is stable across selections); it is the comment that says
something else.

And the tab appears in neither `CHANGELOG.md` nor `docs/SPEC.md`, while the rest of the PR keeps
both carefully (+243/-19 in SPEC, six bullets under the new `### Added`). It matters more than
usual here because this tab is the public answer to #1018 and #670.

## Renumbering

Confirmed at the source rather than from the PR body: `main` ends at 180
(`grep -nE "^\s*version: [0-9]+," server/db.js | tail -1`), so this branch's 176-183 needs
181-188. Still mine to do, as you left it.

One coordination note so it does not surprise you: #1018 was rebuilt into the Schedule module on
4 September and now carries its own answer to the same problem - migration 176 rebuilds
`schedule_pattern_days` the same way your 182 does, and it adds `subject`/`room`/`instructor` as
fixed columns where you use the custom-field registry. A trial merge of the two heads conflicts in
30 files, so they are two implementations of one feature rather than two halves. I am sorting that
out with its author in that thread; nothing for you to do, and it does not change that yours is
the base.

mclgoerg:
@ulsklyc Thanks for going test-by-test and catching a real gap - all in `cf56d60d`.

**Blocking: `test:schedule-feed` false-pass in UTC**
Confirmed your read exactly: `household_timezone` was never pinned, so `householdTimeZone()` fell through to the container's own zone, and two assertions were unanchored (`/DTSTART:\d{8}T140000/` with no `$`), so a UTC container's `...140000Z` still matched by substring. Fixed:
- Pinned `household_timezone` to `'UTC'` at module scope (same as `test-schedule-reminders.js`).
- Anchored the three flagged assertions (`$`, multiline) so they can no longer match the wrong shape by accident.
- Added a new test that pins the household to `Europe/Berlin` and asserts the actual `DTSTART;TZID=Europe/Berlin:...` output - so both shapes are now genuinely proven, not just no-longer-false-passing.

Reran your exact matrix (`TZ=UTC / Europe/Berlin / America/New_York`) - 18/18 pass identically under all three now.

**Overview tab - two measured holes, same root cause**
Both were exactly what you diagnosed: `refreshOverview()` fetched precisely the visible week, so a night shift on the previous day (day view) or the previous week's last day (week boundary) was either filtered out of the active-hours calculation or never fetched at all.
- The fetch window now reaches one day before the visible range.
- Added `touchesVisibleDay()`: an entry counts toward `computeActiveHours()` if it's on a visible day, or it's an overnight shift whose continuation lands on one.
- Reproduced your own numbers to verify: day view of 12 Sept and the first day of the week after a night shift on the 6th/11th both now measure `top:0px height:336px`, not the 18px collapse.

**The `AUSWAHLREIHENFOLGE` comment**
Fixed - it now says what actually happens (`getSelectedUserIds()`'s DOM/alphabetical order) and why that's the better behavior (stable across selection changes), instead of claiming click order.

**CHANGELOG.md / docs/SPEC.md**
Added an Overview tab entry to both - a bullet under `### Added` in the CHANGELOG, and a `**Overview tab (Schedule v3):**` block in SPEC.md alongside the other Schedule v3 sub-features (person picker via `isHouseholdMember()`, fixed lane order, overnight continuation, the holiday banner, no new endpoint/migration).

**Migration renumbering**
Left untouched, as before - yours at merge time.

**#1018 coordination**
Noted, nothing needed from me there.

New tests: `touchesVisibleDay()` directly, plus a `computeActiveHours()` case reproducing the day-view collapse. Full `npm test` green (5043 tests).

ulsklyc (changes requested):
Read `ae7e8023` in a worktree, ran the suites, and counter-proved the parts that a green suite alone cannot vouch for. **All three findings from round two are closed**, and the way they are closed is better than what I asked for. One new thing needs doing before this merges, and it is small.

## The timezone finding is fixed, and the fix is anchored

You pinned `household_timezone` to `'UTC'` at module scope (`test-schedule-feed.js:41`), anchored the three assertions with `$` and `/m`, added `assert.doesNotMatch(ics, /TZID=/)` as a second latch, and wrote a new test that pins the household to `Europe/Berlin` and asserts the `DTSTART;TZID=Europe/Berlin:...` shape positively (`:192`). `withHouseholdTimeZone()` restores UTC in a `finally`.

Measured on your branch: **18 pass / 0 fail under `TZ=UTC`, `Europe/Berlin`, `America/New_York` and `Pacific/Auckland`** - the same number in all four, which is the point. Then the counter-proof: I made `resolveFeedZone()` return `null` unconditionally, so the TZID form can never be produced. **14 pass / 4 fail, identically in UTC and in Berlin.** The suite now fails for the right reason regardless of the machine it runs on, which is exactly what it could not do before.

## Both holes in the Overview tab are closed

Two halves, and I checked each against the code rather than the description:

- `entriesFrom = addLocalDays(from, -1)` (`schedule.js:421`) fetches the day before the visible range, so a night shift starting on the last day of the previous week has something to continue from.
- `touchesVisibleDay()` (`:955`) lets that entry count towards `computeActiveHours()` (`:1157`), so the single-day view right after a night shift no longer falls back to the default band.

What I verified around them: `buildOverviewLanes()` ends on `days.map(...)` (`:996`), so the extra day can never become a column; all three readers of `overview.entries` are accounted for; and the server cap is `MAX_RANGE_DAYS = 731` (`routes/schedule.js:87`), so eight days is nowhere near it.

The tests for `touchesVisibleDay()` (`test-schedule.js:816`) are the right shape - two positive cases and two negatives, including "the continuation does not reach two days out", which is the assertion that stops the helper from quietly becoming a range check.

## CHANGELOG and SPEC are there

Both, and the SPEC entry documents the widened fetch window itself rather than just the feature. That is the sentence that will save the next person the reconstruction I just did.

## The part you were not asked for

`renderMonthDay()` carried a **second copy** of the schedule filter (`state.scheduleEntries.filter(...)` inline), which is why matthiasNX saw every member's shifts with a person filter active. You did not patch the copy - you deleted it, so all four views now go through `scheduleEntriesOnDay()` (`calendar.js:1955`, `:2019`), and that one function calls `passesPersonFilters()` through a three-line adapter (`:2015`) instead of growing a second person-filter implementation for Schedule.

And it is covered by a **real behavioural test** (`test-calendar.js:1322`): it sets the module state, calls the actual function, checks no-filter, person-filter and "assigned to me", and restores everything in a `finally`. That is the kind of test I have been asking two other PRs for this week.

## One thing to add before merge: `entriesFrom` has no test

The counter-proof: I set `entriesFrom = from` on your branch, which makes the week-boundary half of the fix inert and puts the original bug straight back. `test:schedule` 59/0, `test:schedule-feed` 18/0, `test:calendar` 108/0. **All green.** `touchesVisibleDay()` is well covered; the fetch window that feeds it is not covered at all.

It is buildable, and I built it to be sure before asking. Extract the four lines at the top of `refreshOverview()` (`:409`) as a pure function - the same move your own `__test` comment describes for the other helpers:

```js
function overviewFetchRange(weekCursor, weekStartPref) {
  const weekStart = weekStartIndex(weekStartPref);
  const from = startOfLocalWeekKey(weekCursor, weekStart);
  return { entriesFrom: addLocalDays(from, -1), from, to: addLocalDays(from, 6) };
}
```

with a test asserting `entriesFrom` is the day before `from`. On your branch that gives 60/0, and with `entriesFrom: from` it gives 59/1 - so it fails for the reason it exists.

## Two notes, neither blocking

**`test:changelog` fails on this branch, and it is not your doing.** "Getaggte Releases ohne CHANGELOG-Eintrag: 2.64.1" - the branch sits on `2.64.0` and predates the v2.64.1 tag. It disappears on rebase. Worth knowing because the `&&` chain stops there: `npm test` on this branch alone gets through 104 suites and 1570 tests and then halts. I ran the remaining 175 suites separately: **3766 tests, 0 failures.** So everything is green except that one release artefact.

**`test-schedule-feed.js:41` uses a plain `INSERT`.** `sync_config.key` is a PRIMARY KEY, so if a future migration ever seeds `household_timezone`, that line throws at import time rather than failing a test. `INSERT OR REPLACE` costs nothing.

## Merge logistics, since I did the dry run

Both of your commits cherry-pick onto `chore/schedule-v3-renumber` **without conflicts** - they do not touch `server/db.js`, so the 181-188 renumbering stands and your authorship is preserved. Merging `main` after that produces exactly **one** conflict, in `package.json`: the long `test` chain. The resolution is the union, and there is a trap in it - 278 suites are common, 5 are yours (`test:schedule-feed`, `test:schedule-reminders`, `test:schedule-extras`, `test:schedule-custom-fields`, `test:preferences-schedule-templates`), and **one is only on main**: `test:document-folder-delete`, from #1016. Taking "our" side drops that suite out of the chain silently. I will handle this side; noting it so we both know what it looks like.

Add the `entriesFrom` test and this is done. Thank you for the round: three findings closed, one of them with a better test than I described, and a fourth bug fixed that came from someone else's thread.

mclgoerg:
Addressed in `ad51695b`: `overviewFetchRange(weekCursor, weekStartPref)` extracted from `refreshOverview()`, returning `{ entriesFrom, from, to }`, added to the `__test` export, with the new behavioral regression test in `test-schedule.js`. Applied the `INSERT OR REPLACE` fixture hardening in `test-schedule-feed.js` too.

Counter-proof: with `entriesFrom` reverted to equal `from`, `test:schedule` goes to **59 pass / 1 fail** - only the new test fails, for the reason it exists. Restored, it's back to green.

Test results on the branch:
- `test:schedule`: 60/60
- `test:schedule-feed`: 18/18
- Full `npm test`: all suites green, including `test:changelog` this time (no tag drift on this run).

Ready for another look whenever convenient.

ulsklyc:
Merged as `3490a179`. GitHub will not mark this PR as merged, because it did not go in as `ad51695b` - the eight migrations had to be renumbered, so the commits reached `main` with new hashes. **Your authorship is intact on every one of them**, and the eight migrations landed on **182-189** with no gap.

## What changed on the way in, and why

**Renumbering.** Your branch carried 181-188. While it was open, `main` took migration **181** (the budget account fix from #1034), so yours moved up one. That touched `server/db.js`, the schema mirror, `MIGRATIONS_SQL` in the test, and 31 by-number references across 12 files. Measured after: 189 migrations, no gap, no duplicate.

**One conflict in `public/pages/calendar.js`**, and it was worth reading rather than resolving mechanically. Your `scheduleAsAssignable()` adapter had to stay - `scheduleEntriesOnDay()` calls it - but taking the whole branch side would have left the `aria-label` comment from #1037 in the file twice. Resolved to the adapter only.

## One thing I fixed rather than sending back to you

A rebase had put a second `### Added` and `### Fixed` block **inside the already published `[2.63.0]` section**, carrying seven Schedule v3 entries. Neither obvious response was safe:

- Deleting the block would have destroyed **five** feature descriptions - the icon/overtime/ICS work, extra shifts, the template preference, custom fields and the Overview tab existed *only* there.
- Leaving it would have added seven entries to a release that shipped on 2 September without them.

So the five unique entries moved to `[Unreleased]` beside your other Schedule work, the two duplicates went, and the stray blocks are gone. All seven texts now appear exactly once, and `[2.63.0]` is back to one Added, one Changed and one Fixed. Worth knowing for your next rebase - it is the kind of damage that reads as harmless until it reaches a published section.

## Numbers

Full chain on the merge result: **exit 0, 285 suite links, 6169 assertions, 0 failures** - five suites more than `main` had, all of them yours (`schedule-feed`, `schedule-reminders`, `schedule-extras`, `schedule-custom-fields`, `preferences-schedule-templates`).

## Credit where it was missing

@matthiasNX withdrew #1018 in favour of this PR and found, while testing it, that the calendar's person filter never reached Schedule entries - the bug you then fixed in `ae7e8023`. The changelog named the issue number but not him; that is corrected now.

Thank you for four review rounds. Every one of them made the thing measurably harder to break, and the counter-proofs you ran on your own findings are the reason this went in without a second look.

PR #1027, opened 2026-09-05
Make the countdown tile's overdue grace period configurable (#969)

mclgoerg:
## Summary

Closes/addresses #969. The dashboard countdown tile always kept an expired
countdown visible for a fixed 7 days before dropping it. Since the tile sorts
ascending by `days_until`, overdue items (negative values) sort to the *top* -
a full week of grace pushed genuinely upcoming dates out of the few visible
rows. @ulsklyc agreed on the discussion that this should be a household
setting rather than a shorter hardcoded default, so existing installs see no
behavior change (default stays 7).

- `server/services/countdowns.js`: `OVERDUE_GRACE_DAYS` becomes
  `DEFAULT_OVERDUE_GRACE_DAYS`, resolved per-household from `sync_config` via
  a new `overdueGraceDays(d)` helper (same shape as the existing
  `disabledModules(d)`). `0` is treated as a valid, deliberately-set value,
  not "unset".
- `server/routes/preferences.js`: new admin-gated `countdown_grace_days`
  preference (integer, 0-90), same GET/PUT/validation shape as the existing
  `tasks_default_points`.
- New settings leaf at **Settings -> Modules -> Key dates**
  (`public/settings/pages/modules-countdowns.js`), registered in
  `registry.js`. This is its own leaf rather than a new field in
  `modules-options.js` - that file has its own structural guard
  (`test:frontend-audit`) requiring it stay pure toggle-primitive with zero
  raw `<input>` literals, so a numeric field belongs elsewhere.
- Locale keys added to all 24 locales (`de.json`/`en.json` as reference,
  translated - not copied - into the rest, verified via
  `test:i18n-translated`).

## Test plan

- New `test/test-preferences-countdown-grace.js`: GET default (7), PUT
  round-trip, `0` is respected as a real value, 0-90 bounds validation,
  admin-gating (403 for non-admin).
- `test/test-countdown.js`: added cases proving a shortened grace period
  (e.g. 3 days) actually changes what `getCountdowns()` returns for both
  event- and task-sourced entries, and that `0` means "no grace" rather than
  silently falling back to 7. Existing hardcoded-7 assertions are unaffected
  (unset `sync_config` still resolves to the default).
- `test/test-settings-navigation.js`: updated the `modules` domain leaf count
  (5 -> 6) for the new leaf.
- Manually verified in a real browser against a fresh throwaway instance:
  default renders as 7, an out-of-range value (999) is rejected client-side
  with the right message, a valid save (3) persists across a full page
  reload, and saving `0` persists as `0` (not silently reset to 7).
- Full `npm test` is green (5100+ tests, 0 failures).


ulsklyc (changes requested):
Read the branch at `6cc33542` in a worktree and ran the suites rather than reviewing the description. The shape is right and the discipline around it is right: this follows `tasks_default_points` down to the admin gate, it adds the new suite in all three places that a new suite needs (file, `test:` script, and the `npm test` chain) plus `docs/test-suites.md`, and it carries the `public/sw.js` precache entry for the new settings leaf. That last one is the guard that has turned CI red here before, and you did not have to be told.

Suites, all green on the branch: `test:countdown` 41, the new `test:preferences-countdown-grace` 8, `test:settings-navigation` 83, `test:frontend-audit` 359, `test:i18n` 53, `test:i18n-plural` 14, `test:i18n-translated` 92, `test:sw-precache` 7, `test:suite-chain` 4, `test:changelog` 28.

The two behaviour tests in `test-countdown.js` are the right two: a shortened period actually drops what the default would still show, on both the event and the task branch, and `0` is proven to mean "no grace" rather than "fall back to 7". The locale values are real translations rather than copies, and I checked `{{max}}` in all 24 files: present in every one.

Two things to fix before this goes in.

## 1. An empty field saves `0`, and `0` is destructive here

`public/settings/pages/modules-countdowns.js:50`:

```js
const next = Math.trunc(Number(input.value));
if (!Number.isFinite(next) || next < 0 || next > MAX_COUNTDOWN_GRACE_DAYS) { ... }
```

`Number('')` is `0`, so `Math.trunc(Number(''))` is `0`, and `0` passes every check on both sides: `Number.isFinite(0)` is true, the range check is satisfied, and the server accepts it as a deliberate value by design. The form carries `novalidate` (line 15) and the input has no `required`, so an empty field is a perfectly reachable state: clear the number to retype it, hit Save before typing, and the household's grace period is now zero.

The consequence is not symmetric with the field this was copied from. In `modules-rewards.js` the same expression sits under a default of `0`, where an accidental zero changes nothing anyone can see. Here `0` means every overdue countdown disappears from the tile at once - exactly the behaviour your own test at `test-countdown.js:243` proves - and nothing told the user they asked for it.

Suggested fix: treat an empty field as invalid rather than as zero, before the numeric parse. Something like `if (input.value.trim() === '')` showing `countdownGraceDaysInvalid`. Deliberate zero stays reachable by typing `0`, which is the case you built the feature around.

## 2. The default `7` now exists in three places, and nothing binds them

- `server/services/countdowns.js:62` - `DEFAULT_OVERDUE_GRACE_DAYS`
- `server/routes/preferences.js:126` - `DEFAULT_COUNTDOWN_GRACE_DAYS`, with a comment saying it matches the first one
- `public/settings/pages/modules-countdowns.js:6` - `DEFAULT_GRACE_DAYS`

A comment that says "this matches the other one" is the thing that stops being true. The client copy is unavoidable without a bundler, and the comment there is the right way to carry it. The two server-side copies are not: `preferences.js` already imports from `services/` (holidays, birthdays, module-capabilities) and `services/countdowns.js` imports nothing from `routes/`, so exporting `DEFAULT_OVERDUE_GRACE_DAYS` and importing it is the same direction the file already uses. No cycle.

`docs/DECISIONS.md` entry 2 puts it better than I can: "A copy is not a shortcut. It is a second answer, waiting to diverge from the first." That entry is written about access rules rather than constants, but the failure mode is the same one, and it is silent.

## One note, not a request

`overdueGraceDays()` (`countdowns.js:218`) accepts any integer `>= 0` from `sync_config`, while `parseCountdownGraceDays()` (`preferences.js:155`) clamps reads to `MAX_COUNTDOWN_GRACE_DAYS`. Today that cannot diverge, since the route is the only writer and it rejects anything above 90. It would diverge the day the cap is lowered: the settings page would show 30 while the tile kept computing with 90. Worth a thought when you touch the constants for point 2, not worth a separate change.

## Merge logistics

The branch conflicts with `main` in `CHANGELOG.md` and in the `test` chain line of `package.json` - both are additions next to other additions, no logic. A rebase resolves them.

This is a settings-and-dashboard change, so it rides the Tuesday interface train rather than merging on the day it is ready.

mclgoerg:
Addressed in `5ea5f5f9`:

1. **Blank field no longer saves 0.** `parseGraceDaysInput()` in `modules-countdowns.js` now rejects an empty/whitespace-only value before the numeric parse, so `Number('')` can no longer slip through as a deliberate 0. An explicitly typed `0` still round-trips as intended. Exported the pure parser and added `test-settings-navigation.js` coverage: blank/whitespace rejected, `'0'` accepted, in-range accepted, out-of-range/non-numeric still rejected.
2. **One default, one source.** `DEFAULT_OVERDUE_GRACE_DAYS` is now exported from `services/countdowns.js` and imported into `preferences.js`; the second copy (`DEFAULT_COUNTDOWN_GRACE_DAYS`) is gone. The client-side copy in `modules-countdowns.js` stays, per your note - there's no bundler to share it through.

Test results on the branch:
- `test:settings-navigation`: 84/84 (new parser test included)
- `test:preferences-countdown-grace`: 8/8
- `test:countdown`: 41/41
- `test:frontend-audit`: 359/359, `test:i18n`: 53/53, `test:i18n-translated`: 92/92, `test:sw-precache`: 7/7

Manual browser verification on a throwaway instance: cleared the field and submitted - inline error shown, no save, reload still showed the previous value (7). Typed `0` and saved - reload showed `0` persisted deliberately. A normal value (3) and an out-of-range value (91, rejected) were checked too. 7/7 scripted checks passed.

Ready whenever you get to it - and understood on the Tuesday interface train timing.

ulsklyc (approved):
Approving. **And an apology first: you reported both fixes on 6 September at 09:32 and I left you waiting for a day.** I had verified them the same morning and then never said so, which is the one thing a contributor cannot work around.

## Both should-fixes are closed, measured rather than read

**Blank field no longer saves 0.** `parseGraceDaysInput()` rejects empty and whitespace-only input *before* the numeric parse, so `Number('')` can no longer arrive as a deliberate zero, and a typed `0` still round-trips.

```
untouched                     84 pass / 0 fail
blank-rejection line removed  83 pass / 1 fail  <- exactly your new test
```

**One default, one source.** `DEFAULT_COUNTDOWN_GRACE_DAYS` is gone from `preferences.js`; the server imports `DEFAULT_OVERDUE_GRACE_DAYS` from `services/countdowns.js`, and that file is now the only place the 7 is decided on the server.

## The nice-to-have from my first review, unchanged

Two `7`s remain - `services/countdowns.js:62` and the client copy in `modules-countdowns.js` - and **nothing compares them**. The client copy is right, since there is no bundler to share a constant through; what is missing is the seam that notices when they drift. The repo already has that shape: `test-frontend-audit.js` reads `CONFIRM_GRACE_MS` out of `bulk-pill.js` by regex and asserts it against its twin. Not a blocker, and not something I would hold the PR for.

## The rebase got considerably more expensive while this sat

This is the part I owe you, because the delay is mine. When I reviewed on the 6th, you had **two** conflicts. Today you have **27**, because five PRs landed on `main`: #1040, #1038, #1037, #1028 and #1022 (Schedule v3, with 66 new locale keys across all 24 files).

The good news is that 25 of them are mechanical:

- **24 locale files, one conflict each** - your `countdownGraceDays*` block and Schedule's `scheduleTemplates*` block were both appended to `settings`. Union, both blocks. Do not reserialise the files; they carry 4-space indentation and nesting that a `JSON.parse`/`stringify` round trip destroys.
- **CHANGELOG** - union.

The other two need care, and they fail *silently* rather than loudly:

**`server/routes/preferences.js`** - one line, the long destructuring of `req.body`. Your side has `countdown_grace_days`, `main` has `schedule_hidden_templates`. Taking either side whole drops the other field, and a dropped field is never an error: it is simply `undefined` forever, so `if (schedule_hidden_templates !== undefined)` at `:894` never fires and saving quick-start templates stops working with no exception anywhere. **The union is both names.**

**`package.json`** - the same trap one level up:

```
main: 293 test: scripts   your branch: 287   the union must be 294
only on main: 7  (the five schedule-*, plus budget-account-inherit-migration and document-folder-delete)
only in yours: 1  (test:preferences-countdown-grace)
```

These two compound each other, which is why I would resolve them in one sitting: pick "ours" in `package.json` and you lose `test:preferences-countdown-grace`; pick "theirs" in `preferences.js` and you lose `countdown_grace_days`. Do both and the feature is dead *and* the test that would have caught it is no longer in the chain. Each field is well covered on its own (17 assertions for the schedule one, 9 for yours) - the coverage only helps if both survive the merge.

`test:suite-chain` catches a script that exists but hangs in no chain. It cannot catch one that was dropped from both sides.

## Verified

`test:settings-navigation` 84/84 on `5ea5f5f9`, counter-proof anchored on the test that exists for it. Sorry again for the silence - the work was done, the message was not.

mclgoerg:
Rebased onto current `upstream/main` (now includes Schedule v3 / PR #1022) — new tip is `7b764956`, no merge conflicts reported.

**Conflict resolution (27 files touched):**
- `CHANGELOG.md`: both this PR's and Schedule v3's entries kept under `## [Unreleased]`.
- `package.json`: union of both test-chain additions — `test:preferences-schedule-templates` (v3) and `test:preferences-countdown-grace` (this PR) both present in the aggregate `test` chain, verified programmatically against the full `test:*` key list.
- 24 `public/locales/*.json`: each had one conflict between adjacent `scheduleTemplates*`/`countdownGraceDays*` sibling keys, resolved as a targeted splice rather than a full reserialization. Verified every file round-trips canonically and has full keyset parity with `de.json`.
- `server/routes/preferences.js`: merged destructuring keeps both `schedule_hidden_templates` and `countdown_grace_days`; both GET/PUT paths confirmed intact.

**Tests:** focused suites (657 tests: `test:suite-chain`, `test:preferences-countdown-grace`, `test:countdown`, `test:settings-navigation`, `test:preferences-schedule-templates`, `test:frontend-audit`, `test:i18n`, `test:i18n-translated`, `test:sw-precache`) all green. Full `npm test`: 5580 tests, 5578 pass, 0 fail, 2 pre-existing skips.

**Smoke test:** a PUT setting both `countdown_grace_days: 0` and `schedule_hidden_templates` together round-trips correctly on GET (proves the merged destructuring keeps both fields live), and a schedule-only PUT/GET round-trips independently.

ulsklyc:
**Merged**, `0d24d6f8` on `main`.

Your rebase description was accurate, and I checked the three things that could have failed silently rather than taking it on trust. All three hold:

```
package.json      294 test: scripts, nothing from main lost,
                  test:preferences-countdown-grace present          -> the union
preferences.js    both schedule_hidden_templates and countdown_grace_days
                  in the one req.body destructuring, adjacent       -> no dropped field
24 locales        +168 / -0 lines, exactly 7 per file, all 24 parse,
                  both key blocks present, 4-space indent intact    -> a splice, not a reserialise
```

The `-0` is the number I was most glad to see. A `JSON.parse`/`stringify` round trip would have shown up as a few hundred deleted lines per file, and it did not happen once.

Full chain on your branch: **exit 0 read from the log file, 286 suite links, 5580 + 694 assertions, 0 failures.**

Thank you for the turnaround, and sorry again that the rebase grew from two conflicts to twenty-seven while I sat on the review. You did the expensive half of that.

PR #1028, opened 2026-09-05
feat(calendar): let the phone month view show titles instead of dots

ter9ng:
Below 640px the month view reduces every entry in a day cell to a 10px dot. That is a reliable
"something is happening" signal, and it is also why the month has to be read one day at a time:
titles only exist in the agenda and day views.

This adds an "Event titles" switch under Display in the calendar's filter sheet. When it is on,
the month cell keeps the same chips and shrinks them instead: `--text-2xs`, one line with an
ellipsis, at most four per day, then the "+N" overflow the cell already had.

The switch is per device (localStorage, next to the schedule display mode). The question it
answers, "does a title fit on this screen", belongs to the device, and the same person reads the
same calendar on a monitor in the evening. It only appears below 640px, because above that the
month cell already shows titles and the toggle would have nothing to do.

Three things this deliberately does not change:

- **The default.** Dots stay on. Their contrast recipe is measured (the ring exists because five
  of nine seed colours failed WCAG 1.4.11 as a bare fill), and an update should not rebuild every
  household's month view unasked.
- **The recipe itself.** The dot rules are scoped with `:not(.month-view--titles)` rather than
  overridden, so the titles variant inherits the base chip rule and the stylesheet carries the
  recipe once.
- **The tap target.** A 13px title row is as far below the 44px minimum as the 10px dot was, so
  the whole cell remains a single drill-in target on mobile. The comment that used to justify
  this with the chip's shape now names the real reason.

The per-day cap is a custom property on `.month-grid`, read once per fit pass. It is not a second
`matchMedia` beside the media query that already owns the breakpoint. The calendar has paid for
that duplication once already (see `MOBILE_MEDIA_QUERY`).

Drafted with Claude Code, reviewed and tested by me.

## Changes

- `public/pages/calendar.js`: `monthTitles` state persisted in localStorage, `month-view--
  modifier class via `monthViewClasses()`, cap read from `--month-day-max-visible` via
  `monthDayVisibleCap()`, and the cap enforced on both the fits-all and the clip path of the fit
  pass. Both helpers are exported on `__test`.
- `public/styles/calendar.css`: titles variant inside the existing 639px query; dot rules scoped
  with `:not(.month-view--titles)`; cap property on `.month-grid`.
- `public/locales/*.json`: `calendar.toggleMonthTitles` in all 24 languages.
- `test/test-calendar.js`: cases for `monthViewClasses`, `monthDayVisibleCap`, and the cap
  applying without "+N" being skipped when the cell is tall enough.
- `CHANGELOG.md`: entry under Unreleased.

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **The month view on a phone can now show event titles instead of coloured dots.** Below 640px
+  every entry in a month cell was reduced to a 10px dot - a reliable "something is happening"
+  signal, but one that makes reading the month itself a day-by-day affair. A new "Event titles"
+  switch under Display in the calendar's filter sheet keeps the same chips and shrinks them to
+  10px single-line rows with an ellipsis, up to four per day followed by the existing "+N".
+
+  The dots stay the default: their contrast recipe is measured, and an update should not rebuild
+  every household's month view unasked. The switch is per device (localStorage, next to the
+  schedule display mode), because the question it answers - does a title fit on this screen - is
+  the device's, and the same person reads the same calendar on a monitor in the evening. It
+  appears only below 640px, where there is a second fassung to choose. Colour, contrast and the
+  tap target are unchanged: the titles inherit the tinted surface the wider month view already
+  uses, and a tap anywhere in the cell still opens the day.
+
 - **Deleting a document folder can now either keep its documents unfiled or delete the confirmed
   subtree together with its documents.** The dialog previews exact folder and document counts and
   offers destructive deletion only when the user may delete every affected document. The server
--- a/public/locales/ar.json
+++ b/public/locales/ar.json
@@ -903,6 +903,7 @@
         "toggleHolidays": "العطلات الرسمية",
         "toggleSchool": "العطلات المدرسية",
         "toggleBirthdays": "أعياد الميلاد",
+        "toggleMonthTitles": "عناوين المواعيد",
         "assignedToMe": "المُسندة إليّ",
         "deleteRecurringTitle": "حذف الحدث المتكرر",
         "recurringScopeLabel": "ينطبق على",
--- a/public/locales/cs.json
+++ b/public/locales/cs.json
@@ -903,6 +903,7 @@
         "toggleHolidays": "Státní svátky",
         "toggleSchool": "Školní prázdniny",
         "toggleBirthdays": "Narozeniny",
+        "toggleMonthTitles": "Názvy událostí",
         "assignedToMe": "Přiřazeno mně",
         "deleteRecurringTitle": "Odstranit opakovanou událost",
         "recurringScopeLabel": "Platí pro",
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -915,6 +915,7 @@
         "toggleHolidays": "Feiertage",
         "toggleSchool": "Schulferien",
         "toggleBirthdays": "Geburtstage",
+        "toggleMonthTitles": "Termintitel",
         "assignedToMe": "Mir zugewiesen",
         "deleteRecurringTitle": "Serientermin löschen",
         "recurringScopeLabel": "Gilt für",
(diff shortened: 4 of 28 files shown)


ulsklyc (approved):
Read `c8f42483` in a worktree, ran the suites, and checked the three claims in the description that a reviewer cannot take on trust. All three hold. This is careful work, and the care is in the places where it is expensive rather than the places where it shows.

## What I verified rather than assumed

**The specificity trap, and the guard for it.** You noticed that scoping the dot rules with `:not(.month-view--titles)` puts them at 0,3,0 (`calendar.css:1813`), which ties with the old three-class `.is-clipped` rule, and that a tie goes to whichever comes later in the file - which would have been the dot rule, bringing every clipped chip back while "+N" kept counting it as hidden. Raising the clip rule to four classes (`calendar.css:432`) is the right fix, and you covered it.

I ran the counter-proof: reverted the clip rule to its old three-class form on top of your branch and ran `test:calendar`. It fails, and the failing output names the reason rather than just reporting a mismatch:

```
✗ Der Klipp-Guard steht ueber jeder Fassungsregel, die display setzt: .is-clipped
  traegt 3 Klassen, .month-view:not(...) ... traegt 3 - bei Gleichstand gewinnt die
  spaetere Regel, und das ist die Fassung
```

That is a test that will still explain itself to whoever trips over it in a year.

**"One line with an ellipsis" is true, and not by accident.** It works because the rule at `calendar.css:369` already gives the chip's inner `span` all three parts: `min-width: 0`, `overflow: hidden` and `text-overflow: ellipsis`. The `min-width: 0` is the load-bearing one - without it the span is a flex item stuck at its content width, and the chip's own `overflow: hidden` would cut the title off flat with no ellipsis at all. Inheriting the base rules instead of restating them is what keeps that intact at 10px.

**`monthDayVisibleCap` covers the three cases that mean the same thing.** `parseInt('')` is `NaN`, `NaN > 0` is false, and `'0'` gives `0`, so the empty property, the base value on `.month-grid` (`calendar.css:215`) and anything unreadable all resolve to `Infinity`. Worth the separate function, and worth `__test`.

**The height reasoning checks out.** `.month-grid` really does carry `overflow-y: auto` and `grid-auto-rows: 1fr` (`calendar.css:219-222`), so six rows of 108px scroll rather than squeeze, and the cap handles the other end on a tall device. `--text-2xs` exists in `tokens.css:1006` at 10px and is used by seven other stylesheets, so this is not a new size being invented at the edge.

Suites on the branch, all green: `test:calendar` 111 (106 on `main`, so five new), `test:frontend-audit` 359, `test:typography` 17, `test:i18n` 53, `test:i18n-plural` 14, `test:i18n-translated` 92, `test:sw-precache` 7, `test:changelog` 28.

## Two things I was going to raise and then withdrew

Both are here because you might otherwise get them from the next reviewer.

**The reset button leaves the switch on.** `#cal-filters-reset` restores the layers, "assigned to me" and the person filter, and does not touch `monthTitles`. That is consistent, not an omission: the button says "clear all filters", the switch lives under Display, and its neighbour `scheduleDisplay` is exempt for the same reason.

**The hardcoded `min-height: 108px`.** The file already has `min-height: 80px` and `54px` in the same view, and yours is the only one of the three with the arithmetic written out above it. Nothing to change.

## Two observations, neither a request

The switch only appears below 640px, so somebody who turns it on and then opens the calendar on a desktop cannot turn it off there. That is harmless today, because the class does nothing outside the 639px query, and adding a second breakpoint check in JS to hide it is exactly the duplication your comment argues against. Leaving it is the right call; I am noting it so the reasoning is on the record.

At `--text-2xs` a title row is 10px. That is a deliberate trade the user opts into, and the dots stay the default, so this is not an objection. It is the number I would look at first if anyone reports that the month became hard to read.

## Approved, with two merge notes

No blocking findings and nothing I would call a should-fix. Two things before it lands:

- The branch conflicts with `main` in `CHANGELOG.md` only, and only because both sides added entries under Unreleased. A rebase resolves it.
- This changes `public/pages` and `public/styles`, so it rides the Tuesday interface train rather than merging the day it is approved.

Thank you for the comments explaining why things are where they are, and particularly for correcting the old tap-target comment instead of leaving a sentence behind that had quietly become half true.

PR #1030, opened 2026-09-06
fix(calendar): widen mobile Week's loaded range to match its 3-day window (#1006)

### LINKED ISSUE #1006: [Bug] Calendar view inconsistent
### Yuvomi version

2.62.0

### Deployment

TrueNAS SCALE (catalog app)

### Description

This appears to occur on mobile only. Please see screenshot attached. 

When I look in Week view, there appears to be no events on Sunday. However, when I look in the Agenda/Month view, the events on Sunday is shown. 

Image

Image

### Steps to reproduce

I am unsure how this can be reproduced....

Events should show up in either view

### Browser / device / OS

Chrome, mobile

### Logs

```shell

```

### Additional context

_No response_

mclgoerg:
## Summary

Fixes #1006. Mobile Week renders a 3-day window centered on the cursor
(`cursor-1..cursor+1`), but the calendar only loaded and indexed the fixed
7-day desktop week. Whenever the mobile window crossed that week's boundary,
the neighboring day was fetched (thanks to the existing 1-day timezone
margin in `fetchWindow()`) but then clamped back out by `buildDayIndex()`,
which only trusts `state.rangeFrom`/`state.rangeTo`. Calendar tasks (loaded
separately) and Agenda (which reloads a 31-day range) kept showing normally,
which is why the report read as an inconsistent Sunday rather than an
obviously broken day.

- `getWeekRange()` now accepts an optional `{ weekStart, mobile }` and, on
  mobile, returns the union of the configured desktop week and the 3-day
  cursor window whenever they diverge - otherwise it's unchanged.
- `getRangeForView()` passes the mobile viewport check through, so the
  loaded/indexed range and the rendered range can no longer drift apart.
- No second event-membership rule was added; Month, Day and Agenda paths are
  untouched.

## Test plan

- [x] New unit tests for `getWeekRange()` covering both boundary directions
  (Monday-start week + Sunday cursor -> includes the following Monday;
  Sunday-start week + Saturday cursor -> includes the following Sunday) plus
  a desktop and a mid-week mobile case that must NOT expand the range.
  `npm run test:calendar` - 111/111 pass.
- [x] `npm run test:frontend-audit` - 359/359 pass.
- [x] Full `npm test` - green.
- [x] Manual verification in a real browser against a throwaway instance of
  this branch: seeded a timed event on each boundary day, confirmed it was
  missing at 584px/390px before the fix and visible after, at both week-start
  boundary directions, with desktop unaffected either way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Mobile Week no longer silently drops timed events on its boundary day** (#1006). Mobile Week
+  renders a 3-day window centered on the cursor, but the calendar only loaded and indexed the fixed
+  7-day desktop week; whenever that 3-day window crossed the week boundary, the neighboring day's
+  timed events were fetched but then clamped out of the day index, while calendar tasks (loaded
+  separately) and Agenda (which reloads a 31-day range) kept showing normally - the mismatch read as
+  an inconsistent Sunday. The loaded range now covers the union of the desktop week and the mobile
+  window whenever they diverge; Month, Day and Agenda are unaffected.
+
 - **A same-version deployment now invalidates the installed PWA shell.** The served service worker
   receives a build-specific revision, so acceptance builds and rebuilt images no longer reuse an
   older cache merely because the application version has not changed.
--- a/public/pages/calendar.js
+++ b/public/pages/calendar.js
@@ -650,7 +650,10 @@ function setSavedCalendarView(view) {
 
 function getRangeForView(view, cursor) {
   if (view === 'month') return getMonthRange(cursor);
-  if (view === 'week') return getWeekRange(cursor);
+  if (view === 'week') {
+    const mobile = window.matchMedia?.(MOBILE_MEDIA_QUERY).matches ?? false;
+    return getWeekRange(cursor, { mobile });
+  }
   if (view === 'day') return { from: cursor, to: cursor };
   if (view === 'agenda') return getAgendaRange(cursor);
   return getMonthRange(cursor);
@@ -871,9 +874,28 @@ function getMonthRange(dateStr) {
   return { from, to };
 }
 
-function getWeekRange(dateStr) {
-  const weekStart = startOfWeekOf(dateStr);
-  return { from: weekStart, to: addDays(weekStart, 6) };
+/**
+ * Ladefenster der Wochenansicht: das haushaltsweite 7-Tage-Raster, auf Mobile
+ * erweitert um das 3-Tage-Fenster (cursor -1..+1), das renderWeekView() dort
+ * tatsächlich zeichnet. Ohne diese Erweiterung klammerte buildDayIndex() auf
+ * das reine 7-Tage-Raster, und ein Cursor am Wochenrand liess auf Mobile den
+ * hereinragenden Nachbartag ohne seine Termine stehen (#1006) - obwohl
+ * fetchWindow() ihn ohnehin mitlädt. Ein Randtag, der ausserhalb BEIDER
+ * Fenster liegt, bleibt weiterhin draussen; das ist kein Sonderfall des
+ * Ladefensters, sondern dieselbe Klammerung, die buildDayIndex() für jedes
+ * mehrtägige Event ohnehin anwendet.
+ */
+function getWeekRange(dateStr, { weekStart = state.weekStart, mobile = false } = {}) {
+  const start = startOfWeekOf(dateStr, weekStart);
+  const desktopFrom = start;
+  const desktopTo   = addDays(start, 6);
+  if (!mobile) return { from: desktopFrom, to: desktopTo };
+  const mobileFrom = addDays(dateStr, -1);
+  const mobileTo   = addDays(dateStr, 1);
+  return {
+    from: mobileFrom < desktopFrom ? mobileFrom : desktopFrom,
+    to:   mobileTo   > desktopTo   ? mobileTo   : desktopTo,
+  };
 }
 
 function getAgendaRange(dateStr) {
@@ -3034,6 +3056,7 @@ async function openFoundEvent(ev) {
 
 export const __test = {
   fetchWindow,
+  getWeekRange,
   resolveEventColor,
   isVisibleLayer,
   normalizeCalendarView,
--- a/public/pages/calendar.js
+++ b/public/pages/calendar.js
@@ -3057,6 +3057,7 @@ async function openFoundEvent(ev) {
 export const __test = {
   fetchWindow,
   getWeekRange,
+  getRangeForView,
   resolveEventColor,
   isVisibleLayer,
   normalizeCalendarView,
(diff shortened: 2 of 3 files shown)


ulsklyc (changes requested):
Read `cb05e99e` in a worktree and checked the diagnosis at the code rather than taking it from the description. It holds, including the part that explains why the report read the way it did.

## The diagnosis

`buildDayIndex()` clamps events to `state.rangeFrom`/`state.rangeTo` and drops anything outside (`calendar.js:956-965`), while tasks a few lines further down go into their bucket with no clamp at all (`:973`). So the clamp was never total, which is part of why this looked like a display quirk rather than a load failure.

It also matches the report exactly. @aserl-cell compared Week against Agenda and Month, and both of those keep the day: `getAgendaRange()` really does load 31 days (`:901`), and the month range contains the Sunday either way. Only mobile Week, whose rendered window reaches outside the loaded 7-day grid, lost it. Every part of that is checkable and every part checks out.

The fix itself is conservative in the right places:

- `startOfWeekOf(dateStr, weekStart = state.weekStart)` already took the second argument with that default (`:707`), so passing `state.weekStart` through changes nothing about desktop behaviour.
- The union with the desktop grid means a mid-week cursor produces the identical range, so `reloadForView()` does not start refetching on every three-day swipe. Only the two cursor positions at the week edges widen it, by one day.
- `window.matchMedia?.(MOBILE_MEDIA_QUERY).matches ?? false` short-circuits the whole chain when `matchMedia` is absent. I checked in node rather than assuming: it returns `false`, it does not throw. That matters here, because in the test loader `window` is `undefined`.

The five new tests cover the arithmetic properly: desktop unchanged, mid-week no-op, and all three edge cases across both week starts. Suites green on the branch: `test:calendar` 111, `test:frontend-audit` 359, `test:mobile-scroll-layout` 9, `test:display-timezone` 28, `test:household-timezone` 17, `test:changelog` 28. Merges cleanly on `main`.

## One thing to fix: no test reads the caller

All five tests pass `{ weekStart, mobile }` themselves, so they prove `getWeekRange()` and nothing else. The part of the fix that actually closes #1006 is the line in `getRangeForView()` that decides `mobile` - and nothing exercises it.

Counter-proof on top of your branch: I reverted `getRangeForView()` to `if (view === 'week') return getWeekRange(cursor);`, which makes the entire fix inert and puts #1006 straight back. `test:calendar`: **111 passed, 0 failed.** Unchanged.

A test that passes the option in can only tell you the function computes correctly with it; it cannot tell you anyone sets it. This has bitten the repo before, which is why I am asking rather than noting it.

It is buildable - I built it to be sure before asking. `window` is `undefined` under the browser loader, so it has to be stubbed at call time, and `getRangeForView` needs to be reachable:

```js
// public/pages/calendar.js
export const __test = { fetchWindow, getWeekRange, getRangeForView, ... };

// test/test-calendar.js
globalThis.window = { matchMedia: () => ({ matches: true }) };
assert.deepEqual(calendarHelpers.getRangeForView('week', '2026-03-15'),
  { from: '2026-03-09', to: '2026-03-16' });   // Sunday cursor reaches the next Monday
globalThis.window.matchMedia = () => ({ matches: false });
assert.deepEqual(calendarHelpers.getRangeForView('week', '2026-03-15'),
  { from: '2026-03-09', to: '2026-03-15' });   // desktop unchanged
```

Both assertions produce those values on your branch today, and the first one fails the moment the option stops being passed. Restore `globalThis.window` afterwards if the suite's later tests care.

## Timing

This is a `public/pages` change, so it rides the Tuesday interface train.

Two fixes from you in one night on issues nobody had answered yet - thank you. The diagnosis in the description is the useful kind: it names why the symptom looked like something else, which is what made #1006 hard to read from the report alone.

ulsklyc:
Follow-up to my review above. One correction to something I wrote, one addition, and the full chain.

## Correction: I got the `matchMedia` fallback wrong

I wrote: *"`window.matchMedia?.(MOBILE_MEDIA_QUERY).matches ?? false` short-circuits the whole chain when `matchMedia` is absent. I checked in node rather than assuming: it returns `false`, it does not throw. That matters here, because in the test loader `window` is `undefined`."*

The first sentence is true and the last one makes it false. Optional chaining guards `matchMedia` being missing **on an existing `window`**; it does nothing for `window` itself being undeclared, which is a plain `ReferenceError` on the identifier. Measured under the browser loader:

```
window.matchMedia?.('(max-width: 639px)').matches ?? false
  -> ReferenceError: window is not defined
globalThis.window = {};  same expression
  -> false
```

So my "I checked" sentence was checking the wrong half. **No consequence for the shipped code** - in a browser `window` always exists, and your line is correct as written. The consequence is for the test I asked for: the `globalThis.window` stub is not a nicety, it is the only way to call `getRangeForView()` from a suite at all. Which, I now think, is why you exported `getWeekRange` with an explicit parameter in the first place. That was a reasonable place to stop, and I should have said so instead of implying the caller was trivially reachable.

## Addition: your fix restores a convention, it does not add an exception

I said the fix was conservative but not why the *approach* is the right one. I would have reached for clamping `buildDayIndex()` to the fetch window instead, and I now think that would have been worse. Checked what the other views promise:

- `getMonthRange()` already returns the full 42-day grid - the days that are actually painted, trailing days of the neighbouring months included, not the calendar month.
- Day returns the single day it draws.
- Agenda returns its 31.

Every view already answers "the range is what gets drawn". Mobile Week was the only one that did not. Widening the range is therefore the house rule, and clamping the index to a second, wider window would have introduced the special case. Worth having on the record so nobody "simplifies" this later.

## Full chain

My first pass named six suites. Since this rides the Tuesday train I ran the whole thing on `cb05e99e`: **exit 0, 5429 tests, 0 failures** (read from the log file, not from a job notification).

One caveat on my own method, since it nearly produced a false accusation: an earlier run of this chain showed five failures in `test:password-normalization`. That was my fault - I had a second chain running against another worktree, and `test-backup-webdav.js` binds a fixed port. Isolated, that suite is 11/11 on your branch and 11/11 on `main`. Nothing to do with this PR; recording it so the number above is not mistaken for a re-run of a red result.

The test-reads-the-caller point from my review stands unchanged.

mclgoerg:
Added the caller-level regression test from the review.

- Exposed `getRangeForView` via `__test` (no production change).
- New test in `test/test-calendar.js` stubs `globalThis.window.matchMedia` (mobile `true`/`false`) and calls the real `getRangeForView('week', ...)`, asserting the mobile-expanded range and the unchanged desktop range. Restores the original `window` global in a `finally`.
- `npm run test:calendar`: 112/112.
- Counter-proof: temporarily reverted the Week branch to `return getWeekRange(cursor);` — exactly the new test failed (111/1), everything else stayed green. Restored, reran green.
- `npm run test:frontend-audit` (359/359), `test:mobile-scroll-layout` (9/9), `test:display-timezone` (28/28), `test:household-timezone` (17/17), `test:changelog` (27/27, 1 skip for missing git tag info in the sandbox, unrelated) — all green.
- Full `npm test`: exit 0, 5,429 tests, 0 failures.

Pushed to `cddb6713`.

Issue #884, opened 2026-08-26
[Bug] Health module: medication-schedules can't be deleted and random medications can't be taken


SETI-1234:
### Yuvomi version

2.46.0

### Deployment

Other (describe below)

### Description

Hello there,

in the first place the issue in the title seems to seperate but maybe it ist connected.

First problem: after a medication-schedule is created via the dialogue, it can't be deleted with the bin-sign. The error says: "Schedule not found" (I use the German version so the error says: "Einnahmeplan nicht gefunden").

Second Problem: If I want to "take" a medication it randomly says: "dosage-entry not found" ("Dosis-Eintrag nicht gefunden"). But this happens not always with the same medication. On one day the 7am-entry can be taken an on the other day the error occurs. I use the function to track the medication for another person. If Iog in with the account of the other person, the error won't occur on the medication that can't be taken by the other person.

I hope this can be a start to fix the problems and I'm willing to contribute more information.

### Steps to reproduce

First problem:
1. Create a medication and save it.
2. Open the medication.
3. Create a schedule.
4. Try to delete the schedule via the bin.
5. The error message occurs.

Second problem:
1. Create a medication and schedule like above.
2. Try to set the medication as "taken".
3. Sometimes the error message occurs.

### Expected vs. actual behavior

I described it above.

### Logs

```shell

```

### Additional context

I installed the App via the Proxmox Helper Script.

ulsklyc:
Thanks for the detailed report - the two problems are indeed connected, and your observation about the other person's account was the key.

**One cause, two symptoms.** A medication schedule and a dose entry have no `user_id` of their own; they hang off the medication and inherit its scoping. In four places that inheritance was written out by hand as `m.user_id = ?`, which quietly cuts out the caregiver relationship. So as a caregiver you could *create* for the person you look after, but not remove or tick off.

**Why the second one looked random:** as long as you tick a dose yourself, the request goes through the medication and works. Once the scheduler has created the entry ahead of time, the interface hits a different endpoint - and only that one was closed. Same button, working one day and not the next. Exactly what you described.

Fixed in #892, merged. Lab results had the same gap, just never reported.

The fix ships with the next release; I'll close this then. Thanks again - the note that it worked from the other person's account is what made it findable.

ulsklyc:
Released in **v2.48.0**.

[...]

PR #1031, opened 2026-09-06
fix(health): let caregivers correct medicine logs (#999)

### LINKED ISSUE #999: [Bug] Unable to update the medicine log for a child using another account
### Yuvomi version

2.61.0

### Deployment

### Description

When registering medication for a child using a parent account I’m not allowed to update the log entries the way I am when registering medication on my own account page. 

### Steps to reproduce

1. Using an instance with at least one child account and one parent account
2. Navigate to the Health / Medicine tab using the parent account
3. In the dropdown select the Child account to register for
4. Clicking entries in the Log does not show the Edit dialog. 

When clicking entries in the medicine log using your own account a dialog to edit entries is shown. This does not happen when clicking entries for another account that you are allowed to register medication for. 

### Browser / device / OS

Safari on iPhone

### Logs

```shell

```

### Additional context

_No response_

mclgoerg:
## Summary

Closes #999.

Authorized caregivers can now open and correct medication log entries for a cared-for person. The frontend now follows the same explicit care grant already enforced by the backend for recording, taking, skipping, updating, and deleting doses.

## Changes

- Gate intake-log edit controls through the shared caregiver predicate instead of owner-only equality.
- Preserve read-only behavior for unrelated family members.
- Preserve the existing scheduled-entry rule: return to Pending instead of Delete; ad-hoc entries remain deletable.
- Add a focused frontend regression test and an Unreleased changelog entry.

## Verification

- Test-first regression: failed before the fix because child log rows used owner-only equality; passes after the fix.
- npm run test:health-meds: 29/29 passed.
- npm run test:health-api: 104/104 passed.
- npm run test:frontend-audit: 359/359 passed.
- npm test: passed.
- Mobile Chromium at 390 x 844: both scheduled and ad-hoc child log rows opened the correction dialog; saving a scheduled entry as Pending retained its schedule and removed the taken timestamp; scheduled entries exposed no Delete action while ad-hoc entries did.

## Checklist

- [x] npm test passes
- [x] No new frontend dependencies
- [x] No new UI strings
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Authorized caregivers can now correct a cared-for person's medication log entries** (#999).
+  The log row now follows the same explicit care grant as recording, taking and skipping a dose;
+  other family members still see it read-only, and scheduled entries keep their existing
+  pending-instead-of-delete rule.
+
 - **A same-version deployment now invalidates the installed PWA shell.** The served service worker
   receives a build-specific revision, so acceptance builds and rebuilt images no longer reuse an
   older cache merely because the application version has not changed.
--- a/public/pages/health.js
+++ b/public/pages/health.js
@@ -1872,16 +1872,11 @@ function medLogHistoryMarkup() {
   if (!entries.length) return '';
   entries.sort((a, b) => String(b.at).localeCompare(String(a.at)));
 
-  // Korrigieren darf nur, wer IN SEINEM EIGENEN Protokoll ist (#701).
-  //
-  // Nicht `canEditFor`: der Server laesst eine Korrektur ausdruecklich nur den
-  // Eigentuemer machen (`ownLogRow` in server/routes/health/medications.js -
-  // ein Dosis-Eintrag ist eine Aufzeichnung ueber den eigenen Koerper, mitlesen
-  // und nachtraeglich aendern sind zwei verschiedene Rechte). Wer fuer eine
-  // betreute Person eintraegt, darf also buchen, aber nicht nachbessern - und
-  // bekam die Stifte trotzdem angeboten, mit einem 404 dahinter. Ein Knopf, der
-  // nichts tut, ist genau der Fehler, wegen dem #700 ueberhaupt aufkam.
-  const own = meds.personId === meds.meId;
+  // Dasselbe ausdrueckliche Betreuungsrecht wie beim Buchen und Abhaken gilt
+  // auch fuer die Korrektur (#999). Der Server zieht diese Grenze zentral ueber
+  // `ownLogRow` -> `writableChild`; die Oberflaeche muss deshalb denselben
+  // bereits geladenen Grant verwenden statt einer engeren Eigentuemerpruefung.
+  const own = canEditFor(meds.personId, meds.meId);
 
   const rows = entries.slice(0, 10).map((e) => {
     // Uebersprungen und ausstehend sind beide "nicht genommen" und treten
--- a/public/pages/health.js
+++ b/public/pages/health.js
@@ -5539,3 +5539,14 @@ function openCycleSettingsModal() {
     },
   });
 }
+
+export const __test = {
+  canEditFor,
+  // Testseam fuer #1031: setzt `careFor` fuer die drei Berechtigungsfaelle
+  // (eigene Daten, betreute Person, unbeteiligtes Mitglied), ohne das Array
+  // selbst nach aussen zu geben - Tests koennen die Betreuungsliste nur ganz
+  // ersetzen, nicht das Modul-interne Array direkt mutieren.
+  setCareForForTest(list) {
+    careFor = Array.isArray(list) ? [...list] : [];
+  },
+};
--- a/test/test-health-meds.js
+++ b/test/test-health-meds.js
@@ -5,6 +5,7 @@
  * Ausführen: node --loader ./test/test-browser-loader.mjs --test test/test-health-meds.js
  */
 import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
 import test from 'node:test';
 
 const {
@@ -325,4 +326,21 @@ test('scheduledLogs: eine Bedarfsdosis zaehlt nicht als eingehaltener Plan', () 
   assert.equal(computeAdherence(scheduledLogs(viele), 7).rate, 3 / 7);
 
   assert.deepEqual(scheduledLogs(null), []);
+});
+
+// --------------------------------------------------------
+// Betreuungsrechte im Einnahmeprotokoll (#999)
+// --------------------------------------------------------
+
+test('Einnahmeprotokoll: Korrektur folgt dem gemeinsamen Betreuungsrecht', () => {
+  const source = readFileSync(new URL('../public/pages/health.js', import.meta.url), 'utf8');
+  const start = source.indexOf('function medLogHistoryMarkup()');
+  const end = source.indexOf('/** Ein Log-Eintrag', start);
+  assert.ok(start >= 0 && end > start, 'medLogHistoryMarkup muss auffindbar sein');
+
+  const historyMarkup = source.slice(start, end);
+  assert.match(historyMarkup, /const own = canEditFor\(meds\.personId, meds\.meId\)/,
+    'auch eine ausdruecklich betreute Person braucht den Korrekturknopf');
+  assert.doesNotMatch(historyMarkup, /meds\.personId === meds\.meId/,
+    'eine reine Eigentuemerpruefung schneidet Betreuende vom erlaubten API-Weg ab');
 });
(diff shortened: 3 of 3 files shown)


ulsklyc (changes requested):
Read `b496f981` in a worktree and checked the one thing this PR turns on: the branch deletes a comment that claimed the opposite of what it now says, so one of the two had to be wrong. I decided it at the server code rather than by preference.

## The deleted comment was right when it was written, and stopped being right on 27 August

`public/pages/health.js` carried this since `8ac0c188` (17 August, v2.17.0): "Nicht `canEditFor`: der Server laesst eine Korrektur ausdruecklich nur den Eigentuemer machen ... Wer fuer eine betreute Person eintraegt, darf also buchen, aber nicht nachbessern - und bekam die Stifte trotzdem angeboten, mit einem 404 dahinter."

That was an accurate description of the server at the time. Ten days later `0c11c7db` (#884, v2.48.0) rewrote `ownLogRow()` to go through `writableChild()`, and the rule changed underneath the comment:

- `server/routes/health/medications.js:431` - `ownLogRow()` now calls `writableChild(...)`
- `server/routes/health/helpers.js:132` - `writableChild()` appends `writableClause()`
- `server/routes/health/helpers.js:109` - `writableClause()` is `(user_id = ? OR user_id IN (CARED_FOR_SUBQUERY))`

And the doc comment above `ownLogRow` (`medications.js:426`) says so in as many words: "Die Betreuung (#584) liegt innerhalb dieser Grenze: sie ist ausdrücklich erteilt, und ohne sie könnte ein Elternteil die Dosis, die es selbst eingetragen hat, nicht abhaken (#884)."

So the gap since 27 August was the mirror image of the one the old comment described: the server allowed the correction, the interface withheld the button. Your change is correct, and the replacement comment is correct. Please keep the dates in the commit message or the CHANGELOG entry if you can - "the server moved in #884 and the page did not follow" is more useful to the next person than "caregivers can now correct".

The rest checks out too. `canEditFor(personId, meId)` is `personId === meId || careFor.includes(personId)`, which mirrors `canWriteFor(viewer, ownerId)` in `helpers.js:96`, and the argument order matches the ten other call sites in the same file. `medLogHistoryMarkup()` is only rendered from the Meds tab, so reading `meds.personId` is the right state rather than `overview.personId`. After the change there is no remaining owner-only write check in `health.js`: apart from the definition of `canEditFor` itself, the two remaining `personId === meId` comparisons are `readOnlyBannerMarkup` (should the banner show at all) and `isOwnCycleView` (whose prediction settings apply), and neither is a write permission.

Suites green on the branch: `test:health-meds` 29, `test:health-api` 104, `test:health-structure` 5, `test:health-nav` 19, `test:frontend-audit` 359, `test:changelog` 28. Merges cleanly on `main`.

## One thing to fix: the test cannot see the bug it is guarding

The new test reads `health.js` as text and asserts that the block contains `const own = canEditFor(meds.personId, meds.meId)` and does not contain `meds.personId === meds.meId`. I ran two counter-proofs against the block itself:

**A. Revert the calculation** to `const own = meds.personId === meds.meId;` inside `medLogHistoryMarkup()` -> `test:health-meds` fails, 28 pass 1 fail. Good.

**B. Leave the calculation untouched, and neuter its use.** Change `if (!own) {` to `if (false) {` a few lines below - the branch that renders the read-only row disappears, so everyone who can see the list gets the correction controls, including a member with no care grant - which is precisely the see-versus-change line the server comment draws. `test:health-meds`: **29 pass, 0 fail.** The guard never notices, because the line it greps for is still there, word for word.

That is the failure mode this repo has hit before: a text guard proves the spelling, not the rule, and dead or bypassed code passes it unchanged. Right now `own` is used three lines later, so the two are hard to separate - but a future refactor that moves the rows into a helper and forgets to pass `own` would ship exactly the permission bug this PR is fixing, with a green suite behind it.

I do not think a full behavioural test is available here: `public/pages/health.js` has no `__test` export, so `medLogHistoryMarkup()` cannot be called from a suite at all. Given that, two options, either is fine:

1. Extend the same guard to the *use*: assert that the block still contains a branch on `own` (for example `/if \(!own\)/`), so removing the consumer is as red as removing the producer. Cheap, and it closes the exact hole above.
2. If you would rather have a real test, export `medLogHistoryMarkup` (or just the decision) on a `__test` object and assert on the produced markup for three cases: self, cared-for person, unrelated member. That is a bigger change than this PR should carry, and I would take it as a follow-up rather than a condition.

Option 1 is enough for me to merge this.

## Timing

This touches `public/pages`, so it rides the Tuesday interface train rather than merging the day it is ready.

Thank you for going to `writableClause` instead of trusting the comment above the line you were changing. The comment was the more confident of the two sources, and it was the wrong one.

ulsklyc:
Follow-up to my review above, before you act on it. **Option 1 is weaker than I claimed - please do not build on that sentence.** Also one more piece of evidence in your favour, and the full chain.

## Correction: "Option 1 is enough for me to merge this" was wrong

I offered you a cheap way out: extend the text guard to the *use* of `own`, e.g. `assert.match(historyMarkup, /if \(!own\)/)`, and said that closes the hole. It closes the hole I happened to demonstrate. It does not close the one a level below.

I built Option 1 into your test and then broke the rule without touching a single character the guard inspects - I gutted `canEditFor` itself:

```js
function canEditFor(personId, meId) {
  if (personId == null) return false;
  return personId === meId;          // careFor.includes(personId) removed
}
```

`const own = canEditFor(meds.personId, meds.meId)` still reads exactly as asserted, `if (!own)` is still there, and #999 is fully back: a caregiver gets the read-only row again.

```
Option 1 in place, code untouched:              29 tests, 29 pass, 0 fail
Option 1 in place, canEditFor gutted (bug back): 29 tests, 29 pass, 0 fail
```

That is the general shape of the problem and I walked into it while proposing the fix: a guard on a call site tells you the call is written, never that the callee still means anything. Chaining more `assert.match` down the call graph is just moving the blind spot one function further.

So the honest options are now:

1. **Option 2 from my review, the real test**: a `__test` seam on `health.js` exposing `canEditFor` and a way to set `careFor`, then three assertions - self, cared-for, unrelated. That is the only version that survives the probe above.
2. **Or: merge as is, and I take the seam as follow-up work.** Your change is correct, it is one line, and the text guard does prove it was not simply deleted. `health.js` is 5546 lines and is already scheduled to be split; a `__test` export belongs to that work rather than to a bug fix from an outside contributor.

**I would rather you took option 2 and I did the seam.** Asking you to open up the largest page file in the repo in order to land a one-line permission fix is a bad trade, and my review put that cost on you on the strength of a claim I had not measured. Your call either way - say which and I will follow it.

## One more reason your reading is right

My review dated the change through git (`0c11c7db`, #884). There is a second, independent statement of the same rule that I missed, and it names medications explicitly. `server/routes/health/cycle.js:16-22`, explaining why the cycle is the deliberate exception:

> BEWUSSTE AUSNAHME von der Betreuung (#584): Wo Vitalwerte, Medikamente, Laborbefunde und Aktivitäten `careAwareClause()` verwenden, bleibt es hier bei `visibilityClause()`. Fieber messen und Medikamente geben ist Fürsorge; das Zyklus-Tagebuch einer anderen Person mitzulesen ist es nicht.

I checked that this holds per area instead of trusting it: vitals, labs, activities and medications all use `careAwareClause` plus `writableClause`/`writableChild`; cycle uses `visibilityClause` alone. So the rule you restored was not merely implied by a refactor - it was written down, with medications named, and the page had drifted off it.

## Full chain

My first pass named six suites. Whole chain on `b496f981`: **exit 0, 5430 tests, 0 failures** (read from the log file, not from a job notification).

mclgoerg:
Implemented Option 2 from the review — a real behavioral test for `canEditFor()` instead of only a call-site text assertion.

- Added a narrow `__test` export at the end of `public/pages/health.js`: the existing `canEditFor` function plus a defensive setter for `careFor` (copies the array in, resets to `[]` on invalid input). `careFor` itself and `medLogHistoryMarkup()` stay unexposed, and production behavior is unchanged.
- New test in `test/test-health-meds.js` imports the page's `__test` object through the browser loader and asserts all three permission outcomes: own data (`true`), a cared-for person (`true`), an unrelated member (`false`). Resets the grant list in a `finally`.
- `npm run test:health-meds`: 30/30.
- Counter-proof: removed `careFor.includes(personId)` → cared-for assertion failed (29/1); then made the predicate accept every non-null person → unrelated-member assertion failed (29/1). Restored the real predicate, reran green (30/30).
- `test:health-api` (104/104), `test:health-structure` (5/5), `test:health-nav` (19/19), `test:frontend-audit` (359/359), `test:changelog` (27/27, 1 unrelated skip) — all green.
- Full `npm test`: exit 0, 5,431 tests, 0 failures. (One transient, unrelated `test-dashboard.js` real-clock flake showed up on the first full run and was gone on an isolated rerun and a full rerun — confirmed unrelated by also being absent on the #1030 branch's own full run, which doesn't touch dashboard.js or health.js.)

The existing source-guard test and the #884/#999 permission reasoning are untouched. Pushed to `9ea4c07f`.

PR #1032, opened 2026-09-06
fix(calendar): reconcile undoable deletes across ranges

### LINKED ISSUE #975: [Bug] Edit recurring events not possible
### Yuvomi version

2.57.2

### Deployment

### Description

You can't edit recurring events. It only changes the one that you click on which is very inconvenient. I also can't figure out how to delete recurring events which is quite bad. 

### Steps to reproduce

Schedule recurring events and then edit 

Able to edit entire series of recurring events

### Browser / device / OS

_No response_

### Logs

```shell

```

### Additional context

_No response_

PechJiri:
## Summary

Fixes the standalone client-side deletion race identified in #975 (PR 1 of the agreed three-part plan).

During the five-second Undo window the server still owns the row. A range reload could therefore reinsert a just-hidden event, and Undo could append a duplicate. The calendar now treats pending deletions as an overlay over every server response, then reloads the selected range after a successful delayed commit.

This PR intentionally does not close #975; the recurrence UX and linked occurrence-override model remain separate follow-up work.

## Changes

- adds a calendar-specific pending-delete lifecycle for:
  - a single non-recurring event
  - one occurrence
  - this and following
  - the whole series
- reapplies active deletion tombstones after fresh calendar responses
- restores the freshest server row exactly once on Undo
- reconciles the complete selected range after a committed delete
- uses one generation lane for full-range and event-only responses
- rejects stale successes and failures when the view/cursor changed, including a quick out-and-back navigation to an already cached range
- serializes delayed mutations for the same series in initiation order, preventing two out-of-order truncation PUTs from widening the series
- preserves keepalive/pagehide behavior and existing localized error handling
- precaches the new browser helper and documents the behavior in SPEC, CHANGELOG, and the test-suite catalogue

The lifecycle deliberately mirrors `public/utils/document-folder-delete.js`. It remains calendar-specific for now because calendar occurrence identity, scope matching, range coordination, and same-series write ordering are distinct enough that extracting a shared abstraction here would couple two destructive workflows prematurely.

## Reviewer feedback addressed

The implementation includes the additional cases raised in the issue and in two independent review rounds:

- range reload during Undo must not resurrect a row
- Undo after a reload must not duplicate a row
- stale full-range data must not overwrite events, tasks, holidays, schedule entries/warnings, offline/error state, or range bounds
- returning to a cached range invalidates the request for the range just left
- full and event-only loaders invalidate each other
- overlapping tombstones of the same series restore correctly in either Undo order
- a committed overlapping deletion cannot be resurrected by another pending Undo or an older event-only response
- multiple `this and following` writes for one series reach the server in user-action order
- Undoing a queued operation releases the next write

## Data and synchronization impact

- no migration or schema change
- no server API change
- no recurrence expansion change
- no changes to Google, CalDAV/Apple, Outlook, ICS subscription/import/export, reminders, or outbound sync behavior
- the existing endpoints and keepalive options are preserved; only client-side orchestration and subsequent reads change
- applies to all three existing delete paths, regardless of whether the visible row is local or externally managed, without changing source ownership or delete semantics

## Verification

- `npm test` — pass
- `npm run test:calendar` — existing 106 checks plus 22 new deletion/race checks pass
- `npm run test:sw-precache` — 7 pass
- `npm run test:suite-chain` — 4 pass
- `npm run test:changelog` — 28 pass
- `npm run test:layer-boundary` — pass
- `npm run test:frontend-audit` — 359 pass
- `git diff --check` — pass
- mutation probes confirmed the suite fails when:
  - the pending overlay is bypassed
  - an obsolete response is accepted
  - Undo uses append-without-deduplication
  - same-series write serialization is removed
- disposable Docker/browser smoke using a separate in-memory database:
  - created a three-occurrence local monthly series through the real API
  - loaded the real calendar page and production browser module without page errors
  - confirmed the deleted occurrence stays hidden across a pre-commit range reload
  - confirmed Undo restores it exactly once
  - confirmed commit leaves only the two expected later occurrences
  - removed the disposable container, database, and image after the test

## Checklist

- [x] `npm test` passes
- [x] No new UI strings; existing localized messages are reused
- [x] CHANGELOG.md updated

Refs #975


ulsklyc (approved):
@PechJiri Approving this. I read the code rather than the description, ran the chain, and tried to break the suite on purpose. It held every time.

## What I verified

**The full chain on `81f69569`: exit 0, 5451 tests, 0 failures.** Read from the log file rather than from a notification, because a backgrounded run reports the exit code of the last command in the pipe and has lied to me twice this week. `npm run test:calendar` gives 22/22 on the new file.

**Four mutation probes, each undoing the whole cause rather than half of it:**

| Probe | Result |
|---|---|
| overlay filter in `applyPendingCalendarDeleteOverlay` removed | 8 failures |
| Undo appends instead of `mergeInCanonicalOrder` | 8 failures |
| `await previous` removed from `reserveSeriesWrite` (calendar-delete.js:84) | 1 failure |
| generation guard removed from the coordinator | 5 failures |

That is the part I care about most: a green suite that stays green when you break the thing it claims to test is worse than no suite. This one is anchored.

**You picked up the duplicate case I had found but not asked for.** `buildDayIndex` does not deduplicate by id, so an Undo after a range change used to be able to show the event twice. `mergeInCanonicalOrder` (calendar-delete.js:106) keys on `id\0start_datetime` and restores exactly one row. Thank you for reading past the literal request.

**One silent improvement you did not mention.** The error path of `loadRange` now clears `state.scheduleWarnings` (calendar.js:1234). It did not before, so a failed load left the previous range's warnings standing. Worth a CHANGELOG line of its own, but I am not going to hold the PR for it.

## Two nice-to-haves, neither blocking

**1. One of the 22 tests is a text guard, and it guards the seam.** `calendar page guards the complete range and wires every delete scope` (test-calendar-delete.js:406) reads `calendar.js` as a string and asserts on spelling: `assert.match(whole, /reloadEvents: reloadCalendarRangeAfterDelete/)`. The other 21 are real behaviour tests against the helper, so the proportion is fine and I would not have written it differently under the constraint. But be aware what it does and does not catch: rename the helper and it goes red for the wrong reason, while moving `transition.commit()` in front of `await write.run(...)` - a real ordering bug - leaves it green. The seam between page and helper is the one place here still covered by notation rather than by meaning.

**2. `reloadCalendarRangeAfterDelete` reloads four things where one changed.** It calls `loadRange` (calendar.js:1313), which refetches events, tasks, holidays and schedule entries, five seconds after every deletion. A calendar delete cannot change the other three. `reloadCalendarEventsOnly()` would do. If the full load is deliberate - it also resets `rangeFrom`/`rangeTo` and clears `loadError`, and I can see the argument for one authoritative path after a destructive commit - then please say so in the comment, because the next reader will otherwise "fix" it.

## Two observations, no action needed

**The rule catch-up in `deleteThisAndFollowing` is gone on purpose.** The old code patched the remaining instances in place:

```js
for (const e of state.events) if (e.id === event.id) e.recurrence_rule = newRule;
```

That is now carried by the reload, which is the better answer - the server owns the truncated rule, the client was guessing it. It is worth one sentence in the code, since nothing in the diff says why the line left.

**A second-order edge case I could not turn into a defect.** If the commit falls due while the view is inactive, no reload runs; when the user returns to the same range, `reloadForView` skips `loadRange` because the bounds did not change. What is displayed is still correct - the overlay already filtered the row - but a truncated `recurrence_rule` on the remaining instances stays stale until the next real load. I could not build a case where a user sees a wrong result from this, so I am recording it rather than asking for anything.

## On the shared abstraction

Your reasoning for keeping this next to `document-folder-delete.js` instead of merging them convinces me, and I want it on the record so the question does not get reopened as a cleanup later: occurrence identity, scope matching and same-series write ordering have no counterpart in the folder case, and a shared abstraction over two destructive workflows would have to be right for both at once. Two honest copies beat one wrong generalisation here.

## Merge

Conflict-free against `main` (`04f87334`). It also does not collide with #1030 in `calendar.js` - only in the CHANGELOG - so the two can go in either order.

This does not close #975, as you note. PR 2 and PR 3 remain, and PR 3 still waits on the #1022 merge for its migration numbers.

PR #1034, opened 2026-09-06
fix: a recurring entry keeps its account, and the desktop sidebar keeps its scrollbar

### LINKED ISSUE #970: [Bug] Scroll bar for desktop view
### Yuvomi version

2.58.0

### Deployment

Portainer stack

### Description

I noticed that the left side menu does not have a scroll bar option in the desktop browser mode.  It makes it really difficult to navigate that menu with out it.  

### Steps to reproduce

1. Open app in a windows browser, such as Edge or Google Chrome.
2. Move mouse curser to left side but cannot scroll the menu up or down as there is no scroll bar.
3. Left click on an empty part of the menu not a menu item and you can use curser keys sometimes to navigate up and down.  

Would like to see a scroll bar on web browser for non-mobile devices that use a keyboard and mouse.  

### Browser / device / OS

Microsoft Edge/Desktop Computer/Windows 11

### Logs

```shell

```

### Additional context

_No response_

### LINKED ISSUE #973: [Bug] New recurring expense entry stores the account setting only for the first occurrence
### Yuvomi version

2.59.0

### Deployment

Portainer stack

### Description

I created a new Expense adding a new entry on Budget / Overview. Under "More settings" I chose an Account, enabled Recurring and set Interval to Monthly.

Image

The new item showed up on the current month's Transactions list as a recurring expense with the chosen account.

Image

But the recurrences for the following months had no account (i. e. 2026-09).

Image

Then I edited the second occurrence of the expense to set the same account as the first one. I confirmed the changes and chose "Change all future occurrences" on the Recurring payment popup. After the Transactions list was updated automatically, all the occurrences of the expense but the first one disappeared from the list.

I tried to update the transaction list, I refreshed the page in the browser, sign out and sign in again but I was unable to find the expense occurrences from 2026-09 on.

### Steps to reproduce

1. Navigate to Budget / Overview
2. Create a new Entry. Under "More settings" chose an Account, enable Recurring and set Interval to Monthly.
3. Add the entry. The Transactions list displays it as recurring expense with the chosen account.
4. Display the Overview for the next month. The second occurrence has no account set.
5. Edit the second occurrence and set the same account as the first occurrence.
6. Confirm the changes.
7. When Recurring payment dialogue pops up, choose "Change all future occurrences".
8. The Transactions list updates and the second occurrence of the expense disappears.
9. Update the budget Overview. All the occurrences of the expense but the first one are lost.

Expected:
All the occurrences of a newly created recurring expense store the account set on the New Entry popup.

Actual:
Only the first occurrence of a newly created recurring expense stores the account set on the New Entry popup.

### Browser / device / OS

Firefox 154.0.1 / Google Chrome 152.0.7977.65 on Windows 11 25H2

### Logs

```shell

```

### Additional context

_No response_

ulsklyc:
Two reported bugs, both reproduced at the code before anything was changed.

## #973 - a recurring budget entry lost its account after the first month

Two places, not one, and the second is why the reporter's own repair attempt failed.

**`generateRecurringInstances` (server/routes/budget/helpers.js)** materialises every month after the first from the series original. Its INSERT inherited title, amount, category, subcategory, owner, visibility and the pending flag - but not `account_id`. Only the first booking is entered by hand, so its account was always right and the gap only showed from the second month on. Reproduced against an in-memory database before the fix: a monthly series carrying account 9 produced two instances, both with `account_id = null`.

**`PUT /:id/series` (server/routes/budget/entries.js)** never wrote `account_id` at all, while `PUT /:id` has handled it correctly for a long time. So setting the account on a later occurrence and choosing *change all future occurrences* did nothing to the account - and the same route then deleted that occurrence, as it always does, leaving the impression that the entry was gone. It was regenerated on the next read, again without an account; with an account filter active it stayed invisible.

The series route now follows the convention the single-entry route already established: omitted leaves the account untouched, `null` clears it, an id sets it, an unknown account is a 400 before anything is written.

**It deliberately does not reach back into past occurrences.** Visibility must, because a stale wider value is a leak. An account is a fact about a booking that already happened, and the future occurrences inherit the new value anyway, because they are deleted and regenerated.

Five tests: the inheritance itself (second and third month, not the first), the no-account case, the virtual-series branch (it picks its dates differently and runs through the same INSERT), setting and clearing through the route, that a title-only update does not wipe the account, the 400, and that a past occurrence keeps its own. Each was counter-proved by reverting the corresponding half.

**One thing found while writing them:** the hand-written schema in `test-budget-recurrence.js` was three columns behind the real one - `account_id` and `updated_at` missing, and the `visibility` CHECK unaware of `shared_amount` from migration 156 (#659). Brought up to date, with a note that a transcript ages and where to check it against.

## #970 - the desktop sidebar had no scrollbar

`scrollbar-width: none` plus a hidden `::-webkit-scrollbar`, inside `@media (min-width: 1024px)` - removed precisely on the widths where a mouse is the input device.

The soft fade underneath was added for a different question. It answers "is there more below?", which is what made Budget, Health and Settings findable on 800px laptops. It does not answer "what do I drag?". The reporter is on Windows with a mouse; the suggestion in the thread to use a two-finger swipe answers a trackpad, not the report.

The fade stays. The bar returns as `scrollbar-width: thin` with `scrollbar-color` built from `--tint-hint`, the step tokens.css already names for scrollbar thumbs, written like the module scrollbars in `budget.css` but without `--module-accent`: the sidebar belongs to no module and must not take the colour of whichever one happens to be open.

**Stated in the comment rather than implied:** the fade is a `mask-image` on this same element, and a mask covers the element box including the scrollbar, so on platforms with a classic space-taking bar its lowest stretch fades out. Interaction is unaffected - a mask changes opacity, not hit-testing - but this could not be verified here, because macOS draws overlay bars that never appear in a screenshot. If it turns out to bother anyone, the fix is to move the fade from a mask to an overlaid gradient pseudo-element, not to hide the bar again.

The guard asserts the state **inside** the desktop media query, using `eachRule()` rather than a regex over the file: `scrollbar-width: none` remains correct on phones and must not trip it. Counter-proved twice - restoring the old state, and writing the colour as a literal instead of a token.

## Verification

- Full chain: **exit 0, 5434 tests, 0 failures** (5429 on `main` plus 5 new), read from the log file rather than from a job notification
- `test:budget-recurrence` 29, `test:budget-entries-routes` 56, `test:frontend-audit` 360, `test:mobile-scroll-layout` 9, `test:changelog` 28
- Every fix counter-proved: reverting it turns the matching test red and nothing else

## Release timing

`public/styles/layout.css` makes #970 an interface change, so this belongs to the Tuesday train. #973 is server-only and rides along in the same release.

Closes #973
Closes #970


ulsklyc:
Status before merging, since the open review comments above are a mix of fixed and deliberate.

**Fixed since they were written** (the comments keep their line anchors, so they still show against the newest commit even though they describe earlier ones):

- *Dropped PWA changelog entry* - restored in `506c788f`; the diff no longer deletes a line.
- *Cutoff at the first of the month* - moved to today in `506c788f`, then to `todayKey(db)` in `ef51da80` so it uses the household timezone rather than the server's, the same boundary `listAccounts()` reads at.
- *Wiping the parent account from an account-less child* - fixed in `be1682cc`. Reproduced end to end first: an empty account field on a legacy instance sent `account_id: null` and cleared the series account. It now only counts as "remove" when the instance carried one.
- *Smoothed estimates in account balances* - fixed in `ef51da80`. Eight months of a virtual yearly series produced -800 of reported balance for a debit that had not happened. Virtual series no longer pass their account to instances, and the test that asserted the opposite was mine.
- *Instances created before the upgrade* - migration v181 in `be1682cc`, deliberately a one-off rather than a rule in the read path: "instance inherits the series account" applied on every read could never again be told apart from "this booking deliberately had no account", and it would have to be repeated in every read query. It leaves originals, differing accounts and virtual series alone.

**Deliberately not fixed, and worth stating plainly:** changing the account for all future occurrences also rewrites the account of the series original, which doubles as the first hand-entered booking. That is correct as reported. It has also been true for the title and the amount since the route existed - the original is both the template and a booking. Separating them means giving the series its own account column, which is a migration and a data model change, not a bug fix. If it should be separated, that belongs in its own issue.

One change here goes beyond the reported bug and is worth a second pair of eyes: future occurrences are now **updated in place**, and deletion is reserved for a changed rhythm. The delete-and-recreate was what forced the bad trade in the first place - cut early and you move completed bookings, cut late and title and amount never reach the current month. It also cost the rows their identity and their receipts.

Full chain locally on `be1682cc`: exit 0, 5443 tests, 0 failures.

PR #1037, opened 2026-09-06
feat(calendar): clarify recurring events

PechJiri:
## Summary

Implements PR 2 from the agreed plan in #975, following the merged delete-reconciliation fix in #1032.

Recurring events are now identifiable before opening them, and the monthly "last day of month" option explains the concrete dates it will use instead of leaving the user to infer them from a generic sentence.

Refs #975

## Changes

- Consolidates the existing recurring-event marker behind one recurrence-aware renderer, extends it to month view, and gives it an accessible name that survives Lucide's SVG replacement in month, week, day and agenda views.
- Keeps the compact mobile month grid dots-only; opening the day exposes the marked title in the detailed view.
- Gives the repeat marker an accessible name. The month-day and agenda controls, which provide their own overriding accessible names, now carry the recurrence meaning there as well.
- Replaces the calendar's generic month-end hint with localized concrete dates:
  - a start date that is not month-end names both the entered date and the first actual occurrence;
  - a start date already at month-end gets a neutral confirmation;
  - invalid or temporarily empty input safely falls back to the existing generic explanation;
  - imported rules with additional constraints such as `BYDAY`, and rules ending before the calculated first occurrence, also retain the generic wording rather than displaying a plausible but false date.
- Preserves the separate task wording and behavior: tasks remain due on their entered date and only the next run moves to month-end.
- Passes the calendar's active start date through an explicit component input, as requested in #975. The shared RRULE widget never queries or guesses calendar-owned field selectors. The calendar itself chooses the timed or all-day start field.
- Recomputes the preview from the RRULE controls currently visible in the form, including live changes to `UNTIL` and the end mode. It therefore describes the same rule that would be submitted, not a stale rule captured when the modal opened.
- Adds all three new strings to all 24 supported locales with matching interpolation placeholders.
- Updates `CHANGELOG.md`, the recurrence specification and the test-suite catalogue.

## Data and synchronization impact

- No database migration or schema change.
- No API, recurrence storage or recurrence expansion change.
- No change to Google Calendar, CalDAV/Apple, Outlook, ICS subscription, import or export behavior.
- The marker only reflects recurrence metadata already present on rendered rows.
- Imported `RECURRENCE-ID` overrides currently normalize to detached rows without `recurrence_rule` or another reliable presentation flag. They intentionally remain unmarked here: inferring series membership from UID would be ambiguous, and the planned unified override model in PR 3 should provide explicit metadata.

## Review follow-ups

Two separate Superpowers review rounds were run over immutable branch heads and every actionable finding was incorporated.

Round 1 found and resolved:

- concrete dates could be wrong for imported monthly rules combining `BYMONTHDAY=-1` with extra constraints;
- a month cell's own `aria-label` could hide the recurrence meaning of child chips;
- the current `RECURRENCE-ID` normalization boundary needed to be documented explicitly;
- tests needed server-engine cross-checks and stronger accessibility coverage.

Round 2 found and resolved:

- live edits to a series end condition still used the original RRULE and could leave a stale concrete date visible;
- helper/source assertions needed a composed-browser counterpart.

The added browser regression opens the real calendar modal, exercises timed and all-day date fields, changes an early `UNTIL`, saves a recurring event, and inspects the month and agenda accessible names after Lucide has replaced the icon markup.

## Verification

- `npm test` - full suite passes on Node 24.19.0 (exit 0) after the final review fixes.
- `npm run test:rrule-ui` - 33/33.
- `npm run test:calendar` - 114/114 calendar checks plus 22/22 delete-lifecycle checks.
- `npm run test:frontend-audit` - 360/360.
- `npm run test:i18n` - 53/53.
- `npm run test:i18n-plural` - 14/14.
- `npm run test:changelog` - 28/28.
- Focused composed-browser regression - 1/1.
- `git diff --check` - clean.

The new behavioral tests were observed failing before their respective fixes: the imported-rule preview, the month-cell accessible name, and the live early-`UNTIL` case. The separate full `test:document-guards` sweep remains the documented pre-release handrail; its new focused regression has been run independently here.

## AI assistance

This change was drafted and reviewed with OpenAI Codex. I inspected the resulting diff and can explain the implementation and its boundaries.

## Checklist

- [x] `npm test` passes
- [x] CHANGELOG.md updated


PechJiri:
Pushed the follow-up from two separate review rounds in `0a8dd4cf` and `10c3929e`.

What changed after review:

- The concrete last-day preview now stays generic for imported RRULEs with additional constraints (for example `BYDAY`) instead of calculating a date from `BYMONTHDAY=-1` alone.
- Rules whose `UNTIL` is before the first possible month-end occurrence also stay generic. This is cross-checked against the server recurrence engine.
- The live preview now reads the RRULE currently composed by the form and refreshes on end-mode, `UNTIL`, and count changes. A newly entered early `UNTIL` can no longer leave a stale, impossible date on screen.
- The month-day control's accessible name now retains the title and recurrence meaning that its own `aria-label` would otherwise hide from assistive technology.
- The current imported `RECURRENCE-ID` limitation is documented explicitly. Those detached normalized rows have no reliable recurrence presentation flag today; this PR does not guess from UID and leaves that to the agreed unified override model.
- Added a real-browser composition test which opens the calendar modal, exercises the timed/all-day start switch and early `UNTIL`, saves the series, and verifies month/agenda accessible names after Lucide replacement.

Verification after the final commit:

- full `npm test`: pass (Node 24.19.0, exit 0)
- RRULE UI: 33/33
- calendar: 114/114 plus delete lifecycle 22/22
- frontend audit: 360/360
- i18n: 53/53; plural contracts: 14/14
- changelog: 28/28
- focused composed-browser regression: 1/1
- `git diff --check`: clean

No database, API, persistence, recurrence-expansion, or synchronization code changed. The PR body has been updated with the full scope, review trail, test evidence, and the explicit PR 3 boundary.

ulsklyc (approved):
Approving. This is PR 2 as agreed, and it satisfies the condition I set in #975 more strictly than I asked for.

## The condition, and how you exceeded it

I asked that the concrete date arrive as an explicit input from the caller rather than the widget reaching into a field it does not own. You did that - and then added the guard I did not think to ask for:

```js
assert.doesNotMatch(readFileSync('../public/rrule-ui.js'), /modal-(?:allday-)?start-date|modal-allday-start/)
```

That is the better half. `getStartDate` being present today is a fact about today; the widget *never knowing* calendar-owned selectors is the property that has to survive the next refactor. Every `querySelector` left in `rrule-ui.js` targets `#${prefix}-rrule-*` - elements the widget builds itself.

The caller-side handling of the all-day switch is right too: `#modal-allday-start` vs `#modal-start-date` is exactly the ambiguity that made me refuse a DOM lookup in the first place, and it belongs where you put it - in the calendar, which knows its own form.

## What I measured

Three mutation probes, each undoing a whole cause:

| Probe | Result |
|---|---|
| round-trip validation of year/month/day removed | 33 -> 32/1 |
| `canShowConcreteMonthEnd()` short-circuited | 33 -> **30/3** |
| caller stops passing `getStartDate` | `test:rrule-ui` red |

The middle one is the one I care about most, and the three tests that fall are named for the right risk: *"eine komplexere importierte Monatsregel bekommt keine erfundene konkrete Vorschau"*, *"die Live-Vorschau erfindet auch nach Datumsänderung keinen Termin für eine komplexe Regel"*, *"ein live gesetztes Serienende vor dem ersten Monatsletzten nimmt das Versprechen zurück"*. A wrong-but-plausible date is worse than the generic sentence it replaces, and the suite is anchored on exactly that.

The round-trip check earns its comment. `new Date(Date.UTC(2026, 1, 31))` silently becomes 3 March, and a preview built on it would be confidently wrong - the kind of failure nobody reports because it looks like an answer.

Full chain on `10c3929e`: **exit 0, 280 suite links, 6068 assertions, 0 failures**, read from the log file. No conflicts against `main`; the branch is built on `3d82ebd5`, which is current.

Hard constraints: no new `innerHTML`, `esc()` on every interpolated title, no colour literals in the CSS, all three new strings through `t()` and present in all 24 locales (`test:i18n` 53/53, `test:i18n-translated` 92/92).

## One correction to the description, not to the code

"Adds a small repeat icon" understates what happened. The icon already existed on `main` and was rendered in week, day and agenda - but as `<i aria-hidden="true">`, invisible to screen readers, with the `(recurrence_rule || is_recurring_instance)` condition duplicated at all three call sites.

What this PR actually does is better than adding one: it moves the condition into the function, extends it to month, and gives the marker a name. Worth saying because the diff reads as "new feature" while the valuable part is the consolidation.

The class moving from the `<i>` to a wrapping `<span>` is the right call and easy to misread as churn - a class on the `<i>` does not survive Lucide's SVG replacement, which is why `role="img"` and `aria-label` had to move outward. Your document-guard asserts `repeatIsSvg === true` against the real replacement rather than the markup you wrote, which is the only version of that test worth having.

And the duplicate-announcement question is handled: month-day and agenda controls carry overriding accessible names, so the marker's own label would be swallowed there - you folded the recurrence meaning into those labels instead. That is easy to get wrong in the direction of announcing everything twice.

Merging with the Tuesday interface train.

ulsklyc:
One addition to my review, because I understated your test coverage.

I reported the third probe - caller stops passing `getStartDate` - as turning `test:rrule-ui` red. A suite run that finished after I wrote that shows it takes down **two** suites:

```
test:calendar         EXIT=0   114 passed        <- unaffected, as expected
test:rrule-ui         EXIT=1   1 failure         <- the text guard on the bind call
test:document-guards  EXIT=1   1 failure         <- the composed browser check
```

The second one is the stronger belt. A text guard proves the call is written; `test-document-guards.js` renders the real composed form and reads the actual hint text back, so it proves the date arrives and produces the right sentence. That is the difference between asserting a call site and asserting a behaviour, and having both is the right answer here - the text guard is what keeps `rrule-ui.js` from ever learning the calendar's selectors, which is not a behaviour you can observe from output at all.

Worth correcting in your favour rather than leaving my summary as the record.

PechJiri:
Implemented the maintainer's requested description correction. The first Changes bullet now states that this consolidates the existing recurring-event marker behind one recurrence-aware renderer, extends it to month view, and gives it an accessible name that survives Lucide's SVG replacement across month, week, day, and agenda views. No code or head commit changed; the existing approval and CI evidence remain bound to `10c3929e`.

PR #1038, opened 2026-09-06
docs(calendar): explain delete reconciliation

PechJiri:
## Summary

Carries the documentation-only follow-up that was pushed to the source branch of #1032 after that PR had already merged.

## Changes

- Explains why the delayed-delete commit deliberately performs a full authoritative range reload instead of a partial event refresh.
- Records why the former client-side recurrence-rule patch loop is gone.
- Completes the existing changelog entry with the stale schedule-warning cleanup already implemented by #1032.

There is no behavior change. The diff is limited to `CHANGELOG.md` and comments in `public/pages/calendar.js`.

Refs #1032
Refs #975

## Verification

- `npm run test:calendar` - 112/112 calendar checks plus 22/22 delete-lifecycle checks.
- `npm run test:changelog` - 28/28.
- `git diff --check` - clean.

## AI assistance

This documentation was drafted with OpenAI Codex and reviewed against the merged implementation.

## Checklist

- [x] `npm test` already passed for the implementation in #1032; the affected focused suites pass here
- [x] No new frontend dependencies
- [x] No UI strings added
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -217,7 +217,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   delete reloads the visible range so an older in-flight response cannot bring the event back.
   Calendar loads now share one ordering guard, including quick out-and-back navigation, and
   concurrent deletes of one series reach the server in user-action order. This covers one event,
-  one recurring occurrence, this-and-following, and whole-series deletion.
+  one recurring occurrence, this-and-following, and whole-series deletion. A failed range load
+  also clears schedule warnings from the previous range instead of leaving stale warnings visible.
 - **Authorized caregivers can now correct a cared-for person's medication log entries** (#999).
   The log row now follows the same explicit care grant as recording, taking and skipping a dose;
   other family members still see it read-only, and scheduled entries keep their existing
--- a/public/pages/calendar.js
+++ b/public/pages/calendar.js
@@ -1331,7 +1331,15 @@ async function reloadCalendarEventsOnly() {
   }
 }
 
-/** Reconcile every calendar layer after a delayed destructive commit. */
+/**
+ * Reconcile every calendar layer after a delayed destructive commit.
+ *
+ * A delete changes only events, but this intentionally uses the full range
+ * loader: it makes events, range bounds, load/offline state and every rendered
+ * layer one authoritative snapshot after the five-second delay. Reusing the
+ * event-only loader would leave the next reader tempted to restore the old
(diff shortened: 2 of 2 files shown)


ulsklyc (approved):
Approving. A documentation-only PR still makes claims about the code, so I checked all three against it rather than reading the diff for plausibility.

**The changelog addition is accurate, and it documents something that was genuinely new.** I compared the error path of `loadRange` across the merge:

```
before #1032 (04f87334)   after #1032 (main today)
  state.loadError           state.loadError
  state.events   = []       state.events   = []
  state.tasks    = []       state.tasks    = []
  state.holidays = []       state.holidays = []
  state.scheduleEntries=[]  state.scheduleEntries = []
                            state.scheduleWarnings = []   <- new
```

So a failed range load really did leave the previous range's schedule warnings on screen before, and really does not now. That was the one improvement in #1032 that neither of us had written down.

**The comment on `reloadCalendarRangeAfterDelete` earns its place.** The function is three lines and looks like it could trivially be `reloadCalendarEventsOnly` - a delete only changes events, after all. The comment is the answer to the question the next reader will actually have, and it names the race rather than just asserting a preference.

**The historical claim checks out too.** `git log -S "recurrence_rule = newRule"` shows the client-side patch loop arriving in v1.33.0 and leaving in `81f69569`, so the comment describes something that existed rather than a hypothetical.

That last one is the only place I hesitated, because a comment about absent code is a claim that ages: it stays in the file after everyone has forgotten what it refers to, and nothing fails when it stops being true. Here it is worth it - it prevents someone from re-adding the loop as an "optimisation" - but it is the kind of line that should be short and factual, which yours is.

Measured on `63b7ae6f`: `test:calendar` 112 + 22, `test:changelog` 28/28, `test:frontend-audit` 360/360, no conflicts against `main`.

**One scheduling note, not a criticism:** the release cadence guard classifies by path, not by content, and `public/pages/` is track 1. So even a comments-only change to `calendar.js` rides the Tuesday interface train rather than going out on its own. That is the guard being blunt rather than wrong, and it costs nothing here since #1032 itself is on the same train.

I will merge this with the Tuesday batch.

PechJiri:
Apologies — an automated cleanup agent closed this PR and deleted its head branch by mistake. That cleanup action was incorrect.

I have restored the branch at the exact original head commit (`63b7ae6fbe0c273bb518ffe468cb2a654f7a3fd3`) and reopened the PR. No PR content was changed.

PR #1040, opened 2026-09-06
test(i18n): catch a counting key that has no plural variant (#1010)

### LINKED ISSUE #1010: test:i18n-plural only checks one direction: a counting key with no variant is invisible to it

ulsklyc:
Closes #1010.

`test:i18n-plural` only walked from the variant to the base key. That catches "variant without base" and is blind to the mistake people actually make: writing `{{count}}` and forgetting the `_one` variant. Two PRs shipped three such keys each on 2-3 September with green CI - the submit button read *"1 Dateien hochladen"*.

## Why a map instead of a strict assertion

The stock carries **70** such keys. A guard that fails 70 times on its first run gets weakened rather than satisfied, and then it checks nothing again. So the stock is frozen in `PLURAL_EXCEPTIONS` and every **new** counting key without a variant is red from now on.

Note the number: #1010 measured **67** on 3 September. It is 70 today. The stock grows while nothing holds it.

## The categories say *why*, and they come from the caller

The issue's own warning was worth heeding - both keys it first listed as "n=1 cannot happen" turned out to be reachable once the call site was read. So each entry is classified by what the caller does, not by how the string sounds:

| Category | Meaning |
|---|---|
| `NO_NOUN` / `PARENTHETICAL` / `ABBREV` | the German string is number-neutral (`"3 ausgewählt"`, `"Dateikonflikte (3)"`, `"3 Min."`) |
| `PAIR_LEGACY` | hand-built pair from before `_one` existed - correct today, older spelling |
| `NOT_A_COUNT` | `count` carries a formatted string, not a count - `health.meds.doseQty` receives `fmtNum()` output |
| `GUARDED` | n=1 excluded at the caller, with file and line - `subscriptions.everyCycle` is the clean case (`cycle_interval === 1 ? … : …`) |
| `TODO_ONE` | a real gap: n=1 is reachable and the sentence is wrong there. Frozen, not fixed here |
| `DEAD_KEY` | the key exists in all 24 locales and is called nowhere |

## The half that keeps the map from rotting

An entry whose key later gains a variant, or disappears, fails the test. Without that half the map only grows and nobody ever cleans it out - same shape as `INTENTIONALLY_NOT_IN_INSTALLER`.

## Measured

| Case | Result |
|---|---|
| untouched | 15 pass, 0 fail |
| a new counting key without `_one` | 14 pass, **1 fail**, names the key and says what to do |
| a mapped key given a `_one` variant | 14 pass, **1 fail**, "map is stale" |
| full `npm test` | exit 0, 280 suite links, 6058 assertions, 0 failures |

The guard found a mistake in my own map on its first run (`tasks.bulkDeleteConfirm` missing), which is exactly what it is for.

## Side finding, deliberately not fixed here

Four locale keys are dead - present in all 24 language files, called nowhere (0 hits outside `public/locales/`, not assembled dynamically either):

- `housekeeping.monthTotal`
- `housekeeping.moreWorkers`
- `tasks.overdueDay`
- `tasks.bulkDeleteConfirm`

Removing them touches 24 files that must not be reserialised, so they are recorded as `DEAD_KEY` rather than deleted in passing. The `TODO_ONE` entries are likewise left alone: adding `_one` variants means real translations in 24 languages, which is its own change.

`flatten()` became a shared `flattenLocale()` so both tests use one walker.

### DIFF
--- a/test/test-i18n-plural.js
+++ b/test/test-i18n-plural.js
@@ -153,9 +153,9 @@ test('jede Pluralvariante hat einen zählenden Basisschlüssel in allen Locales'
   for (const file of files) {
     const entries = flattenLocale(JSON.parse(readFileSync(new URL(file, LOCALE_DIR), 'utf8')));
     for (const [key, value] of entries) {
-      if (!/_(one|two|few|many|other)$/.test(key)) continue;
+      if (!/_(zero|one|two|few|many|other)$/.test(key)) continue;
       if (typeof value !== 'string' || !value.includes('{{count}}')) continue;
-      const base = key.replace(/_(one|two|few|many|other)$/, '');
+      const base = key.replace(/_(zero|one|two|few|many|other)$/, '');
       assert.ok(entries.has(base), `${file}: ${key} ohne Basisschlüssel ${base}`);
       assert.match(entries.get(base), /\{\{count\}\}/, `${file}: ${base} zählt nicht`);
     }
@@ -179,6 +179,14 @@ test('jede Pluralvariante hat einen zählenden Basisschlüssel in allen Locales'
 // Schluessel. Ein Guard, der beim ersten Lauf 70-mal scheitert, wird abgeschwaecht
 // statt erfuellt - danach prueft er wieder nichts. Die Karte friert den Bestand
 // ein; jeder NEUE zaehlende Schluessel ohne Variante ist ab sofort rot.
+//
+// BEKANNTE GRENZE: erkannt wird ein zaehlender Schluessel am `{{count}}` im WERT.
+// Ein Aufrufer darf `count` aber auch nur zur Pluralwahl uebergeben, waehrend der
+// Text andere Platzhalter interpoliert - `settings.addressbooksEnabledOfTotal` macht
+// genau das (`{{enabled}} von {{total}}`, Aufruf mit `count: addressbooks.length`,
+// sync-contacts.js:163). Solche Schluessel sieht dieser Guard nicht; sie zu finden
+// hiesse, jeden `t()`-Aufruf auf ein `count:`-Argument zu lesen. Bewusst offen
+// gelassen: die haeufige Luecke ist die hier gepruefte.
 //
 // Die Kategorie sagt, WARUM der Schluessel keine Variante braucht - und sie ist
 // am AUFRUFER belegt, nicht am Wortlaut. Ein String, der klingt, als koenne er
@@ -260,6 +268,9 @@ const PLURAL_EXCEPTIONS = {
   'subscriptions.everyCycle': 'GUARDED',
   // personal-calendar.js:214 - der Wert ist die Konstante MAX_DEFAULT_REMINDERS.
   'settings.calendarDefaultRemindersMax': 'GUARDED',
+  // subscriptions.js:150-154 - dueLabel() faengt d<0, d===0 und d===1 vorher ab,
+  // diese Zeile sieht nur noch d >= 2.
+  'subscriptions.dueInDays': 'GUARDED',
 
   // --- echte Luecken, eingefroren statt behoben ---------------------------
   // Alle unten sind bei n=1 grammatisch falsch und n=1 ist erreichbar.
@@ -288,7 +299,6 @@ const PLURAL_EXCEPTIONS = {
   'tasks.navLabelOverdue': 'TODO_ONE',              // router.js:1151, Guard ist `> 0`
   'subscriptions.listCount': 'TODO_ONE',
   'subscriptions.overdueDays': 'TODO_ONE',
-  'subscriptions.dueInDays': 'TODO_ONE',
   'subscriptions.reminderMeta': 'TODO_ONE',
   'subscriptions.metaInUseWarning': 'TODO_ONE',     // umgeht den Plural im String: „Abonnement(s)"
   'settings.recipeProviderDeleteAccountConfirm': 'TODO_ONE',
@@ -306,9 +316,13 @@ test('ein zaehlender Schluessel ohne Variante steht in der Ausnahmekarte (#1010)
   const entries = flattenLocale(localeFile('de'));
   const ohneVariante = [];
   for (const [key, value] of entries) {
-    if (/_(one|two|few|many|other)$/.test(key)) continue;
+    if (/_(zero|one|two|few|many|other)$/.test(key)) continue;
     if (typeof value !== 'string' || !value.includes('{{count}}')) continue;
-    if (entries.has(`${key}_one`)) continue;
+    // Nicht nur `has()`: ein `_one`, das null, eine Zahl oder ein leerer String ist,
+    // faellt zur Laufzeit auf den Plural-Basisschluessel zurueck (oder laesst `t()`
+    // `.replace()` auf einem Nicht-String rufen) - die Variante waere da und wirkungslos.
+    const variante = entries.get(`${key}_one`);
+    if (typeof variante === 'string' && variante.trim() !== '') continue;
     ohneVariante.push(key);
   }
 
--- a/test/test-i18n-plural.js
+++ b/test/test-i18n-plural.js
@@ -199,8 +199,12 @@ test('jede Pluralvariante hat einen zählenden Basisschlüssel in allen Locales'
  *
  * NO_NOUN        Auf die Zahl folgt kein Substantiv („3 ausgewaehlt", „3 aktiv").
  *                Im Deutschen numerusneutral, braucht keine Variante.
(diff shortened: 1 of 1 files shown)


ulsklyc:
Both reviewers found real gaps. I checked all four against the code rather than taking them on trust, and acted on three; the fourth is recorded as a known limit. Pushed as `5c18633c`.

**1. `zero` missing from the variant regex - correct, and worse than reported.** It was missing from the *pre-existing* check too, which I had copied. `resolvePluralKey()` resolves whatever `Intl.PluralRules` returns and this repo ships Arabic, so a future `key_zero` would have been read as a new base key demanding `key_zero_one`. Measured both ways: without `zero` in the regex the guard rejects a valid `probeZero_zero`; with it, the same locale passes.

**2. Validate the `_one` value, not just its presence - correct.** `has()` accepts `null`, a number or `""`, all of which fall through to the plural base at runtime (or make `t()` call `.replace()` on a non-string). Now requires a non-empty string. Measured: an empty `_one` is reported instead of silently accepted.

**3. `subscriptions.dueInDays` is `GUARDED`, not `TODO_ONE` - correct, and this one is the useful catch.** `dueLabel()` returns early for `d < 0`, `d === 0` and `d === 1`, so that line only ever sees `d >= 2`. I had sorted it by how the string reads - which is precisely the mistake #1010 warns about, just in the opposite direction from the one it describes.

That made the rest of the classification suspect, so I re-read the comparable call sites. They hold: `overdueDays` and `cycle.status.overdue` take `Math.abs(d)` with `d < 0`, so 1 is reachable; `cycle.status.inDays` is reached with `d >= 1`; `reminderMeta` reads a 0-365 column. Only `dueInDays` was wrong.

**4. Count-driven keys without a literal `{{count}}` - correct, and not fixed here.** `settings.addressbooksEnabledOfTotal` is exactly that shape: the text interpolates `{{enabled}}` and `{{total}}` while `sync-contacts.js:163` passes `count: addressbooks.length` for plural selection. This guard cannot see those. Closing that hole means reading every `t()` call for a `count:` argument, which is a different piece of work; the common mistake is the one checked here. Recorded as a documented limit above the map rather than left implicit.

Full chain after the changes: exit 0, 280 suite links, 6058 assertions, 0 failures.

ulsklyc:
Second round, both findings correct and both fixed in `d09ccd5b`.

**1. Three titles were not number-neutral.** You are right, and the distinction matters: a number in brackets is neutral only when the rest of the sentence carries no number itself.

| Key | German | at n=1 |
|---|---|---|
| `contacts.importSubmit` | `Importieren ({{count}})` | fine - stays `PARENTHETICAL` |
| `shopping.clearChecked` | `Abgehakt löschen ({{count}})` | fine - stays |
| `category.errorInUse` | `... wird noch verwendet ({{count}}).` | fine - stays |
| `folderUpload.fileConflictsTitle` | `Dateikonflikte ({{count}})` | **"Dateikonflikte (1)"** - moved |
| `folderUpload.folderConflictsTitle` | `Ordnerkonflikte ({{count}})` | **moved** |
| `folderUpload.rejectedTitle` | `Dateien, die nicht ... ({{count}})` | **moved** |

And `renderFolderUploadPreview` renders all three from a single conflict onward, so n=1 is not hypothetical.

This is the third key in this change that I sorted by the shape of the string rather than reading all of it - `subscriptions.dueInDays` was the same mistake last round. **The classification is not testable**: the map asserts that every counting key is accounted for, never that its category is the right one. So the fix is the doc comment, which now says outright that the bracket alone makes nothing neutral and that the whole string has to be read. A sharper sentence is the only guard available here, and I would rather say that than pretend a test covers it.

**2. The value check only read `de.json`.** Correct, and the gap is exactly as you describe: key parity guarantees `key_one` exists everywhere once it exists anywhere, never that it holds anything usable, so an empty `_one` in `fr` falls through to the plural precisely where nobody looks.

Now checked across all 24 locales - and **strict, not ratcheted**, because I measured the stock first: 95 variants across 24 locales, 0 unusable. There is nothing to freeze.

Measured: emptying a single `_one` in `fr.json` turns it red and names both the locale and the key (`fr.json: nav.moreBadge_one = ""`).

**On your fourth finding from round one** - a key receiving `count` only for plural selection while its text interpolates other placeholders - it stays a documented limit. I tried to quantify it and could not honestly: a line-based grep for `t('key', { ... count ... })` misses multi-line calls, which is the exact shape `settings.addressbooksEnabledOfTotal` has. So I have no number, and I have not implied one.

Full chain: exit 0, 280 suite links, 6059 assertions, 0 failures.

PR #1042, opened 2026-09-06
feat(shopping): compact rows, collapsible categories, and a self-hiding checked-items pill (#1039)

### LINKED ISSUE #1039: Shopping list UX: reduce dead space, collapse categories, improve checked-off banner and scrollbar
The shopping list becomes quite cumbersome to use once it contains a lot of items, especially on mobile.

There are four related UX issues here:

1. **A lot of vertical space is wasted**
   - Rows, section spacing, and controls are fairly tall.
   - On mobile, only a small number of items fit on screen at once, so longer shopping lists require a lot of scrolling.
   - A more compact layout would make the list much easier to scan and use while shopping.

2. **Categories cannot be collapsed**
   - Once several categories contain items, the page becomes very long.
   - It would be useful to tap a category header to collapse/expand that section, similar to collapsible task sections.
   - Ideally the collapsed state could remain while using the list so completed or currently irrelevant categories can stay out of the way.

3. **The `X items checked off` banner stays visible at the bottom**
   - After checking off items, the banner remains on screen with the Delete action.
   - On mobile this takes up a significant amount of usable space and overlaps the shopping list area.
   - It becomes especially annoying while actively going through a list because it is always present once items have been checked.
   - It would be better if it could be dismissed, auto-hide after a short period, or otherwise not permanently occupy the bottom of the screen.

4. **The scrollbar on Windows is very prominent and visually detached from the list**
   - On desktop/Windows, the shopping list shows a large native vertical scrollbar far to the right of the actual list content.
   - It stands out heavily in dark mode and makes the scrollable area feel disconnected from the shopping list.
   - A thinner/themed scrollbar, or avoiding the nested scroll area and using the normal page scrollbar, would fit the UI better.

These issues are most noticeable on mobile and Windows desktop, but the general layout could be denser and more consistent across both.

### Suggested behavior
- Allow category sections to be collapsed/expanded by tapping the header.
- Reduce unnecessary vertical padding/row height where possible.
- Make the checked-off-items banner dismissible or temporary, while keeping the Delete action accessible somewhere else.
- Style the Windows scrollbar to match the app, make it less intrusive, or avoid the nested scroll area if possible.

### Screenshots
Desktop and mobile examples showing the spacing, persistent checked-off banner, and scrollbar behavior:

Mobile shopping list

Image

mclgoerg:
Closes #1039

## What changed

**Compact rows.** A single-line shopping row was effectively 64px tall: the shared `.list-row`
grammar carries 8px of block padding around a control that already sizes itself to the 48px touch
target (`.item-check`). That padding is redundant here, so Shopping drops it (`padding-block: 0` on
`.shopping-item`) - no change to the shared `.list-row`/`.list-group` grammar in `list-row.css`, so
Pantry, Recipes, Tasks, Contacts, and Budget are unaffected. Multi-line names/metadata still grow
naturally; every control keeps its 44/48px target.

**Collapsible categories.** Each category header is now a real disclosure: an `h2` containing a
`<button>` with `aria-expanded`/`aria-controls` and the same rotating chevron Tasks already uses for
its own groups (#812). Folding toggles `[hidden]` on the existing `.list-rows` node in place instead
of rerendering, so Sortable/keyboard-reorder state and swipe closures survive a fold/unfold. New
categories are expanded by default; only the collapsed set is persisted, in `localStorage`, scoped by
household member AND by list, keyed by the category's stable ID (a rename doesn't lose the state) with
a normalized-name fallback for unknown/legacy categories, versioned, validated on read, and pruned once
list/category metadata is fully loaded. A `?highlight=<id>` deep link now expands its category first
if it was collapsed, before scrolling to the item.

**A self-hiding "checked items" pill.** The action pill (move to Pantry / delete checked) used to
call `setBulkPill()` unconditionally whenever any item was checked - including items that were already
checked when the list loaded, making it effectively permanent for such lists. It's now a small state
machine (`idle -> visible -> deferred -> suppressed`) that opens the pill for five seconds only after
a *live* check takes the count from zero to non-zero, holds it open while hovered or focused, and
suppresses it for the rest of that batch once it's been hidden once (a return to zero resets the
batch). Since Shopping has no page-unmount hook, a stale timer or a stale hover-end handler checks its
owning root's connection to the document before touching the shared bulk-pill layer - Pantry and
Contacts render their own pill through that same shell slot, so an unguarded stale callback could
otherwise clear a different module's pill after a list switch or a navigation away.

**A themed scrollbar.** `.list-scroller` (shared by Shopping, Pantry, and Recipes) gets a thin,
module-tinted scrollbar, porting Budget's existing recipe into `list-row.css` instead of duplicating
it, with an added hover state (a bare `--tint-hint` edge doesn't carry a state on its own per its own
documentation in `tokens.css`).

## Testing

- New `test/test-shopping-ux.js` (19/19): behavioral tests via a `__test` export rather than
  source-text guards - rename-safe stable keys, per-user/per-list storage scoping with safe fallback
  on malformed or version-mismatched data, pruning of orphaned keys, the toggle's DOM effects in both
  directions, and the full pill state machine including two race-safety cases (a stale list-switch
  timer, and a stale post-navigation hover-end handler) that must not touch the shared layer. Every
  guard in this suite was counter-proved: I temporarily broke the corresponding production code and
  confirmed exactly the expected test(s) failed, then restored it.
- A few structural CSS/markup assertions were added to the existing `test-shopping.js` (43/43) for the
  parts that are genuinely markup/CSS contracts (the disclosure shape, the Shopping-only padding/rhythm
  overrides, the scrollbar rules).
- `test:frontend-audit` 360/360, `test:popover-menu` 9/9, `test:sortable-reorder` 40/40,
  `test:mobile-scroll-layout` 9/9, `test:changelog` clean.
- Full `npm test`: exit 0, 5,484 tests, 0 failures.
- Verified in a real throwaway browser instance (seeded multi-category list, similar density to the
  issue's own screenshots) at 390x844 and 1856x978, light and dark: folding one category visibly
  reveals several more rows below the fold; a live check opens the pill, which auto-hides after five
  seconds without user action.

## Out of scope (per the issue)

No change to the nested list scrollport's ownership, the shared list row's global geometry, checked-
item/Pantry-transfer/drag-order/delete-confirmation semantics, or any API/schema/sync behavior.


ulsklyc (changes requested):
Read the code rather than the description, ran the suites, and tried to break each of the three parts. **No blocker.** One thing to fix before merge, two notes, and a rebase whose one conflict is the dangerous kind.

## The compact row is right, and the reasoning is the load-bearing part

This was the piece I expected to argue about, because #1039's first point runs straight at a measured invariant: the row height comes from `.list-row`, and that height *is* the touch target.

It holds up. The arithmetic, checked at the tokens:

```
.list-row      padding: var(--space-2) …   →  8px top + 8px bottom
               min-height: var(--target-lg) →  48px
.item-check    width/height: var(--target-base)
               → 44px pointer, 48px touch  (@media (hover:none), tokens.css:1663)
```

48 + 8 + 8 = the 64px you measured. Drop the padding and the row is 48px, because `min-height: --target-lg` still holds it there and the control still carries its own 44/48. **Nothing about the target changes** - and `list-row.css:266` already says why in its own comment: *"die Zielgröße gehört auf die Bedienelemente"*. You did not shrink a target, you removed padding that duplicated one.

Scoping it to `.shopping-item` rather than `.list-row` is the right call for exactly the reason your comment gives: Shopping is the one list whose row contains its own 48px element.

## To fix before merge: the comment says "moved", the code duplicates

The scrollbar block says *"dieselbe Rezeptur wie `.budget-list` (budget.css), hierher verschoben statt dupliziert"*. Measured: `budget.css` is untouched, and its `.budget-list::-webkit-scrollbar` rules are still there. So the recipe now exists in two places, and the comment claims the opposite - which is the version a future reader will believe when they change one of them.

Either move it for real (drop the budget.css rules and let `.budget-list` use the shared class), or say plainly that it is a second copy and why. Cheap either way; the wrong comment is the expensive part.

**Related note, not a defect:** `.list-scroller` is used by Shopping, **Pantry and Recipes** (measured across `public/pages/`), so this scrollbar change lands in three modules. #1039 came from one. That is defensible for a shared class - I would just say so in the changelog entry, since a Pantry user will see it without having asked.

## The other two parts, and how hard I pushed

| Probe | Result |
|---|---|
| `saveCollapsedCategories()` made a no-op | 15 pass / **4 fail** |
| `aria-expanded` no longer set on toggle | 17 pass / **2 fail** |

Both anchored. The tests read the real `toggleCategoryCollapse()` and assert `aria-expanded`, `[hidden]` and the chevron together, which is the combination that actually breaks in isolation.

Two design decisions worth naming because they are easy to get wrong and you got them right: folding toggles `[hidden]` on the existing node instead of rerendering, so Sortable and swipe state survive - and the collapsed set is keyed by category **ID** with a normalised-name fallback, so a rename does not silently drop the state.

The pill state machine (`idle → visible → deferred → suppressed`) is more machinery than I would have guessed the problem needs, but I checked the bug rather than taking your word for it, and it is real: `updateCheckedActions()` (shopping.js:1466) branches on `checkedCount` alone, and it is called from the render path at `:1294` - so a list that *loads* with checked items shows the pill immediately, with nothing on the way to close it. Holding it open while hovered or focused is the part that makes it not annoying.

## Rebase: one conflict, and it is the chain trap

Your branch sits on `3d82ebd5`; `main` is now `3490a179`. Exactly one conflict, in `package.json` - and it is the one that fails silently:

```
main: 293 test: scripts   PR: 289   union must be 294
only on main: the five schedule-* suites (from #1022, merged today)
only in PR:   test:shopping-ux
```

Taking "ours" drops your new suite; taking "theirs" drops five. **The resolution is the union, never a side.** `test:suite-chain` will catch a script that exists but hangs in no chain - it will not catch one that was dropped from both.

No overlap with #1045, so the two can go in either order.

## Verified

`test:shopping` 43/43, `test:shopping-ux` 19/19, `test:frontend-audit` green, `test:suite-chain` 4/4 with the new suite correctly in the chain. No new `innerHTML`, no colour literals, no new locale keys (the toggle takes its accessible name from the category label, icons are `aria-hidden`).

`test:document-guards` I started but did not finish - Sonde 1 alone runs 96s and I stopped waiting. Its Sonde 4 is the documented touch-target check, so it is the one that would confirm the arithmetic above empirically; worth naming in your final run rather than only `npm test`.

ulsklyc:
Follow-up: I said I had started `test:document-guards` and stopped waiting. It finished, so here is the result rather than leaving my review with an open end.

**Sonde 4 passes** - *"eine Reihe traegt ihre Dichte, ein Einzelziel ist allein treffbar"*, green after 312s. That is the documented touch-target probe, measured on the real document rather than on the stylesheet, and it is the one that could have contradicted the arithmetic in my review. It does not.

For completeness, the whole file on your branch: Sonde 1 ✔ (96s), Sonde 2 ✔ (458s), Sonde 3 ✔ (132s), Sonde 4 ✔ (312s). Roughly sixteen minutes, which is why it is a `test:document-guards` handrail rather than part of `npm test` - worth knowing if you plan to run it before the rebase.

So the compact row is confirmed from both directions: the token arithmetic says the target never shrank, and the probe that measures rendered geometry agrees. Nothing changes in my review; the two points from it still stand (the "moved vs. duplicated" comment, and the `package.json` union).

ulsklyc:
Correction to my follow-up above. The verdict does not move, but two numbers in it were wrong and one of them is a planning number you might act on.

**The run is now actually finished, and it is green.** `test:document-guards` on `pr1042` (`effe3adb`): **35 tests across 11 suites, 35 pass, 0 fail**, not a single `✖` in the log. Sonde 4 is green, so the token arithmetic in my review and the rendered geometry still agree. That part of the follow-up holds.

**What was wrong is that I wrote it before the run had finished.** I read the `✔` lines that had already scrolled past in a still-running log and called them the result.

- I listed Sonden 1 to 4 as "the whole file". The file carries **21 named probes, 35 tests**. I quoted the first four and presented them as all of it.
- I said "roughly sixteen minutes". Those four are sixteen minutes. **The file is 79** (`duration_ms 4724282`, so 78 min 44 s on this machine). If you were going to run it before the rebase on the strength of my number, you would have budgeted a fifth of what it needs.

Where the hour actually goes, in case you want to run only part of it:

| Probe | Time |
|---|---|
| Sonde 19 - module head in the regular size class | 640 s |
| Sonde 14 - icon on tinted ground holds 3:1 | 537 s |
| Sonde 2 - WCAG AA on the composited background | 458 s |
| Sonde 7 - a row sequence is not a column of cards | 397 s |
| Sonde 18 - nothing operable under the FAB at scroll end | 370 s |
| Sonde 4 - target sizes | 312 s |
| everything else | ~2010 s |

The expensive ones are expensive because each runs twice. Sonde 19 and 20 go per locale (`de`, `uk`), Sonde 5 per locale too but `de` against `ar`, so the RTL side is measured rather than assumed. Sonde 4 and 18 go per device world, since `--target-base` switches on `(hover: none)` and a probe that measures one world checks half a rule.

Nothing in the review itself changes. The two points from it are still open: the "moved vs. duplicated" comment, and the `package.json` union.

mclgoerg:
Addressed the review comment and rebased onto current `upstream/main` (now includes Schedule v3 / PR #1022) — new tip is `e505d154`, no merge conflicts reported.

**Review fixes:**
- Corrected the `.list-scroller` comment in `public/styles/list-row.css` — the Budget scrollbar rules still live in `budget.css`, so the shared recipe was duplicated/adapted, not moved.
- Updated the changelog entry to disclose that the shared `.list-scroller` styling is visible in Shopping, Pantry, and Recipes, while compact-row/category behavior stays Shopping-only.

**Rebase:** `package.json`'s test-chain conflict resolved as a union — every script already on `main` (including the five Schedule v3 suites) plus `test:shopping-ux`, verified against `test:suite-chain`.

**Tests:** `test:shopping`, `test:shopping-ux`, `test:frontend-audit` all green. Full `npm test`: 5588 tests, 5586 pass, 0 fail (2 pre-existing skips). Browser smoke-tested: compact rows keep their touch targets, category disclosure/persistence still works, the checked-items pill still self-hides, and the shared scrollbar renders correctly in Shopping, Pantry, and Recipes.

**`test:document-guards` (not part of the `npm test` chain, run separately since this PR touches shared list-row CSS that Documents' guards also exercise):** 34/36 pass. Two failures, both unrelated to this PR's diff (which only touches `shopping.js`/`list-row.css`/`shopping.css`) and reproducible on plain `upstream/main` before this branch's changes are applied:
- Sonde 3 (button-shape guard): `.cal-task-chip--{high,low,medium,urgent}` on the Calendar page carry their own shape outside the four exempt categories.
- Sonde 18 (FAB overlap guard, desktop): two FABs detected over content instead of one — the dashboard speed-dial isn't docking/collapsing as expected.

Flagging these since they surfaced during this PR's verification, not to expand its scope — happy to open separate issues for them if useful.

mclgoerg:
Addressed both points from the review, rebased onto current upstream/main, and re-ran the suite.

## Fixed before merge

**The scrollbar comment** (`list-row.css`) no longer claims the Budget recipe was moved. It now says
it was ported here as a second, adapted copy, and that `budget.css` kept its own (unchanged) copy -
matching what's actually in the tree. No change to `budget.css` itself.

**Changelog** now says explicitly that the compact-row and category-folding behavior is Shopping-only,
while the shared `.list-scroller` styling (thin, module-tinted scrollbar) is visible in Shopping,
Pantry, and Recipes as well, since those three modules share the scroll container.

New commit: `e505d154` (on top of the original `22959334`, rebased onto `upstream/main` at `a3c77e8a`).

## Rebase

One conflict, in `package.json`, exactly where you flagged it. Resolved as a union, not a side: kept
all five Schedule v3 suites from main (`test:schedule-feed`, `test:schedule-reminders`,
`test:schedule-extras`, `test:schedule-custom-fields`, `test:preferences-schedule-templates`) and kept
`test:shopping-ux` from this branch, in its original position right after `test/test-shopping.js`.
Diffed the resolved chain array against the full `test:*` script keyset both ways before committing -
nothing dropped from either side. GitHub now reports this PR as mergeable, no conflicts.

## Tests

- `test:suite-chain` 4/4
- `test:shopping` 43/43
- `test:shopping-ux` 19/19
- `test:frontend-audit` 361/361
- Full `npm test`: 5588 tests, 5586 pass, 0 fail, 2 skipped, exit 0
- `test:document-guards`: 34/36 pass. The 2 failures are pre-existing on `upstream/main` and unrelated
  to this PR - confirmed by diff (this branch touches zero files/selectors either failure is about):
  - Sonde 3 flags `.cal-task-chip` (Calendar priority chips) as an unregistered button shape; traced
    to #1028 (calendar month titles on phone), which added the chip variants without a `SHAPE_EXEMPT`
    entry.
  - Sonde 18 (desktop) measures 2 floating FABs instead of 1 (only the Dashboard speed-dial should
    float); some other module isn't docking, most likely introduced by the Schedule v3 merge (#1022)
    given the file scope. Neither touches Shopping/Pantry/Recipes or anything in this branch's diff.

## Manual verification

Seeded a fresh throwaway instance and drove it with headless Chromium: compact 48px rows keep their
48px `.item-check` touch target, the category disclosure toggles `aria-expanded`/`[hidden]` together
and its collapsed state survives a reload, the checked-items pill appears on a live check and
self-hides after ~5s, and `.list-scroller` renders its themed thin scrollbar (`scrollbar-width: thin`)
in Shopping, Pantry, and Recipes. 9/9 checks passed.

ulsklyc (approved):
Approving. Both points from the review are closed, and both are the kind that only a reader can verify, so I read them.

**The comment no longer claims a move.** `list-row.css` now says the recipe was taken here as a second, adapted copy and that `budget.css` kept its own unchanged. I checked that the claim is true rather than just better worded: this PR does not touch `budget.css` at all, and its four `scrollbar-thumb` rules are still there. A comment that describes another file is a copy and ages like one, so it mattered that this one now describes what is actually in the tree.

**The changelog now draws the line where the code draws it:** Shopping-only for the compact rows and the folding, while the shared `.list-scroller` styling is named as visible in Shopping, Pantry and Recipes too, "since those modules use the same scroll container". That is the sentence a household reads before wondering why their Pantry looks different.

Also checked: `package.json` is the union - 294 scripts, nothing from `main` lost, `test:shopping-ux` added.

Merging with the interface release. #1039 closes with it.

ulsklyc:
**Merged**, `18ed383c` on `main`. #1039 closed with it.

One thing you should know, because it happened after your last push and I resolved it rather than sending it back: **#1027 landed first and your `package.json` started conflicting.** Both PRs append a suite to the same aggregate `test` chain, so the two versions of that one line disagreed.

That is precisely the trap from the review, arriving from the other direction. Resolved as the union, verified rather than eyeballed:

```
HEAD (main, incl. #1027)   292 links, only it has  test:preferences-countdown-grace
your branch                292 links, only it has  test:shopping-ux
merged                     293 links, both present, 0 duplicates
```

Worth restating why taking a side would have been quiet rather than loud: `test:suite-chain` catches a script that hangs in **no** chain. It cannot catch one that was dropped from a chain while still existing as a script, which is exactly what picking one side produces.

Full chain on the merge result: **exit 0 read from the log file, 287 suite links, 5597 + 709 assertions, 0 failures**, `test:suite-chain` green.

Both fixes you made were the right ones. The comment in `list-row.css` now describes what is in the tree - I checked that this PR really does not touch `budget.css` and that its four `scrollbar-thumb` rules are still there - and the changelog draws the Shopping-only line where the code draws it.

PR #1045, opened 2026-09-06
fix(calendar): give overlapping Schedule blocks their own lane (#1043)

### LINKED ISSUE #1043: [Bug] Overlapping shifts with the same start time are not displayed correctly
### Yuvomi version

v2.64.1

### Deployment

### Description

When two shifts/events on the same day start at exactly the same time but have different end times, only the description of the shorter/first shift is visible in the calendar.

The second shift appears to be positioned underneath the first one and its title/description cannot be read. The only indication that a second shift exists is its different calendar color.

This occurs in both the Day view and Week view.

Screenshot

Image

### Steps to reproduce

Create a shift/event for a day from 08:00 to 12:00
Title/description: Schultag Basti
Create a second shift/event on the same day from 08:00 to 12:45
Title/description: Schultag Emma
Open the calendar in Day view or Week view.

Actual behavior

Only Schultag Basti is visible.

Schultag Emma appears to be rendered underneath the first shift, so its title and description cannot be read.

The different color of the second shift makes it possible to see that another event exists, but the event itself is not identifiable from the calendar.

Expected behavior

Both overlapping shifts should be fully visible or otherwise clearly distinguishable in the calendar, including their titles/descriptions.

The user should be able to identify both:

08:00–12:00 – Schultag Basti
08:00–12:45 – Schultag Emma

The exact visual implementation is not important, as long as overlapping events do not completely obscure each other and their titles/descriptions remain accessible.

### Browser / device / OS

Browser: Firefox Operating system: Windows Affected views: Day view and Week view

### Logs

```shell

```

### Additional context

The issue seems to occur specifically when two events have the same start time but different end times. Events with different colors make the overlapping event detectable, but its title and description are not readable.

mclgoerg:
Closes #1043.

## Root cause

Ordinary calendar events avoid overlapping via `layoutOverlaps()`, a lane-assignment pass that gives each overlapping event its own column. Timed Schedule entries in the calendar's "blocks" display mode never went through it: `renderScheduleTimeBlock()` gave every block the same fixed full-width bounds, so two shifts with the same or overlapping time range painted directly on top of each other - only the last one drawn was readable, with just its color hinting a second entry existed underneath. Reproducible on `main` today with two household members each holding a shift on the same day (no Schedule v3 features needed).

## Fix

`public/pages/calendar.js`:
- Factored `layoutOverlaps()`'s grouping/lane math into shared `overlapGroups()`/`assignLanes()` helpers.
- Added `layoutScheduleBlocks()`, the Schedule-side equivalent - same algorithm, but keyed by the entry *object* rather than an id. Schedule entries don't reliably carry one (pattern-sourced entries only have `pattern_id`+`position`), and two entries can legitimately share a shift type and time on the same day, so object identity is the only key that can't collide.
- `renderScheduleTimeBlock()` now accepts a computed lane and derives `left`/`width` from it, reusing each view's existing outer gutters (Week: 2px both sides; Day: 4px/10px) - a block with no overlap keeps exactly its old full-width bounds.
- Both Week and Day view now compute and pass a schedule layout alongside the existing `layoutOverlaps()` call for ordinary events.

## Tests

10 new tests in `test/test-calendar.js`: the reported `08:00-12:00`/`08:00-12:45` pair, touching vs. partially-overlapping ranges, a three-way overlap, duplicate-content entries (same shift type + time, different source), and the rendered Week/Day HTML itself no longer using the old fixed full-width bounds. Counter-proved by temporarily reverting the fix: all 10 new tests failed and nothing else did.

`npm run test:calendar` 122/122, `test:frontend-audit` and `test:schedule` green, full `npm test` exit 0.

## Manual verification

Reproduced and confirmed fixed on a throwaway instance built from this branch (screenshots below) - two overlapping shifts ("Schultag Basti" 08:00-12:00, "Schultag Emma" 08:00-12:45) now render as separate, fully readable lanes in both Day and Week view, matching the issue's exact reported case.

### DIFF
--- a/public/pages/calendar.js
+++ b/public/pages/calendar.js
@@ -3350,6 +3350,8 @@ export const __test = {
   layoutScheduleBlocks,
   scheduleBlockTimeRange,
   renderScheduleTimeBlock,
+  renderWeekView,
+  renderDayView,
 };
 
 function renderAgendaEvent(ev, dayStr) {
--- a/public/pages/calendar.js
+++ b/public/pages/calendar.js
@@ -2235,12 +2235,44 @@ function renderScheduleChip(entry, className = 'allday-holiday') {
   const start = type.start_time ? '<small class="schedule-entry__start">' + esc(type.start_time) + '</small>' : '';
   return `<div class="${className} schedule-entry" style="--holi-color:${esc(type.color)}" title="${esc(scheduleEntryTitle(entry))}"><span>${esc(label)}</span>${scheduleEntryExtraBadge(entry)}${start}</div>`;
 }
-function renderScheduleTimeBlock(entry, className) {
+/**
+ * Zeit-Fenster eines Schichtplan-Blocks in Minuten seit Mitternacht - dieselbe
+ * Dauer-Arithmetik wie zuvor inline in renderScheduleTimeBlock() (Mindestlaenge
+ * 30 Min., Ende-vor-Start heisst "geht ueber Mitternacht"), jetzt eigenstaendig,
+ * damit layoutScheduleBlocks() dieselbe Zeitspanne fuer die Ueberlappungs-Logik
+ * kennt wie die Darstellung selbst.
+ */
+function scheduleBlockTimeRange(entry) {
   const type = entry.shift_type;
   const start = timeToMinutes(type.start_time);
   const end = timeToMinutes(type.end_time);
   const duration = Math.max((end > start ? end : 24 * 60) - start, 30);
-  const bounds = className === 'week-event' ? 'left:2px;width:calc(100% - 4px);' : 'left:calc(4px);width:calc(100% - 14px);';
+  return { start, end: start + duration };
+}
+
+/**
+ * Spalten-Position eines gelegten Blocks in CSS umrechnen - dieselbe Formel wie
+ * renderWeekEvent()/renderDayEvent() fuer gewoehnliche Termine, aber mit den
+ * Schichtplan-eigenen Aussenraendern (Woche: 2px beidseitig; Tag: 4px links /
+ * 10px rechts, damit die Stunden-/Jetzt-Linie hinter dem Balken sichtbar bleibt).
+ * Ohne Layout (Fallback) ergibt sich exakt der alte volle Balken (1 von 1 Spalte).
+ */
+function scheduleLaneStyle(layout, gutter) {
+  const cols = layout?.totalCols ?? 1;
+  const idx = layout?.colIndex ?? 0;
+  return {
+    left: `calc(${(idx / cols) * 100}% + ${gutter.left}px)`,
+    width: `calc(${100 / cols}% - ${gutter.total}px)`,
+  };
+}
+
+function renderScheduleTimeBlock(entry, className, layout = null) {
+  const type = entry.shift_type;
+  const start = timeToMinutes(type.start_time);
+  const end = timeToMinutes(type.end_time);
+  const duration = Math.max((end > start ? end : 24 * 60) - start, 30);
+  const gutter = className === 'week-event' ? { left: 2, total: 4 } : { left: 4, total: 14 };
+  const { left, width } = scheduleLaneStyle(layout, gutter);
   // Sichtbar statt nur im title: .schedule-time-block traegt pointer-events:none
   // (es ist eine Hintergrundflaeche, echte Termine malen darueber, siehe die
   // Begruendung an der Klasse in calendar.css) - ein Hover erreicht dieses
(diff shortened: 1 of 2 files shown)


mclgoerg:
Summary for reviewers: this is a standalone fix cut fresh from `main` (not stacked on #1022), since the bug already exists in the shipped Schedule module and shouldn't wait on the v3 PR.

- `test:calendar` 122/122 (10 new), `test:frontend-audit` and `test:schedule` green, full `npm test` exit 0.
- Every new test was counter-proved: reverting the fix makes exactly those 10 fail and nothing else.
- Manually verified in a real browser on a throwaway instance built from this branch - both overlapping shifts render as separate, fully readable lanes in Day and Week view (screenshots in the PR description).

ulsklyc (changes requested):
Read the code rather than the description, ran the suites, and tried to break the refactor on purpose. The diagnosis is right, the fix is the right shape, and there is **one should-fix** plus a rebase you cannot avoid.

## The refactor is equivalent, and I checked it properly

Pulling `overlapGroups()`/`assignLanes()` out of `layoutOverlaps()` touches the path every ordinary calendar event takes, so "it still passes" is not enough. I extracted both versions - `layoutOverlaps` from `3d82ebd5` and yours - and ran them against the same inputs.

**306 sets, 0 differences.** That includes the cases the boundary actually turns on: exactly touching ranges, chains of them, identical ranges, three-way overlaps.

My first attempt at this probe was worthless and I nearly reported it as proof. It used random durations, which almost never produce an exactly touching pair - so when I deliberately broke the new version (`end <= range.start` → `end < range.start`), it still reported 0 differences. A probe that cannot fail proves nothing. With touching edges generated on purpose, the broken version shows 2 differences and the real one still shows 0. That is the number I trust.

## Should-fix: no test reads the caller

The layout is computed in `renderWeekView`/`renderDayView` and handed down as a third argument. **Nothing checks that it arrives.**

```
assignLanes() forced to one lane for everyone   ->  116 pass / 6 fail
both call sites stop passing the layout          ->  122 pass / 0 fail   <-- silent
```

The second one is the regression that matters: drop `scheduleLayouts[i].get(entry)` and `scheduleLayout.get(entry)`, and every block falls back to `layout = null`, which is by design exactly the old full-width bar. #1043 is back, and the suite stays green.

Your ten tests are good, but all four HTML ones call `renderScheduleTimeBlock()` with a layout you construct in the test. They prove the renderer honours a layout, never that the view computes and passes one.

You have closed this exact hole before: it is the same shape as the `getRangeForView` gap in #1030, where you added a test that stubs `matchMedia` and calls the real caller. One test that renders a week with two overlapping shifts and asserts the two blocks get different `left` values would cover it.

## The rebase, and why it is not mechanical

Your branch is based on `3d82ebd5`; `main` is now `3490a179` and carries Schedule v3 (#1022, merged today). Two files conflict, three places in total, and the first needs judgement:

- **`calendar.js`, 45 lines** - both sides rewrote `renderScheduleTimeBlock()`. You changed where the block sits (`left`/`width` from the lane); v3 changed what is inside it (two fixed single-line rows, time and custom-field value sharing the second). **They are compatible** - position and content - but the merge has to keep both, and taking either side whole loses the other.
- **`test/test-calendar.js`, 203 lines** - both appended tests at the same spot. Union.
- A third, small one in the `__test` export block.

Worth doing on top of current `main` rather than resolving at merge time, because v3 also added `scheduleOverlayMeta()` to that same function and your gutter arithmetic should be measured against the version that ships.

## Verified

`test:calendar` 122/122 and `test:frontend-audit` green on `ba0ed3a8`. No new `innerHTML`, no colour literals, no new locale keys. The lane fallback really is the old bar: with `layout = null` the style computes `calc(0% + 2px)` / `calc(100% - 4px)`, arithmetically identical to the previous `left:2px`.

Keying the schedule layout by object identity rather than an id is right, and the comment explaining why (pattern-sourced entries only carry `pattern_id`+`position`, and two entries can legitimately share type and time) is the kind of note that saves the next reader an hour.

ulsklyc:
One number in my review was short, in your favour. I wrote **`test:calendar` 122/122**; the script is two files, and the second one also ran green:

```
node ... test/test-calendar.js  &&  node --test test/test-calendar-delete.js
      122 tests, own harness              22 tests, node --test
```

So it is **144/144** on `ba0ed3a8`, not 122. The two halves report in different formats, I took the tally from one of them and put the script's name next to it.

Nothing follows from this for your side: both counter-proof numbers came from the same first half and are unaffected (`116 / 6` when `assignLanes()` is forced to a single lane, `122 / 0` when both call sites stop passing the layout, which is still the silent one). The should-fix and the rebase note stand as written.

mclgoerg:
Addressed both points from the review round and rebased onto current `upstream/main` (now includes Schedule v3 / PR #1022) — new tip is `37bbfee1`, no merge conflicts reported.

**Caller-level coverage (the should-fix):** exported `renderWeekView`/`renderDayView` through `__test` and added two new tests that render overlapping Schedule blocks through the *real* view functions (not a directly-constructed layout), asserting the resulting blocks land at distinct `left` positions. Counter-proven by temporarily stripping the lane argument at both `scheduleLayouts[i].get(entry)` / `scheduleLayout.get(entry)` call sites: exactly these two new tests fail, all 131 others stay green, then reverted before committing.

**Rebase:** two real conflicts in `public/pages/calendar.js` (`renderScheduleTimeBlock` — kept both v3's two-row content via `scheduleOverlayMeta()` and this PR's lane-derived `left`/`width` positioning, same gutters as before) and `test/test-calendar.js` (union of both branches' test additions), plus the small `__test` export list conflict. `test:calendar` is now 133/133 (was 122 before v3 added its own tests).

**Browser verification:** stood up an isolated throwaway instance from this branch (fresh DB, Schedule module enabled, two shift types 08:00–12:00 and 08:00–12:45 assigned to two different household members via `PUT /api/v1/schedule/overrides/:dateKey`) and drove it with Puppeteer. Both the reported pair and the merged v3 two-row content (title + time/overlay) render correctly side-by-side in both Week and Day views — screenshots posted in chat, not attached here.

**Tests:** `test:calendar` 133/133, `test:frontend-audit` 361/361, `test:schedule` 60/60, full `npm test` 5569 tests / 5567 pass / 0 fail (2 pre-existing skips).

ulsklyc (approved):
Approving. The should-fix is closed, and I reproduced your counter-proof rather than reading it.

`renderWeekView`/`renderDayView` are exported through `__test` and the two new tests drive the real view functions, so the gap I described - the renderer honours a layout, nothing proves the view computes and passes one - is now covered where it has to be.

I stripped the lane argument at both call sites myself (`calendar.js:2387` and `:2644`), on a `cp` backup rather than a checkout, and ran `test:calendar`:

```
untouched                       133 pass / 0 fail
lane argument removed at both   131 pass / 2 fail
```

The two that fall are exactly your two new ones, by name. That is what a load-bearing test looks like, and it is the same shape as the `getRangeForView` test you wrote for #1030.

Also checked: `package.json` keeps all 293 scripts from `main` with nothing lost in the rebase.

Merging with the interface release.

ulsklyc:
**Merged**, `be002758` on `main`. #1043 closed with it.

Clean merge, no conflicts, and the caller-level tests you added went in with it. Thank you for running the counter-proof yourself and describing it precisely enough that I could repeat it in two minutes instead of reconstructing it.

PR #1052, opened 2026-09-07
docs: add RepoCloud one-click deploy button

cosark:
Added a "One-Click Deploy" section with a RepoCloud deploy button to the README, placed between the Install and Before you commit sections.

 - New `## ☁️ One-Click Deploy` section added after Install
 - Button links to https://repocloud.io/details/yuvomi/

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,12 @@ covers engine setup, HTTPS, backups and troubleshooting step by step.
 
 ---
 
+## ☁️ One-Click Deploy
+
+[![Deploy on RepoCloud](https://d16t0pc4846x52.cloudfront.net/deploylobe.svg)](https://repocloud.io/details/yuvomi/)
+
+---
+
 ## Before you commit
 
 **What if this project stops?** Nothing changes on your machine. It is MIT-licensed and


ulsklyc:
@cosark Thank you for the clean, minimal patch, and for linking a listing that actually exists rather than a placeholder. I checked it properly instead of reacting to the idea, so here is everything I found, including the objection I expected to raise and could not.

## One thing blocks it, and it is mechanical

`npm run test:readme-consistency` goes red on your branch:

```
✖ README.md und README.de.md haben dieselbe Struktur
  README.de.md hat 7x "h2", README.md hat 8x. Die Uebersetzung folgt der
  englischen Fassung; ein Abschnitt fehlt oder ist doppelt.
```

The two READMEs are held to the same structure by a guard, so a new `##` section in one needs the same section in the other. Adding the equivalent block to `README.de.md` fixes it. That is the only hard failure; the other 16 assertions in that suite pass.

## What I checked and found in your favour

**The badge does not leak anything.** My first concern was the image coming from `d16t0pc4846x52.cloudfront.net`. This project went to some trouble to embed the GitHub star count at build time precisely so that opening a page sends nothing to a third party, and an external image looked like the same thing coming back in. It is not: I rendered your markdown through GitHub's own API and the `src` comes back rewritten to `camo.githubusercontent.com`. GitHub proxies it, so no reader's IP reaches CloudFront. The README is also not embedded anywhere on the website, so there is no second path. That objection is off the table.

**Both targets are live** (HTTP 200 each), and the listing's description is accurate: it says 19 modules, and the README table has exactly 19 rows. I counted rather than assumed.

**And the scope document does not forbid this,** which is worth saying because it would have been the easy thing to reach for. `docs/SCOPE.md` has a "No hosted demo" entry with quite absolute language, but it is about a public instance run by *this project*, filled with a fictional family's data. Your button is a reader deploying their own instance on infrastructure they rent. Different thing. Using that entry against you would have been a misreading.

## What is left is an editorial decision, and it is the maintainer's

Three things I would want him to weigh rather than decide myself:

**The placement works against the text below it.** The button lands directly above the "Before you commit" section, whose opening paragraph says *"It is MIT-licensed and self-hosted, there is no server of ours anywhere in the path"*. A one-click deploy to a commercial cloud immediately above that reads oddly, even though the two are compatible in substance. Under the Install section would sit better.

**It would be a first of one particular kind.** I nearly told you this would be the README's first external image, and that would have been wrong: there are already twelve, the shields.io badges in two HTML blocks (lines 15 to 18 and 199 to 206), which my first pattern missed because they are `` tags rather than markdown. So an externally hosted badge is established practice here and no argument against you.

What would be new is narrower, and it holds when you separate links from images: the README's six external `href` targets are GitHub (4), the project's own site (1) and @Kyrodan's community documentation (1). `shields.io` appears only as an image source, never as somewhere a reader is sent. So this would be the first link pointing a reader at a commercial service. That is a threshold, and a fair thing for a maintainer to decide once rather than drift past.

**The listing is not ours to keep true.** Its version, description and pricing can change without anybody here noticing. Today they are right, and I say that having checked rather than assumed. But a permanent button points at a page somebody else maintains, and the project already spends real effort keeping its store listings accurate by hand. If this goes in, it is worth deciding who checks it and how often, rather than discovering a stale claim from a user's bug report.

So: the German README is required either way. The rest is Ulas's call, and I would rather hand him an accurate picture than dress a preference up as a rule. Thanks again for a patch that was easy to review.

ulsklyc:
@cosark Thank you for waiting, and for the patch. I have decided not to add the button to the README, and I want to give you the rule rather than a mood, because the rule is what will answer the next provider too.

The README is the project's landing page, and every link on it today sends a reader to the project itself or to community documentation. A deploy button for a hosting service would be the first link that sends a reader to a commercial offer, placed a few lines above the sentence that says there is no server of ours anywhere in the path. The two are compatible in substance, but the README is where that promise is made, and I would rather keep it unmixed.

The second half is maintenance. A button is a standing claim about a listing this project does not control: its version, price and description move without anyone here noticing, and I already keep three store listings accurate by hand. I will not add a fourth that I cannot edit.

This applies to every provider equally, so it is not about RepoCloud. Your listing is welcome to exist, the description there is accurate today, and anyone searching for a managed way to run Yuvomi will find it. If this question comes up a second time I will write the rule into `docs/SCOPE.md`, so it is answered in one place.

Closing this without merging. Thanks again for a patch that was easy to read.

PR #1055, opened 2026-09-08
fix(calendar): preserve recurring occurrence identity

PechJiri:
## Summary

Closes #975.

PR3 from the sequence discussed in #975, after #1032 and #1037. Eligible local recurring events now store edited occurrences as linked replacements rather than unrelated detached rows. An edited occurrence retains its series scopes and inherits fields that were not deliberately changed.

**Final-review candidate: `9b76fc0f`.** This branch includes the maintainer's `pr1055-rebase` integration at `0433cc12`, the subsequent author-owned fixes from `5f48975e`, and `main` at `b8ce81bf` (including the #1089 end-time representation fix). Both histories are preserved; no force push was used. Earlier verification statements about `c62ebe36` are historical, not the evidence for this candidate.

## Identity and model boundary

- `recurrence_parent_id` identifies the recurring master.
- `recurrence_id` identifies the original recurrence slot, not the moved display date. The integrated timezone calculation keeps this aligned with the stored UTC-day identity when local and UTC days differ.
- `overridden_fields` records which values the replacement owns; other values inherit from the current master occurrence.

Migration **194** appends the columns, parent/slot unique constraint, and lookup indexes, with the matching test mirror. The parent foreign key cascades with the series.

This parent-plus-slot model subsumes imported `uid::recurrenceId` identity, but the current day-based key allows **one overridden original slot per series per day**. Supporting multiple imported occurrences on one day would require widening the key, index, and validation first. This PR does not migrate imported overrides or guess historical detached edits back into a series.

Schedule custom-field values and the computed Waste calendar layer proposed in #1063 remain separate, domain-owned mechanisms. This PR neither materializes those layers into `calendar_events` nor implements the Waste module. The boundary is recorded in `docs/SPEC.md`.

## Mutation behavior

- **Only this:** atomically upsert the linked replacement and exclude its original slot. Re-editing updates the same child. Deleting removes linked state and retains the exclusion.
- **This and following:** truncate the master, create a successor, retain the remaining `COUNT`, and reparent compatible future replacements by original slot.
- **Whole series:** retain existing visibility-based authorization and refresh inherited projections and ACL state. Deletion-only EXDATEs survive rule changes, so changing daily -> weekly -> daily does not resurrect deleted occurrences.

[...]

PR #1044, opened 2026-09-06
feat(notes): add category management UI

PechJiri:
## Summary

Adds the browser UI for the note-category data model and API merged in #991. This is a clean reconstruction on current `main`, intentionally excluding the stale branch's duplicate schema, core Notes API and permission-engine commits.

Merge this PR after #1055. The latest follow-up merges `main` at `4ba47c0c`. #1055 itself now has upstream conflicts and is not included in this branch; combined runtime verification must be refreshed after it is updated/merged (details in the latest review comment).

## Changes

- add personal and household category management to Notes
- add an accessible category picker to the note editor
- filter notes by the intersection of selected categories
- add category-aware Notes widget options and badges on the dashboard
- expose household-category capability controls in the permissions UI
- add complete UI copy in all 24 supported locales
- add regression coverage for picker identity, lifecycle updates, permissions, dashboard filtering and saved widget options

## Verification

- `npm test`
- `git diff --check`
- JavaScript syntax checks for every changed or added JS file
- earlier combined integration test; refresh required after #1055 is updated/merged against current `main`

## Checklist

- [x] `npm test` passes
- [x] Follows [CONTRIBUTING.md](../CONTRIBUTING.md) conventions
- [x] No new frontend dependencies (vanilla JS, no frameworks)
- [x] UI strings use `t('key')` (no hardcoded text)
- [x] CHANGELOG.md updated (if user-facing change)

Drafted with OpenAI Codex.

PR #1095, opened 2026-09-10
feat(waste): Waste collection module — manual schedules, ICS import, reminders, Dashboard/Calendar integration (#1063)

### LINKED ISSUE #1063: Waste collection: an optional module for pickup dates, built by @mclgoerg
Accepted in #1033, where @mclgoerg offered to build it and @4bjd25469m-collab asked for it. Filed so the accepted scope has a place. The build is his; this issue holds what was agreed, so the review has something to check against.

## What this is

An optional core module for household waste collection: named waste types with icon and colour, pickup dates, the next pickup per type on a dashboard tile, and a switchable layer in the family calendar.

## Decided

- **Core module, off by default** (households opt in, the way Inventory does since migration 145). A third-party module cannot add a calendar layer - the layers are enumerated in `public/pages/calendar.js:2868-2889` - and the layer is half the value.
- **Calendar layer after the Schedule pattern** (`docs/DECISIONS.md` entry 6): computed on read, never materialised; the calendar fetches a range (`GET /schedule/entries?from&to`, `calendar.js:1240`) and renders a layer it can switch off, it does not own the rows. No rows in `calendar_events`, no hidden ICS subscription. Tile and layer are two views of one table.
- **Moving or skipping a single pickup is in step 1** (an override table like `schedule_overrides`); a computed rhythm is wrong exactly around holidays otherwise.
- **ICS file import is in step 1**, using the existing parser `server/services/ics-parser.js` (RFC 5545, time zones, RRULE expansion). Preview of every distinct summary before writing; the person maps each to a type or ignores it; unknown summaries are never mapped silently. The source, the mapping, the external UID and the import time are stored so next year's file updates the same source.
- **"No upcoming pickups for this source, import the new file"** is part of step 1: a stale file has to show itself.
- **Monthly "n-th weekday" rules are deferred.** `server/services/recurrence.js` knows `DAILY/WEEKLY/MONTHLY`, `BYDAY` as a weekday filter, `INTERVAL`, `UNTIL`/`COUNT` and `BYMONTHDAY=-1`, not an ordinal `2MO`. "Every n weeks on a weekday" and "fixed day of month" are enough for step 1; if the ordinal comes, it goes into `recurrence.js` once, for every caller.
- **Dashboard tile: the next pickup per type**, so a frequent type never hides a rare one; follows the list-tile rules (`listRowCap`, `dashboard.js:1063`) and "Customize dashboard".
- **Reminders and push are not in step 1.** The reminder centre has the pattern when it is time (migrations 140, 162, 184).
- **The filter axis per calendar/subscription (#987, filed as #1064) stays separate.** The waste layer shows or hides all pickups at once; which external calendars to show is a different question.
- Permissions per module and widget (migration 74); API under `/api/v1/<module-id>` with an OpenAPI entry.

## What a core module brings with it

- Every string through `t()`, and **all 24 locale files with real translations**: `test:i18n-translated` lets the count of untranslated values only fall. The module name per language follows the existing `nav.*` vocabulary.
- README table and the module count ("nineteen" appears three times; `test:readme-consistency` holds table and number together), a section in `docs/SPEC.md`, a line in `docs/test-suites.md`, an own `test:<id>` suite.
- Migrations append-only. 190 is taken by #1055; the numbers are claimed in the thread when the PR is ready.
- No `innerHTML`, tokens from `tokens.css`, honest empty states.

## Still open

Module id and English name. Proposal: `waste` / "Waste collection", German "Entsorgung". @mclgoerg picks and says so in #1033, so the locale keys are right from the start.

Origin: #1033.

mclgoerg:
Closes #1063

## Summary

An optional, off-by-default **Waste collection** module (accepted in Discussion #1033): manual pickup rhythms with per-occurrence overrides, mapped ICS import (file and automatic URL subscriptions), per-user reminders, a Dashboard widget, a Calendar layer, and interoperability/fine-grained controls — plus a follow-up audit that found and fixed 5 blockers, 6 High, 12 Medium, and 7 Low findings against the module before this PR.

Rebased onto current `main` and squashed into one commit; after the round-3 review, `main` (through `e0c4bf26`) is merged in and the migrations are renumbered **197–207** (194 landed on `main` as #1055; 195/196 are promised to #1109/#1144). 103 files changed, +17,698/−184.

## Features, by phase

**Phase 1 — domain foundation.** `waste_types`, `waste_schedules`, `waste_schedule_overrides`, `waste_one_off_pickups`. A pure domain service (`server/services/waste-domain.js`) adapts the existing shared `recurrence.js` (also used by Tasks/Calendar/CalDAV/Outlook) rather than reimplementing recurrence math, bounds any range read to 731 days, and coalesces schedule/one-off/import origins into one occurrence DTO with full provenance preservation — a move or skip is recorded as an override keyed on the schedule, never mutates history.

**Phase 2 — manual module vertical slice.** Full CRUD: types (6 icon/color presets, archive instead of hard-delete while referenced, reorder), schedules nested under a type (weekly or monthly-fixed-day, pause, optional end date), a per-occurrence move/skip/restore menu, one-off pickups. Registered end-to-end (scopes, permissions, module registry, OpenAPI, `?type=&date=` deep links). All 24 locales carry real per-language translations, not copies.

**Phase 3 — ICS file import.** Upload an annual municipal waste calendar, preview per-label mapping (including inline "create a new type"), resolve blocking diagnostics (e.g. an unbounded recurrence) with an explicit skip, commit atomically with a digest+version concurrency guard. Re-import diffs additions/changes/removals against the previous snapshot. Imported pickups are a third occurrence origin, coalesced the same way as manual ones — Waste never writes into `calendar_events`.

**Phase 4 — Dashboard widget.** Opt-in, hidden by default. One row per active type (next date, "moved"/"coalesced" flags, a needs-refresh warning), deep-linking into the Waste page.

**Phase 5 — Calendar layer.** Device-local, off by default. Pickup chips render in month/week/day/agenda, never person-filtered (household-wide), clicking routes to the Waste page rather than opening an event editor. Its own load path, isolated from the main Calendar fetch.

**Phase 6 — accepted-core milestone audit.** A full pass against every invariant in `PLAN.md` (single occurrence DTO, centralized permission enforcement, all bounded-work limits, source-isolation cascade design) before building further — no discrepancies found.

**Phase 7 — automatic ICS URL sources.** A subscribed URL is fetched on a schedule (60 min – 30 days, conditional GET, SSRF-guarded, size-capped, timed out) and auto-commits only when every label already has a *remembered* mapping decision and no diagnostic is unresolved; otherwise it pauses for a manual, reviewed refresh through the same mapping wizard Phase 3 built.

**Phase 8 — per-user reminders.** Personal, opt-in: choose which types to be reminded about, how many days ahead (0–14), and what time of day. Anchored on the same coalesced-occurrence identity the rest of the module uses, so a moved pickup gets its old reminder cleaned up and a new one created rather than a duplicate.

**Phase 9 — shared advanced recurrence.** Ordinal-weekday monthly schedules ("2nd Monday", "last Friday") added once to the shared `recurrence.js` engine — every existing consumer (ICS import/export, CalDAV, Outlook, Tasks, Calendar) benefits automatically, with zero changes needed in any of them.

**Phase 10 — interoperability and fine-grained controls.** A revocable, read-only per-user ICS feed (optionally scoped to a subset of types) for subscribing Waste pickups into any external calendar app. Mapping-profile export/import lets a household copy its own label→type decisions between sources (no shipped municipal catalog). Per-type Calendar visibility and a Dashboard widget type filter. The waste type catalog is indexed in the global search.

## Independent audit and fixes (post-acceptance, pre-merge)

Before opening this PR, the module was put through an evidence-based validation against a deployed test instance — automated checks plus live browser testing across viewports, themes, locales, and permission levels (admin / read-only / denied) — followed by four more rounds of user-reported issues from real mobile/desktop use, and a final pass closing every remaining Medium/Low finding from the original report.

**5 blockers fixed**, including: a schedule with an old anchor date (a real ICS-import shape) could silently return **zero** occurrences with no error instead of throwing; every occurrence row's action menu rendered at viewport origin `(0,0)` and intercepted clicks meant for other rows; read-only members saw fully interactive write controls that only failed at the server; an ICS event just after local midnight imported one day early; CI never actually ran a single Waste test because an unrelated suite aborted the chain first.

**6 High findings fixed**, mostly around URL-source refresh concurrency (an unthrottled retry loop on a parse/commit failure, a lost-update race between a manual refresh and the scheduler, an optimistic-concurrency check that read outside its own transaction) and an N+1 query on the Dashboard's critical path.

**User-reported issues fixed** across three more rounds: a header with 4 actions that didn't fit any width; a calendar filter toggle that spuriously asked to discard changes it couldn't have made; a native `<dialog>` (icon picker, avatar cropper) pinned to the top-left app-wide because a global CSS reset defeated the browser's own centering; an icon picker showing an unrelated starter set; an import-source panel whose action row overflowed by up to 256px (present at every width, not just mobile).

**12 Medium and 7 Low findings closed** in a final pass: a private IP address leaking into a read-only member's view via an error message; a lone-CR ICS injection vector; import diagnostics that printed hardcoded English text next to a translated label in 23 of 24 locales (now genuinely translated in all 24); orphaned reminder rows and stale feed-type references left behind on type deletion; three un-transacted multi-step writes; a wrong code comment that told a future maintainer a load-bearing safety flag wasn't; badge text under the 4.5:1 AA contrast floor at its actual (not assumed) background; a page that ignored the router's own navigation-lifecycle contract; a route bypassing the shared data-access layer; a non-timing-safe token lookup; a keyboard-inaccessible custom control; a missing database index for a hot query path; and more (full list and rationale for the handful of findings left deliberately unchanged in `STATUS.md`/PR discussion).

## Test coverage

- **264 focused Waste tests** across 9 dedicated suites (`npm run check:waste`): domain logic, migrations (fresh schema + upgrade + constraints + cascades), routes, UI, import, Dashboard widget, Calendar layer, URL sources, reminders.
- Every regression suite this module touches re-verified: `test:frontend-audit` (363/363), `test:i18n`/`-plural`/`-translated` (all 24 locales, including a guard that checks nav labels match API-token-scope labels), `test:module-registry-parity`, `test:openapi-structure`/`-coverage`, `test:permissions`/`-routes`, `test:ssrf`, `test:reminders`/`-routes`, `test:notifications`, `test:suite-chain`, `test:changelog`, `test:readme-consistency`, `test:docs-landing`.
- **Full `npm test` (300+ suites) exits 0.**
- Live browser verification at every stage against a real deployed instance: activation, full manual CRUD, ICS import/mapping/commit (file and URL), Dashboard widget, all 4 Calendar views + deep links, reminders end-to-end with a real reload round-trip, the responsive matrix (390/768/1024/1440 × light/dark), reduced motion, and — for the audit rounds — admin/read-only/denied roles, real DST-boundary math, and adversarial SSRF/XSS/CSRF probes.

## Notes for review

- Migration numbers were originally assigned 190–200 against `upstream/main` at `3017d0b2`; by the time this PR was opened, `main` had moved 122 commits and independently claimed 190–193 for unrelated features. Rebased and renumbered to 194–204 before opening this PR; after the round-3 review they moved again to **197–207** (`main` landed 194 as #1055, and 195/196 are promised to #1109/#1144). Until those two land, `test:migrations-append-only`'s gapless check reports the reserved 194→197 gap as its one finding — the base-prefix and description checks stay green, and the gap closes by itself when 195/196 merge.
- The rebase also surfaced two things two newer `main` guards now check that didn't exist at this branch's fork point — both fixed as part of this PR, not left for CI to catch: the aggregator for the 9 Waste suites is now `check:waste` and is deliberately NOT in the root `test` chain — the nine individual `test:waste-*` scripts run there once. An earlier revision of this description claimed `test:suite-chain` forces the aggregator into the chain; that was wrong (`test-suite-chain.js` only collects keys starting with `test:`, `check:release-cadence` being the precedent), and keeping `test:waste` in the chain ran all nine suites twice per `npm test`; and 13 places in `waste.js` re-rendered after an `await` inside a modal-close handler without calling `refocusAfterRender()`, per the newer focus-restore guard — fixed with the same call already used in `budget.js`/`calendar.js`/`documents.js`.
- No shipped municipal ICS catalog — import/mapping is entirely the household's own data, by design (see `PLAN.md`).
- Nothing in this PR touches `calendar_events`; Waste occurrences are computed at read time from `waste_schedules`/`waste_one_off_pickups`/`waste_imported_pickups` and never materialized as calendar rows.


ulsklyc (changes requested):
Thanks for this, it is a large and carefully built module. Review of head `73e8abaf`. The migration renumbering (195-205) from my earlier comment still applies and is not repeated here.

A lot held up when I read it: recurrence goes through `recurrence.js` instead of a second implementation, URL fetches have a timeout, a size limit and the connect-time SSRF guard, and the reminder sync checks both the module switch and module rights. One finding blocks, and four more should be fixed before merging.

## Blocking

### A new URL source can stay in "needs mapping" for good

Preview and commit fetch the URL separately: `reimport/preview` and `reimport/commit` in `server/routes/waste/sources.js` both call `icsTextForReimport()`. `commitImport()` then rejects the commit when the SHA-256 of the raw text differs from the preview's (`server/services/waste-store.js:642`, `computeDigest()` in `waste-import.js:45`), and the client sends `preview_digest` for URL sources too (`public/pages/waste.js:1100`).

For a feed whose bytes change on every request, the two fetches never match. A `DTSTAMP` that carries the time of the request is enough. Then:

1. A new URL source has no remembered mappings, so `refreshUrlSource()` leaves it at `needs_mapping = 1` with `next_attempt_at = NULL` (`waste-url-source.js:281-291`).
2. The scheduler skips such sources (`listDueUrlSources()`, `waste-store.js:544`), and the Refresh button runs the same `refreshUrlSource()` into the same state.
3. The mapping wizard's commit fetches again and gets 409 "The file content changed since you last previewed it", on every attempt.
4. Mapping rows are only created inside `commitImport()` (`waste-store.js:754`), so neither `PUT /:id/mappings/:mappingId` nor the mapping-profile import (which reports `unmatched_pattern` without a row) can get the source out. What remains is deleting it and importing the file by hand.

The test for the two-attempt path (`test/test-waste-url-source.js:284`) does serve different content per request, but it commits through `store.commitImport` with the first fetch's text, so the route's second fetch is never exercised.

A digest over what the preview actually shows (labels, events, diagnostics) instead of the raw bytes keeps the guard's purpose, committing what was reviewed, without depending on `DTSTAMP`. Keeping the previewed text server-side and committing exactly that would work as well.

## Should fix

### A token can read the subscription URL and `last_error` without write access

`hasWasteWriteAccess()` (`server/routes/waste/helpers.js:44`) checks only the token scope for API tokens. `tokenAllows()` returns `true` for an unscoped token (`server/scopes.js:174`), while module rights apply to token subjects as well (`applyRoleModuleAccess()` in `server/auth.js`). So an unscoped token, or one with `waste:write`, issued for a member with read-only Waste access gets `url` and `last_error` unredacted from `GET /api/v1/waste/sources`. The same member sees them redacted in the app, and the same token gets 403 on every write. Checking `moduleAccessVerdict(req.sessionModuleAccess, 'waste', 'write')` for tokens too, on top of the scope, closes it.

### `WASTE_SOURCE_ALLOW_PRIVATE_NETWORK` is not documented

`server/services/waste-url-source.js:29` introduces it, but it is missing from `tools/installer/env-schema.js` (the source of truth for env variables), from `.env.example` and from the `*_ALLOW_PRIVATE_NETWORK` list in `test/test-installer-schema.js`. `ICS_SUBSCRIPTION_ALLOW_PRIVATE_NETWORK` is in all three. Without the variable every `http://` URL is rejected (`:52-54`), and an operator with a feed on the LAN has no documented switch. The installer guard cannot catch this by itself, because the name reaches `readPrivateNetworkOptIn()` as a string constant, not as a literal `process.env` read.

### The module count is only partly updated

The module table now lists Waste, but `README.md:10` still says "Nineteen modules", `README.de.md:10` "Neunzehn Module", and `docs/index.html` says 19 / nineteen / neunzehn in seven places (for example `:1010`, `:1109`, `:1181`, and the two translation blocks from `:1606`).

### References to a PLAN.md that is not in the repository

19 added lines point at a PLAN.md ("invariant #8", "invariant #9", phase numbers): `public/pages/calendar.js` (6), `test/test-waste-calendar.js` (3), `server/services/waste-store.js` and `waste-domain.js` (2 each), and one each in `waste-url-source.js`, `waste-reminders.js`, `routes/waste/schedules.js`, `routes/waste/pickups.js`, `public/pages/dashboard.js` and `test/test-waste-ui.js`. Nobody reading the code or a failing test can look those up. The rules belong in `docs/SPEC.md` or in the comment itself.

## Nice to have

- **COUNT is no longer enforced for ordinal rules.** `lastOccurrenceOf()` in `server/services/recurrence.js` now returns `null` when `bydayOrdinal` is set, so `nextOccurrenceAfter()` has no end for a synced `FREQ=MONTHLY;BYDAY=2MO;COUNT=6`, and once the sixth occurrence has passed, the Key dates tile (`countdowns.js`) names a seventh. The bound is computable: a single ordinal token under MONTHLY has exactly one occurrence per interval, the same kind of reasoning the `BYMONTHDAY=-1` branch right below uses.
- **UTF-8 is decoded per chunk** (`waste-url-source.js:127`, `text += chunk.toString()`). A multi-byte character on a chunk boundary becomes two U+FFFD, and a label like "Grünschnitt" then no longer matches its remembered mapping. Collecting the buffers and decoding once avoids it. `ics-subscription.js` on main has the same pattern, so this one is not yours alone.
- **The SSRF pre-check is a copy.** By its own comment it mirrors `ics-subscription.js#checkSSRF` exactly, which main exports, and the two differ only in which opt-in they read. Taking the opt-in as an argument would leave one copy. The connect-time guard is already shared, so there is no gap today.
- **The URL source scheduler ignores a disabled module.** `runDueWasteSourceRefreshes()` fetches and commits due sources even when the household has switched Waste off. `waste-reminders.js` checks `disabled_modules`, the scheduler does not.
- **The Waste suites run twice in `npm test`:** the chain calls the nine `test:waste-*` scripts and then `test:waste`, which runs the same nine again.

ulsklyc (commented):
Round 2, reviewed at `c21cca96`. Thanks, this is a thorough round. I checked each fix by taking it out again and running the suite.

## The blocker is fixed

Comparing what the preview showed instead of the raw bytes is the right cut, and checking it against the commit's own preview (`waste-store.js`, after `previewImport`) keeps the guard's purpose. With the new test fed a `DTSTAMP` that really differs between the two fetches, it is green on this head and red with the raw-bytes digest put back.

**But the test as committed does not show that.** The feed builds its stamp from `new Date(Date.now() + call)` and then cuts the milliseconds (`test/test-waste-url-source.js:393`). `call` adds 1 and 2 ms, so both fetches inside the same second return identical bytes, and the test also stays green with the old digest (measured: 20/20 with `digest: computeDigest(icsText)` in `buildImportPreview`). Making the bytes differ for sure, for example `Date.now() + call * 1000` or the counter written into the stamp directly, makes it bind: with that change the old digest turns exactly this test red. The companion test for a real date change is fine as it is.

## Should-fix items: all in

- **Token access.** With the old scope-only line back in `hasWasteWriteAccess()`, the new redaction test in `test:waste-routes` goes red.
- **Private network switch.** It is in `env-schema.js`, `.env.example`, the installer schema list, the compose doc and the Unraid template; `test:installer-schema` is green.
- **Module count and landing page.** Fixed in both READMEs and on the page. Adding the card rather than only changing the number is the better fix, thanks for following the guard there.
- **PLAN.md references.** None left.

## Nice-to-haves

- COUNT for ordinal rules, UTF-8 decoding once, the shared SSRF check with the opt-in as a parameter, and the scheduler checking the module switch are all in. The UTF-8 and the scheduler change are both bound: taking either out turns its new test red.
- Keeping the double `test:waste` run because of `test:suite-chain` is fine with me.

## Before merge

- **No CI run exists for `c21cca96`.** GitHub created none for this push, neither waiting for approval nor failed, so there is nothing for me to approve. In the meantime the branch conflicts with `main` in `docs/test-suites.md` (from #1128, merged after your push), so the next push has to bring in `main` anyway, and that push will get a run. Locally, the full chain on this head: green, except `test:changelog`, which fails only because this branch is still on 2.65.1 while the `v2.65.2` tag exists. Merging `main` clears that; the rest of the chain after it (194 steps) is green.
- **Migrations 194-204 move to 195-205** once #1055 lands, as you planned. #1055 needs one more merge with `main` first, so this is not today.

ulsklyc (changes requested):
Round 3. **The round-2 finding is closed** - and I verified it rather than taking the description: reinstating `digest: computeDigest(icsText)` at `waste-import.js:353` turns the suite to 19/1 with `actual: 409, expected: 200` at `test-waste-url-source.js:419`, and neutering the guard itself costs the other test its 409. So it binds in both directions, and your `call * 1000` reasoning holds.

Both merge findings you reported yourself are real and complete. The refocus guard reaches `waste.js` - deleting the `refocusAfterRender()` you added after `deletePickup()` gives 383/1 naming `waste.js:1788` - and your three fixes are the full set the scanner finds. Thank you for flagging them instead of letting me find them.

Three things block.

## Blocking

**1. `public/pages/calendar.js:1170` - the filter counter is inverted for Waste.**

```js
if (wasteEnabled() && !state.layerWaste) n += 1;
```

The Waste layer is opt-in and defaults to **off** (`:627`, `:1640`). Measured through the real `__test.activeFilterCount()` in the browser loader:

| state | count |
|---|---|
| module on, untouched defaults | **1** |
| user switches the layer **on** | **0** |
| module off (control) | 0 |

Inverted. Every member of a household that enables Waste gets `cal-toolbar__filter-btn--active`, the "1" badge and `aria-label calendar.filtersActive {count:1}` from first load, while nothing is being hidden. Worse: `#cal-filters-reset` (`:3457`) explicitly sets `state.layerWaste = false` - the documented default - so **"Reset filters" cannot reach 0**.

The two neighbours avoid this on purpose: `layerBirthdays`/`layerSchedule` default to *on*, and holidays are gated `hp.holiday_show_public && !state.layerHolidays`, i.e. counted only when the household has the layer on and this device turned it off. Waste has no such household gate. The file states the rule itself at `:1445`: *"sonst bliebe ein Filter aktiv, der nichts wegnimmt, und der Zaehler am Knopf loege."*

Fix: drop line 1170, keep 1171 (a chosen type subset does hide something).

**2. Migration 194 is taken, and one half of the collision merges silently.**

`main` landed 194 today as `72af8302` (#1055). #1109 -> 195 and #1144 -> 196 are promised, so Waste needs **197-207**. Renumbering is my job per CONTRIBUTING, but I am flagging it because one part does not announce itself: on the real merge, `server/db.js` and `server/db-schema-test.js` conflict **visibly**, while `test/test-search.js` merges with **no conflict** and then execs `MIGRATIONS_SQL[194]` twice (`:39` calendar, `:45` waste) - the second throws `duplicate column name: recurrence_parent_id` and `waste_types` is never created.

Full blast radius on this head, so nothing is missed: `server/db.js:7605`-`:8067` (11 `version:` fields) plus comments at `:7815`/`:8074`; `server/db-schema-test.js:1116` + `:1132`; `test/test-search.js:35,37`; `test/test-search-index-duplicates.js:58,63`; **`test/test-waste-migrations.js:109,113,114,121`** (this one selects off `MIGRATIONS`, not `MIGRATIONS_SQL`, and went 37/1 when I renumbered `db.js` alone); `docs/SPEC.md:1503, 2051, 2119`; `docs/test-suites.md:284`; comments in `search.js:205`, `waste-domain.js:199`, `waste-ics.js:20`, `waste-reminders.js:12`, `waste-store.js:87`.

**3. The branch is `CONFLICTING`/`DIRTY`** - the green checks are bound to `baseRefOid 2c852f1e`, ten commits behind. Four files conflict: `package.json` (main gained `test:calendar-occurrence-overrides` and `test:sync-lock`; union, no choice), `server/db.js`, `server/db-schema-test.js`, and `server/services/ics-export.js:8` where main adds `localToUTC, shiftDateKey, utcToWall` and the PR adds `timingSafeEqual` - **taking either side alone is a `ReferenceError`**.

I resolved all of this in a throwaway clone to measure the rest, and on the resolved merge 30 suites are green including all nine waste suites. So it is mechanical - but `test/test-search.js` needs reading, not merging.

## Should fix

- **`waste-import.js:181-186`/`:208` - the fixed "ICS event just after local midnight imported one day early" blocker has no binding test.** Counter-proof: insert `timeZone = 'UTC';` as the first line of `buildImportPreview`, reinstating the pre-fix behaviour, and all nine waste suites - 264 tests - exit 0. No waste test passes a non-UTC `timeZone`, and `test-waste-reminders.js:51` pins `household_timezone` to `'UTC'`. The one zone-looking test (`test-waste-import.js:152`, `TZID=Europe/Berlin:20260701T060000`) passes either way because 06:00 is nowhere near midnight. Real repro: `TZID=Europe/Berlin:20260701T000000` gives `date_key 2026-06-30` at UTC vs `2026-07-01` at Europe/Berlin. `docs/test-suites.md:230` currently claims this axis is covered.
- **`package.json:308` - the nine waste suites run twice** in every `npm test` and CI run (individually, then again via `npm run test:waste`). The PR says `test:suite-chain` forces this; it does not - `test-suite-chain.js:21` only collects keys starting with `test:`, and `check:release-cadence` at `:334` is the existing precedent. Counter-proved: renaming to `check:waste` and dropping it from the chain leaves `test:suite-chain` 4/0 green and saves ~11s per run.
- **`public/pages/waste.js:764`** - `return target.toISOString().slice(0, 10);` in `nearestOrdinalAnchor()`, and the value becomes the schedule's `anchor_date`. CLAUDE.md bans exactly this spelling. The value happens to be *correct* here (`target` is built via `Date.UTC` from `todayKey()`), but the file imports `todayKey, addLocalDays` and not `toLocalDateKey`, and the guard that would catch it (`test-frontend-audit.js:487`) is a hardcoded two-file list. Route it through the helper, or add `waste.js` to that list so the next edit cannot silently become the real bug.

## Nice to have

- `package.json:308` - the `test:schedule` entry is glued onto the same physical line as `test`. Valid JSON, but a formatting regression that makes an already-conflicting hunk bigger.
- `server/routes/preferences.js:486` - `normalizeWidgetOptions` was widened to "or any finite number" for **every** widget to carry `wasteVisibleTypeIds`. `Number.isInteger(v)` admits exactly what Waste needs instead of letting `1e308` and floats into `sync_config` app-wide.
- `public/pages/dashboard.js:2209` - the Waste widget's overflow line reuses `dashboard.countdownMore`. Nothing breaks (the text is generic and has its `_one`), the key just names the wrong widget.
- `public/sw.js:236` - `/pages/waste.js` is not in `PAGE_MODULES`, so the page will not open offline. Consistent with `housekeeping` and `schedule`, so not a regression - but worth a deliberate yes/no rather than an omission.
- `public/styles/tokens.css:544` - column alignment differs from its neighbours.

The module itself is in good shape - the domain split, the ICS import and the reminder path all read well, and nine dedicated suites is the right amount of care for a new module. Get the filter counter and the renumber done and this is close.

ulsklyc (commented):
Round 4, reviewed at `e3f74201`, merged locally with current `main` (`f254b57c`, three commits newer than your merge). That merge is clean: no conflicts (`CHANGELOG.md`, `docs/test-suites.md` and `test/test-frontend-audit.js` auto-merge).

**All three round-3 blockers are fixed, and each fix is bound by a test.** I checked every one by putting the old code back and running the suite.

## Round-3 blockers: verified

**1. Filter counter** (`public/pages/calendar.js:1178`). The off-default no longer counts, and the type-subset line stays. `test/test-waste-calendar.js:204` drives the real `activeFilterCount()` through all three states. Counter-proof: with `if (wasteEnabled() && !state.layerWaste) n += 1;` put back, the suite goes to 30/1 with `unangetastete Vorgaben duerfen keinen Filter zaehlen: 1`. Restored, it is 31/0. `test:calendar` stays green either way (152 + 25), so this test is the only one that binds the counter, and it does.

**2. Migration numbers.** `server/db.js` declares 194 (main), then 197-207 for Waste, with no duplicates. The `version: 197` entry is the types table, 206 the FTS triggers and 207 the covering index. The mirror keys in `server/db-schema-test.js` are 194 / 197 / 206. `test/test-search.js:39,45,47` execs `[194]`, `[197]` and `[206]`, and `test/test-search-index-duplicates.js:63` uses `[197, 206]`. `test/test-waste-migrations.js:113-121` selects `< 197` / `=== 197`. On the merge `test:search` is 22/0. Counter-proofs:
- `test/test-search.js:45` back to `MIGRATIONS_SQL[194]` makes it throw `duplicate column name: recurrence_parent_id`, the exact silent-merge failure from round 3. So the suite would catch a regression.
- Waste's first entry in `db.js` back on `version: 194` makes `test:waste-migrations` fail with `UNIQUE constraint failed: schema_migrations.version`.

I also compared every renumbered reference with its old number at `478e6488`. All moved by exactly +3 except one, see Should fix.

**3. Branch state.** `MERGEABLE`, and it still merges cleanly with today's `main`.

**Round-3 should-fix items:**
- **ICS midnight** (`test/test-waste-import.js:165`). With `timeZone = 'UTC';` as the first line of `buildImportPreview` the suite goes to 27/1 (`the household-local day, not the UTC day before (2026-06-30)`). Restored, it is 28/0. It binds now.
- **Double run.** I parsed the root `test` chain: 315 steps, 315 unique, no duplicates. The nine `test:waste-*` scripts appear once each, and `check:waste` is not in the chain. `test:suite-chain` is 4/0.
- **`toISOString().slice`** is gone from `waste.js`. The guard binds the spelling: with `return target.toISOString().slice(0, 10);` put back into `nearestOrdinalAnchor()`, `test:frontend-audit` goes to 385/1 naming `waste.js`.

## Suites on the merge (`e3f74201` + `f254b57c`), each run on its own

| suite | result |
|---|---|
| test:waste-domain | 56/0 |
| test:waste-migrations | 38/0 |
| test:waste-routes | 57/0 |
| test:waste-ui | 30/0 |
| test:waste-import | 28/0 |
| test:waste-dashboard | 23/0 |
| test:waste-calendar | 31/0 |
| test:waste-url-source | 20/0 |
| test:waste-reminders | 11/0 |
| test:calendar | 152/0 + 25/0 |
| test:search | 22/0 |
| test:search-index-duplicates | 9/0 |
| test:frontend-audit | 386/0 |
| test:i18n | 54/0 |
| test:i18n-translated | 92/0 |
| test:openapi-structure | 11/0 |
| test:openapi-coverage | 3/0 |
| test:changelog | 28/0 |
| test:readme-consistency | 17/0 |
| test:suite-chain | 4/0 |
| test:schema-reconcile | 4/0 |
| test:migrations-append-only | 2/1 (gap 194 -> 197, expected) |

I also ran the whole root chain on the merge, every one of its 315 steps separately so an early failure could not hide the rest: **314 exit 0, one exit 1** (`test:migrations-append-only`, the gap above). Nothing else fails.

## Blocking

None on your side.

## Should fix

**`server/db.js:8062`: one migration reference was missed by the renumber, and after merge it can never be fixed.** Migration 204's comment says:

```js
// through its own ON DELETE CASCADE (same hazard as migrations 52/199/200).
```

At `478e6488` this read `52/199/200` as well. In the old numbering that meant the URL-sources rebuild and the reminders migration, the two Waste entries with `foreignKeysOff: true`. Those are now **202 and 203**. The new 199 (ICS import sources) and 200 (mappings) have no `foreignKeysOff`, so the comment now points a reader at the wrong migrations. Please make it `52/202/203`.

This is Should fix and not Nice to have because of timing. `test-migrations-append-only.js:246-259` compares the MIGRATIONS block as text, comments included, against the merge base. Once this lands, correcting the comment turns that guard red for whoever does it. Right now it is a one-line change.

## Nice to have

- **`test/test-calendar.js:2144-2147`: a comment describes the bug you just fixed as current behaviour.** It says `layerWaste` must be `true` in `basis`, "sonst zaehlt Wastes eigener Opt-in-Default (aus) hier faelschlich als zusaetzlicher Filter". Since `7de80c16` that is no longer true: the off-default counts nothing. The `layerWaste: true` itself is harmless, but the reason given is now wrong. Drop the sentence, or say the value is only there to keep the test independent of the Waste axis.
- **`server/routes/preferences.js:489`: `Number.isInteger(v)` is not bound by a test.** With it mutated back to `Number.isFinite(v)`, `test:preferences-dashboard` stays 17/0. `test/test-preferences-dashboard.js:434-443` covers `[1, 2, 3]`, mixed lists and `[Infinity]`, but no fraction. One assertion like `withOptions({ list: [1.5] })` -> 400 would bind the round-3 change.
- **`public/pages/waste.js:762-770`: the local-date rewrite of `nearestOrdinalAnchor()` is only guarded textually.** The frontend-audit guard catches the `toISOString().slice` spelling, but nothing runs the function. It is nested inside `openScheduleModal()`, so the move from `Date.UTC` getters to local getters, which is exactly where a western zone would shift a day, has no behavioural test. Hoisting `nthWeekdayOfMonthLocal()` plus the anchor choice into an exported helper and testing it under a non-UTC `TZ` would close that. Not required: the code reads correct (local midnight in, `toLocalDateKey()` out, no mixed getters).

## Delta since round 3

I read the five commits after `478e6488` plus the merge resolution:
- No `innerHTML`/`outerHTML` assignments.
- No new UI text outside `t()`.
- No design values outside `tokens.css`; that file only has the column-alignment change.
- No `toISOString().slice(0, 10)` left.
- `dashboard.wasteMore` and `dashboard.wasteMore_one` are present in all 24 locale files, with no `[de: ...]` placeholders.
- The `ics-export.js:8` resolution keeps both `timingSafeEqual` and `localToUTC, shiftDateKey, utcToWall`.
- The `sw.js` decision comment reads as a decision now, thanks.

## Info: merge order (not on you)

CI on `e3f74201` is red on both Node versions for one reason only. `test:migrations-append-only` reports "Zwischen 194 und 197 fehlt mindestens eine Nummer" while its base-prefix and description checks pass (2/1). That is the reserved 195/196 gap. #1144 (now 195) and #1109 (now 196) have to land on `main` before this PR; the gap then closes without any change here.

One side effect to be aware of: that guard is step 3 of 315 in the `&&` chain, so CI has not run the other 312 steps on this head. The local run above stands in for it until the gap closes. The next CI run after #1109 and #1144 land is the first real CI signal for this branch.

ulsklyc (approved):
Round 5, reviewed at `32b8e518`. **Approved.** The round-4 should-fix and all three nice-to-haves are in, and the #1044 merge is resolved on both sides.

## Where I measured

For the first time, the gap guard could run for real. I built a throwaway merge in landing order: current `main` (`fb0311ab`), then #1144 (195, approved), then #1109 (196, approved), then this head. The conflicts with those two were append-only, and I resolved them the way they will land: `server/db.js` 194 -> 195 -> 196 -> your 197-207, `db-schema-test.js` keys 197/206 after 196, `test:recurrence` and the nine `test:waste-*` steps in the chain at your positions, the Waste bullet at the end of `### Added`.

On that merge, each suite on its own:

| suite | result |
|---|---|
| `migrations-append-only` | **3/0** - 194 through 207 without a gap |
| `schema-mirror` / `schema-reconcile` / `test-db.js` | 3/0 / 4/0 / exit 0 |
| `recurrence` | 21/0 |
| `waste-domain` / `-migrations` / `-routes` | 56/0 / 38/0 / 57/0 |
| `waste-ui` / `-import` / `-dashboard` | 31/0 / 28/0 / 23/0 |
| `waste-calendar` / `-url-source` / `-reminders` | 31/0 / 20/0 / 11/0 |
| `calendar` | 152/0 + 25/0 |
| `search` / `search-index-duplicates` | 22/0 / 9/0 |
| `preferences-dashboard` | 19/0 |
| `frontend-audit` / `i18n` / `i18n-translated` | 386/0 / 54/0 / 92/0 |
| `openapi-structure` / `openapi-coverage` | 11/0 / 3/0 |
| `changelog` / `readme-consistency` / `suite-chain` | 28/0 / 17/0 / 4/0 |
| `shopping-versions` (#1109) / `contact-categories` (#1144) | 19/0 / 19/0 |

## Round 4: verified

- **Migration 204's comment** reads `52/202/203`.
- **`test-calendar.js`**: the comment now says the value only keeps the test independent of the Waste axis.
- **`Number.isInteger` is bound.** Counter-proof on the merge: `Number.isFinite(v)` back in `server/routes/preferences.js` turns `test:preferences-dashboard` red on exactly the `[1.5]` assertion (18/1). Restored, 19/0.
- **`nearestOrdinalAnchorDateKey()` runs under real zones.** Counter-proof: `return target.toISOString().slice(0, 10);` instead of `toLocalDateKey(target)` turns `test:waste-ui` red on the three-zone case (30/1). Restored, 31/0.
- **The CHANGELOG entry** is under `[Unreleased]` on your head, not inside the published 2.66.0 notes.

## The one thing left: after #1144 and #1109 land

As soon as both are on `main`, please merge `main` into this branch once more. That merge will conflict in `server/db.js`, `server/db-schema-test.js`, `package.json`, `docs/test-suites.md` and `CHANGELOG.md`, all append-only and in the order above. Push it even if nothing else changes: CI has not run the full chain on this branch yet, because the gap guard stopped it at step 3, and that run is the first real CI result. I will merge once it is green. #1146, #1147 and #1099 come after that.

Issue #1083, opened 2026-09-09
Focus restore: two coverage gaps left over from #1070


ulsklyc:
---

## 4. The ordering filter blinds the ratchet to the most common shape (defect, not a gap)

> **Fixed in #1087** (merged 9 September 2026); the counterproofs and the nesting distinction are in that PR.

This one is different from the three above: it is not missing coverage, it is a guard that claims
protection it does not give.

The "handler renders asynchronously while the dialog is open" ratchet skips any `await load…()` that
appears textually before the `open(Shared)?Modal({` line in the same function. That filter was added
to silence a false positive in `openEditMemberModal` (admin-family.js), where the await really is
dialog preparation. But it also excludes the far more common shape, where the callback is declared
first and handed to the modal afterwards:

```js
const onChanged = async () => {
  await loadBudgetMeta();
  renderBody();
  refocusAfterRender();
};
openSharedModal({ /* ...onChanged... */ });
```

Five sites have that shape: `budget.js` (`openCategoryManager`), `inventory.js`
(`openLocationManager`, `openCategoryManager`), `pantry.js` (`openLocationManager`), `shopping.js`
(`openCategoryManager`).

**Measured:** deleting `refocusAfterRender()` from `budget.js` leaves both suites green
(`test:frontend-audit` 377 passing, `test:modal-utils` 45 passing). The calls are correct today, but
nothing holds them there.

The distinction the filter should draw is nesting, not order: an `await` inside a nested function
expression is a refresh, one at the top level of the opening function before `openModal` is
preparation. That separates the five sites from the admin-family false positive without excluding
either.

PR #1099, opened 2026-09-10
fix(schedule): audit findings across backend, UI, reminders, and i18n

mclgoerg:
## Heads-up: stacked on #1095

This branch was cut from `feat/waste-collection` (#1095), so it currently carries that PR's commit alongside its own — the two shared groundwork (the audit methodology and a few conventions, e.g. the read-only page lock and the calendar layer's per-member gating follow the pattern the Waste module established there). **Merge #1095 first**; this PR then reduces to the single `fix(schedule)` commit. Happy to rebase onto `main` at that point if anything shifts. The diff below describes only the Schedule commit.

## What this is

A full validation pass over the Schedule module — every route exercised with valid/invalid/boundary/foreign-user inputs, the UI walked at five viewports in both themes and both languages, read-only and denied members tested against both the API and the rendered page, and the resolved-entry consumers (ICS feed, reminders, dashboard widget, calendar layer) checked against each other. Everything it found, fixed in one sweep. No migrations.

## Server

- **`PUT /patterns/:id/days` is now capped at 500 rows** (`MAX_PATTERN_DAY_ROWS`). It was the one uncapped write loop in a module that caps everything else: every stored row is re-emitted as its own entry on every resolved read, so a single oversized save (verified with 5,000 rows in one request) became permanent read amplification on `GET /entries`, the ICS feed, the calendar layer, and the dashboard — stored work, not scheduling.
- **`field_values` writes are transactional and reject duplicate ids.** `id()` normalizes object keys via `parseInt`, so `{"1": …, "01": …}` passed validation and then hit the UNIQUE constraint mid-write — on the override route *after* the upsert had already committed, leaving half the values written and answering a raw 500. Duplicates are now a 400 up front, and the upsert plus value replacement run in one transaction (same for `POST /extras`, where a throw after the insert used to let an idempotent retry double-create the extra).
- **Omitting `field_values` from `PUT /overrides/:dateKey` now preserves stored values**, matching what `PUT /extras/:id` (and the OpenAPI description) already promised. Previously the override route silently wiped them on every note-only edit. An explicitly sent object — including `{}` — still replaces.
- **Deleting a pattern or a user no longer leaks custom-field values.** The value rows are polymorphic (`entry_type`/`entry_id`, no real FK), so the FK cascade on pattern days and user rows left them orphaned forever. Both delete paths now clean up inside their existing transactions, and SPEC.md's delete-path list names them.
- **`fieldValuesFor()` chunks its `IN (…)` lookup** so `GET /overrides`/`GET /extras` can't eventually break on SQLite's bound-variable limit once a household accumulates enough rows.
- **Override writes trigger the immediate reminder resync** the extras routes and the preferences route already did — and every inline resync call (here and in the extras routes) is now guarded, so a throwing sync can no longer turn a committed write into a 500.
- **`schedule_hidden_templates` is validated before any preference write.** The admin check sat mid-handler, so a non-admin's mixed payload partially applied before the 403.
- `scheduleData()` and its helpers moved from `routes/schedule.js` into `services/schedule.js`: the reminder sync needs the resolver and the router now needs the sync, and the module's own comments (see `routes/schedule-extras.js`) have always warned against exactly that route↔service import cycle. The route re-exports them, so nothing downstream changes.

## Page

- **Cancelling the "fill this date range?" confirmation no longer destroys the form.** Both override-save paths called `confirmModal()` over the open edit modal, which force-closes it — Cancel (or a failed save after Confirm) threw away everything typed. They now use `confirmOverModal()`, which parks and resumes the form; the modal component ships it for precisely this case.
- **Read-only members get an honest page.** The banner was always there, but every create/edit/delete control, the quick-start entries, the FAB, and the settings inputs rendered anyway and only failed at the API. They're now suppressed via the same `isNavModuleReadOnly()` idiom the Waste page uses; the server-side gate is unchanged and remains the real lock.
- **Statistics and Overview survive revisits and races.** Their state lives at module level, so returning to the page re-labelled cached entries under the freshly reset user (an admin who had picked another member saw that member's numbers titled as their own), Overview reopened stale, and "today" was computed once at module load. `render()` now resets and refetches the active tab, and a per-view generation counter stops a slow older response from overwriting a newer one. Both tabs also gained real loading and error states — a failed fetch used to render "0 shifts · 0 h" and silently empty lanes, indistinguishable from a genuinely free week.
- Enabling shift-start reminders defaults to a 15-minute lead instead of "at shift start" (`Number(null) === 0` had the 0-preset pre-selected and made the intended `?? 15` fallback dead code). The tab rail gained the arrow-key/roving-tabindex behavior its ARIA roles promised (shared `wireTablist()`), the printed statistics sheet no longer leads with the personal settings card or an emptied filter shell, hours format with the locale's own decimal separator, other members' pattern editors render disabled instead of editable-looking, the create modal's "Add" mode shows the field inputs for the pre-selected type, and a mid-loop failure while deleting an extras range re-renders instead of leaving already-deleted rows on screen.

## Calendar & dashboard

- `scheduleEnabled()` now checks per-member module access like `wasteEnabled()` directly below it always has — a member with schedule set to "none" no longer sees a dead "Schedule" layer row and no longer fires a guaranteed-403 `GET /schedule/entries` on every calendar load.
- The "who's working today" widget refreshes with the dashboard's 15-minute cycle instead of carrying the morning slice all day (and yesterday's staff past midnight), and a failed slice load renders the shared error tile with its retry button instead of the "create a shift type" onboarding.

## Reminders

- **Fire times are DST-correct.** The shared one-pass wall-clock→UTC conversion is up to an hour off around a transition and doesn't detect the spring-forward gap; the sync now uses a local two-pass fixpoint conversion (fold resolved deterministically, nonexistent times shifted past the gap), leaving the shared util untouched for its other callers. Verified against hand-computed Europe/Berlin values for both 2026 transitions, including inside the gap hour where the old conversion is provably wrong.
- **Stranded reminders are swept.** `schedule_reminder_entries.shift_type_id` is the module's one CASCADE reference, so deleting an override and then its now-unreferenced shift type inside the sync's 60-second window cascaded the anchor away and left the `reminders` row firing once as a title-less push, then sitting forever — invisible to both cleanup paths, which walk existing anchor ids. Every sync now starts with an indexed `NOT EXISTS` sweep for both schedule entity types.

## Translations

- `schedule.weeklyHoursHint` in all 24 locales described a "scaled to the number of days in the range" rule that was replaced by the rolling 7-day worst-window during this feature's original review — the hint now describes what the flag actually does, translated per language rather than copied. Five dead `schedule.*` keys removed everywhere.

## Tests

26 new tests pin the fixes and close pre-existing gaps: the row cap at both sides of the boundary, duplicate-key rejection with a no-partial-state assertion, omission-preserves vs. `{}`-clears, both orphan-cleanup paths (the user-deletion one runs through the real `DELETE /users/:id` route), the resync-on-override-write including a simulated throwing sync, the stranded-reminder sweep, the three DST cases with hand-derived constants, `PUT`/`DELETE /patterns/:id` (previously untested entirely, including the cycle-shortening refusal and foreign-user 403s), NULL-`valid_from` ranking and `is_active = 0` exclusion, overnight shifts through the overtime math (previously only a source-text grep), `extraGroups()` behaviorally (now parameterized like `overrideGroups()`), extra-shift reminders (`schedule_extra_entry` had no coverage at all), and the 731-day cap on the override range delete. The high-value ones were counter-proved by reverting the production line and watching them fail. The dashboard guard that pinned the old carry-over refresh by its literal text now pins the new mechanism instead, and the header comments in three suites no longer recommend the bare `node --test` invocation that bypasses `DB_PATH=:memory:`.

Full `npm test` passes end to end on this branch. The user-visible fixes were additionally verified in a real browser: Cancel returns to the intact form with the typed input preserved, a read-only member sees zero write affordances, a denied member gets no layer row and no 403 noise, the 15-minute default and the corrected hint text are live.


ulsklyc (changes requested):
Thanks @mclgoerg - this is a big, careful pass, and the server side holds up. Every fix I tried to undo turned a test red. Reviewed at `0837d50c`: your `54875f4e` plus one fix from me for my own rebase commit (below).

## Blocking

**1. Part of the CHANGELOG is in the released 2.66.0 notes.** The "A second pass, this time on the module's comprehension and everyday polish..." paragraph comes from `a87bd12c`. It sits under `## [2.66.0]` / `### Fixed`, attached to the unrelated "Module options settings page" bullet (around line 471). On your old base that bullet was still under `[Unreleased]`. After the release it moved into 2.66.0, and my rebase carried your paragraph along without a conflict. I should have caught that, sorry. The 2.66.0 notes are the published release body.

Please move it to `[Unreleased]` as its own bullet. While you are there, `[Unreleased]` does not mention the overtime-tracking toggle (migration 208) anywhere yet.

## Should fix

**2. The `/schedule/preferences` exception opens the module gate for members with no Schedule access** (`server/index.js`)

`scopedModuleKey` passes `null` as the module key, and `moduleAccessVerdict()` returns `allow` for any `null` key before it looks at the level. The intent in the comment is that a `read` member may write their own preferences. What the code does, measured against `moduleAccessVerdict` and `moduleForPath` from this branch:

| Request | Level | `main` | this PR | suggestion |
|---|---|---|---|---|
| GET/PUT `/schedule/preferences` | `none` | denied | **allow** | denied |
| PUT `/schedule/preferences` | `read` | read-only (403) | allow | allow |
| GET/PUT `/schedule/preferencesX` | `none` | denied | **allow** | denied |

The `startsWith` also matches longer paths. Suggestion: keep `moduleForPath(req.path)` and only lower the requirement for exactly this path, e.g. `req.path === '/schedule/preferences' ? 'read' : requiredAccess(req.method)`. Then `read` is allowed and `none` stays denied.

Nothing binds this today, because the schedule suites mount the router directly, past the global gate. Please put the key/level decision in a small function next to `moduleForPath` and test it for `none` -> denied and `read` -> allowed.

**3. A third parked form still closes before its save** (`public/pages/schedule.js:2237`)

The overlap question when **creating** a pattern calls `confirmOverModal` without `closeOnConfirm: false`, and `await api.post('/schedule/patterns', data)` follows. It is the same defect `54875f4e` fixed at the two override call sites. The test at `test/test-schedule.js:1893` only checks for `confirmOverModal(`, so it stays green.

**4. `dirtyPatternIds` gets out of step with the cards** (`schedule.js:42`, `:1687`, `:1698`)

The set is cleared only on these paths:
- the tab-switch discard;
- a save or delete of that same card;
- the page-level reset.

Saving pattern B, deleting an override range, or a successful save in the create modal all run `load()` + `renderPage()`, which rebuilds every card. Card A's unsaved cycle-day edits then disappear without the S-03 question. A also stays in the set, so the next tab switch asks to discard edits that are already gone. Either ask before those re-renders when the set is not empty, or at least clear it when the cards are rebuilt.

**5. Switching the statistics range does not outdate a running request** (`schedule.js`, `statistics-range`)

The handler resets `range`, `entries` and `bounds`, but it neither increments `statisticsRequestId` nor resets `error`. As a result:
- a statistics request already in flight still passes its `statisticsRequest === statisticsRequestId` check and draws the old range's numbers under the new choice;
- an error state survives the switch.

`++statisticsRequestId` and `error: false` there would settle both.

**6. En dash in a new locale value.** `schedule.cycleLengthTooShort` writes `{{from}}–{{to}}` in 19 locales: ar, cs, de, el, es, fil, hi, hu, id, it, nl, pl, pt, ru, sv, tr, uk, vi, zh. CONTRIBUTING ("Hyphens, not dashes") covers locale values; en, fr, ja and ko already use `-`.

## Mine, already fixed: `0837d50c`

My rebase commit `c7fcab26` called `refocusAfterRender()` after **every** successful inline save. The restore memo lives until the next dialog opens, so a save that never asked anything could send focus to the trigger of an earlier, unrelated dialog. That is the exact case the `gefragt` note in `action()` exists for.

It now waits for the overlap confirmation:
- `test:frontend-audit` 386/0 and `test:schedule` 105/0;
- without the call, the #1083 guard still flags the line.

**Please fetch before your next push:**

```
git fetch origin && git reset --hard origin/fix/schedule-audit-findings
```

## Nice to have

- **Keyboard tab switching navigates on every arrow press.** `wireTablist` calls `onChange` on Arrow/Home/End, and `guardedActivateView` calls `navigate()` each time. That means a history entry per press, a fetch on Statistics/Overview, and, with unsaved edits, the discard question on every press. Manual activation (Enter/Space) or replacing instead of pushing for keyboard moves would avoid it.
- `.schedule-body` has no `role="tabpanel"`; `budget.js` has one.
- **en wording:** "pattern" is now "Schedule plan" in five strings, but `overlapWarning`, `typeInUse` and `discardCycleDayEditsDetail` still say "pattern".
- **de:** `patternOverlapConfirmTitle` is "{{user}}s aktuellen Schichtplan ersetzen?", which reads wrong for names ending in s, x or z. "Den aktuellen Schichtplan von {{user}} ersetzen?" avoids that.
- **Dashboard:** `dashboard.js:2473` uses `widgetShown('schedule')` without the `isModuleDisabled` check that `domainInCockpit` two lines above has. Worth a look whether a disabled module can still steer the "free today" logic.
- **Tests:** about 25 of the new assertions are regexes over the source text. `__test` is already exported, so behaviour tests would be cheap, especially for the request-id races, which have no test at all.
- **router.js:** the nav-highlight mapping `path.startsWith('/schedule')` also matches `/schedules...`.

## Verified

- **Suites on `54875f4e`, each on its own:**
  - `schedule` 105/0, `schedule-reminders` 31/0, `schedule-feed` 18/0, `schedule-extras` 4/0, `schedule-custom-fields` 7/0, `schedule-default-off-migration` 5/0;
  - `dashboard` 111/0, `calendar` 153/0 + 25/0, `preferences-dashboard` 19/0, `notifications` 46/0, `family-routes` 4/0, `waste-calendar` 31/0;
  - `migrations-append-only` 3/0, `schema-mirror` 3/0, `schema-reconcile` 4/0, `test-db.js` exit 0;
  - `frontend-audit` 386/0, `openapi-structure` 11/0, `openapi-coverage` 3/0, `i18n` 54/0, `i18n-translated` 92/0, `i18n-plural` 18/0, `changelog` 28/0, `suite-chain` 4/0, `sw-api-cache` 16/0.
- **Server counter-proofs**, each from a `cp` backup and restored; every one turned exactly the named test red:

| Mutation | Red test |
|---|---|
| reminder time via the old one-pass `localToUTC` | `schedule-reminders`: nonexistent wall-clock time is shifted |
| `sweepStrandedReminders()` call removed | `schedule-reminders`: stranded reminder row is swept |
| 500-row cap removed | `schedule`: `PUT /patterns/:id/days rejects more than MAX_PATTERN_DAY_ROWS` |
| duplicate field-id check removed | `schedule`: key duplicated by normalization (2 failures) |
| override always replaces `field_values` | `schedule`: omitting `field_values` preserves stored values |
| pattern delete without value cleanup | `schedule`: `DELETE /patterns/:id` cascades days and values |
| user delete without override-value cleanup | `schedule-custom-fields`: deleting a user cleans up values |
| override `PUT` resync not guarded | `schedule`: a throwing resync does not turn the write into a 500 |
| `schedule_hidden_templates` check back mid-handler | `schedule`: non-admin mixed payload |

- **DST fixpoint**, hand-traced for Europe/Berlin:
  - the nonexistent 02:30 on 29 March becomes 03:30 CEST;
  - the doubled 02:30 on 25 October resolves deterministically to 01:30Z;
  - 01:30 and 03:30 land on the correct side of both transitions.
- **The sweep** goes through `NOT EXISTS` rather than anchor ids, which is the only way to find those rows.
- **Migration 208** is appended after 207. `overtimeEnabled: row?.o !== 0` keeps existing users on.
- **Moving `scheduleData()` into the service** breaks the route-service import cycle, and the re-export keeps existing imports working.
- **Service worker and router:** `sw.js` precaches the new `utils/schedule-tabs.js`, and the router registers the exact tab routes, so the notification deep link `/schedule/patterns` resolves.
- **Second pass on escaping and i18n:** no `innerHTML`; shift type, pattern, field, user and note values go through `esc()`; 27 new keys in all 24 locales without reserialisation. The overview week-flip race holds (`++overviewRequestId`, stale responses dropped).

I did not check the UI in a browser myself.

ulsklyc (approved):
Re-review on `8ca21847`: approving.

**Blocker.** The `## [2.66.0]` section is now byte-identical to the `v2.66.0` tag (compared by hand, since this branch's base predates the frozen-section guard from #1183). The second-pass paragraph sits as its own bullet under `[Unreleased]/Fixed` and covers the overtime switch.

**Should fix, each checked at the code path:**

- **(a) Module gate.** `sessionModuleAccessRequirement()` keeps the module key and only lowers the level for exactly `/schedule/preferences`. I ran the counter-proof myself: with the old `null` key and `startsWith` restored in `server/index.js`, `test:schedule-preferences-gate` goes 2/3 red (both `none` cases and `preferencesX`). With your version it is back to 5/0. Good that the test goes through the real server.
- **(b)** The third `confirmOverModal` now passes `closeOnConfirm: false`.
- **(c)** `dirtyPatternIds.clear()` runs before every `renderPage()` that follows a `load()` outside the page reset, and the walker test binds it.
- **(d)** `statistics-range` bumps `statisticsRequestId` and resets `error`.
- **(e)** `cycleLengthTooShort` uses a hyphen in all 19 locales.

**Nice to have.** `wireTablist({ manualActivation })` is opt-in and leaves the other callers unchanged. A click after Enter/Space cannot fire `onChange` twice, because `setActive()` returns early when the id is already current. The tabpanel wiring matches `#budget-body`.

Isolated run on the head, 26 suites, all green: `schedule` 108, `schedule-preferences-gate` 5, `tablist` 4, `dashboard` 112, `schedule-feed` 18, `schedule-extras` 4, `schedule-reminders` 31, `schedule-custom-fields` 7, `schedule-default-off-migration` 5, `dashboard-permissions` 18, `preferences-dashboard` 19, `permissions` 26, `permissions-routes` 14, `migrations-append-only`, `schema-mirror`, `schema-reconcile`, `frontend-audit` 386, `openapi-structure` 11, `openapi-coverage` 3, `i18n` 54, `i18n-translated` 92, `i18n-plural` 18, `changelog`, `suite-chain`, `router-guest-guard`, `token-scopes` 16.

Migration stays **208**. Before merging I will bring main in (two commits ahead: #1165 and #1183) and let CI run the full chain once more.

PR #1106, opened 2026-09-10
fix(settings): Module options description no longer omits half its sections

mclgoerg:
## What this is

A one-string fix, found as a side-observation while auditing the Schedule module (unrelated to that work otherwise).

## The problem

Settings → Modules → Module options describes itself as:

> Set individual options for Budget, Health and Housekeeping.

That was accurate when written, but the page has since grown a Tasks section (subtask-expansion default) and a Schedule section (quick-start templates) without this sentence ever being updated - it silently went stale twice.

## The fix

Reworded to describe the page's *purpose* instead of enumerating specific module names:

> Extra options for a few modules.

This can't go stale the same way a third time, since it doesn't name anything that could later be added or removed. Translated into all 24 locales (each with a real, per-language phrasing, not a copy of the English string).

## Test plan

- [x] `npm run test:i18n` / `test:i18n-plural` / `test:i18n-translated` green (all 24 locales)
- [x] `test/test-settings-copy.js` green (the guard that checks every noun in a leaf's description actually appears in what the leaf renders - the first wording I tried tripped this, since it named "settings page" content the page itself never renders; simplified to just "options"/"modules", which the page's own title already contains)
- [x] Full `npm test` exit 0
- [x] Verified live in the browser: the page now reads "Module options / Extra options for a few modules."

### DIFF
--- a/public/locales/el.json
+++ b/public/locales/el.json
@@ -2262,7 +2262,7 @@
         "pageKitchen": "Κουζίνα",
         "pageKitchenDescription": "Επιλέξτε ποια γεύματα εμφανίζονται στο πρόγραμμα γευμάτων.",
         "pageModuleOptions": "Επιλογές ενοτήτων",
-        "pageModuleOptionsDescription": "Ορισμός μεμονωμένων επιλογών για Προϋπολογισμό, Υγεία και Οικιακή βοήθεια.",
+        "pageModuleOptionsDescription": "Πρόσθετες επιλογές για μερικά modules.",
         "pageCalendarModule": "Ημερολόγιο",
         "pageCalendarModuleDescription": "Ορίστε αργίες, έναρξη εβδομάδας και προεπιλεγμένη διάρκεια συμβάντος.",
         "pageSyncCalendar": "Συγχρονισμός ημερολογίου",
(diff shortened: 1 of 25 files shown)


public/locales/el.json
```diff
@@ -2262,7 +2262,7 @@
         "pageKitchen": "Κουζίνα",
         "pageKitchenDescription": "Επιλέξτε ποια γεύματα εμφανίζονται στο πρόγραμμα γευμάτων.",
         "pageModuleOptions": "Επιλογές ενοτήτων",
-        "pageModuleOptionsDescription": "Ορισμός μεμονωμένων επιλογών για Προϋπολογισμό, Υγεία και Οικιακή βοήθεια.",
+        "pageModuleOptionsDescription": "Πρόσθετες επιλογές για μερικά modules.",
```
ulsklyc: Greek mixes in the English word here; the title uses `ενότητες`:

```suggestion
        "pageModuleOptionsDescription": "Πρόσθετες επιλογές για ορισμένες ενότητες.",
```

ulsklyc (commented):
Thanks - checked at `691fff5f`. The new wording replaces the stale enumeration in all 24 locales, the 4-space indentation is intact, the CHANGELOG entry sits under `[Unreleased]` / `### Fixed`, and CI is green on Node 22 and 24.

One translation to fix before this goes in: the Greek sentence carries the English word (`μερικά modules`). The page title right above it already uses `ενότητες` (`"pageModuleOptions": "Επιλογές ενοτήτων"`), so the description should follow it - suggestion inline. The other 23 locales read naturally as far as I can check them.

Issue #976, opened 2026-09-01
Der Router raeumt die Timer einer Seite nicht ab


ulsklyc:
Found while reviewing #974 (kitchen timer on the wall surface), but this is older and wider than that feature.

`public/router.js` has no teardown contract for pages. Its `cleanup()` only removes route stylesheets; a page controller is never aborted. The dashboard aborts its own `_fabController` at the top of `render()`, which covers "navigate away and come back" but not "navigate away and stay away".

So after leaving `/` via Back or any other route, these keep running against a detached container:

- the clock minute tick (which also drives `syncWallMode`)
- the silent dashboard refresh
- the weather timer
- and, since #974, the wall kitchen timer

For the timer the visible consequence is the sharpest: on expiry it chimes and calls `rerender()` on a container that is no longer on screen, which then starts dashboard requests behind another page. The others fail more quietly, which is why nobody has noticed.

**The fix belongs in the page lifecycle, not in any one page.** Something like: the router owns an `AbortController` per page render, hands the signal to `render()`, and aborts it when the route is replaced. Every page that today keeps its own controller could then drop it.

Related: the superseded-render problem in the same page.

ulsklyc:
Shipped in v2.65.1 (`80315002`). The router now owns one AbortController per page render and hands `signal` to `render()` in both context shapes; `utils/page-lifecycle.js` bridges it to the controller a page keeps per own rebuild, and the dashboard binds every timer and listener to it. Measured in the browser: three long timers live after the dashboard renders (15-minute refresh, clock tick, 30-minute weather), none after navigating to the calendar. Third-party modules get the same `signal` (MODULES.md). Guards: `test:page-lifecycle` plus two source guards in `test-frontend-audit.js`. The render-token half is #977.

Issue #276, opened 2026-06-07
[Bug] check on shopping list moves list to the top


Nihmar:
## Description

I don't really know if I'm reporting a bug or a feature, but I think this is not working how someone would expect it to work.

If I have a shopping list longer then the height of the screen (so I have to scroll down to see it whole) and I check/uncheck an item, the list scrolls back to the top, instead of staying in place.

If think it is a bug, because when you are shopping, generally you find products of the same category in the same isle, and it doesn't make sense to check an item, and then having to scroll to the same point to check another.

## Steps to Reproduce

1. Go to kitchen, shopping list. Make a long list of items
2. Click on an item to the bottom
3. See the list go back to the top

## Expected Behavior

To me and my wife, it should stay in place

## Actual Behavior

What actually happened is that it goes back to the top

## Environment

- **Browser:** Brave and Edge
- **Device:** Realme 12+
- **OS:** Android 15 and 16
- **Docker version:** 0.6.2 (but looking at the releases, I don't think this was tackled

Nihmar:
Mark it closed, I just updated and it is fixed.

Sorry for making you lose time.

ulsklyc:
Thanks for the clear report — you're right, this was a bug, not intended behaviour.

**Cause:** the whole item list shares a single scroll container, and checking/unchecking an item re-rendered (emptied and rebuilt) that entire container. Emptying it clamps the scroll position back to the top, so the list jumped up.

**Fix:** toggling now updates only the affected row in place instead of rebuilding the list, so your scroll position is preserved while shopping. This applies to both tapping an item and the swipe-to-check gesture.

Released in **v0.64.2** — it'll be in the next Docker image. Thanks again to you and your wife for spotting it! 🛒

Issue #1002, opened 2026-09-02
Health API has no change feed: a bridge can read and write, but cannot ask what changed


ulsklyc:
Named in three threads and tracked in none: #746, #743 and #639 all run into the same wall, and all three answers describe it as a gap rather than a decision. This is the place for it.

## Why it blocks

The shape this project wants for provider integrations is a bridge: Yuvomi stays the server, the bridge is a client holding a scoped API token. That works for pushing data in. It does not work for anything that stays in sync, because a bridge cannot ask what changed since it last looked - so a two-way or resumable sync has no foundation, and the only options are push-only or a full sweep every run.

## What exists today

- `health_vitals` and `health_activities` carry both `created_at` and `updated_at`, kept current by an `updated_at` trigger. **The data a feed needs is already in the database.**
- The list endpoints filter on `type` and on `from`/`to` - but `from`/`to` apply to `measured_at` and `performed_at` (`server/routes/health/vitals.js:19`, `server/routes/health/activities.js:19`). That is *event* time, not *record* time.

## The gap, in three parts

1. **No record-time filter.** Correcting a three-week-old weight does not move it into a `from=<yesterday>` window, so a poller that walks event time misses every edit. The column exists; nothing exposes it.
2. **Deletions cannot be learned.** `DELETE /vitals/:id` is a hard delete with no tombstone (`server/routes/health/vitals.js:113`). A bridge that already copied a row out has no way to find out it is gone. The calendar solved this exact shape for itself - migration *"Calendar: tombstones for events deleted locally but still remote"* (#593) - so both the precedent and the cost are already in the repo.
3. **No cursor to resume from.** `updated_at` has second granularity and will tie, so resuming needs `(updated_at, id)` or a dedicated sequence rather than a bare timestamp.

## Sketch

- `?updated_since=` on the health list endpoints, ordered by `(updated_at, id)`, returning the cursor to resume from.
- Tombstones for deletions with a retention window - **or** the explicit decision that deletions are out of scope and a bridge has to reconcile by full sweep. Either answer is usable; the absence of an answer is not.
- Health-only or the shape every module gets? Health is where it blocks today, but the same question returns for tasks, calendar and budget, and answering it once is cheaper than three times.

## Adjacent, and also still unanswered

[...]

PR #1109, opened 2026-09-10
feat(shopping): live-update an open list when another member changes it

### LINKED ISSUE #1108: Shopping: an open list does not update when another member ticks items off
Two people shopping together with the same list open on two phones see two different lists. What one of them ticks off stays unticked on the other's phone until that page is reloaded (a list switch or a full reload). The same holds for items added or removed by the other person, and for items a meal-plan import or the CalDAV to-do sync puts on the list while it is open.

The page is already prepared for this: `public/pages/shopping.js` carries the intent/`settledAt`/`_loadSeq` rules that let a refresh overtake a tap without flipping the row back ("jemand anderes im Haushalt" is named in those comments). What is missing is anything that *triggers* a refresh while the list is on screen - the only callers of `loadItems` are user actions.

**Proposed approach** (PR follows):

- A per-list change counter (`shopping_list_changes`, one migration) bumped by triggers on `shopping_items` for insert, update and delete. Triggers rather than calls in the routes because `shopping_items` is written from six server modules (shopping, meals, recipes, housekeeping, MCP, CalDAV reminders sync) - one rule in one place, in the sense of DECISIONS.md entry 2.
- `GET /api/v1/shopping/feed`, a Server-Sent Events stream that reads that table once a second while anyone is connected and only says *which* list moved (`versions` on connect, `change`, `ping`). No content on the stream: the page reloads through the same `GET /:listId/items` it already uses, so there stays one read path. Nothing polls while nothing changes, and a household behind one NAT address does not eat into the 300 req/min limiter the way client polling would.
- The page redraws rows in place when only the checkbox moved (same gesture as one's own tap, no rebuild, no scroll jump, #276) and rebuilds only when an item was added, removed or renamed. The stream hangs on the router's abort signal (#976). Fallback when the stream cannot get through (buffering proxy, no EventSource): catch up on `visibilitychange` and every 30 s.
- `no-transform` keeps `compression()` out of the stream, `X-Accel-Buffering: no` keeps nginx from buffering, and the service worker passes `text/event-stream` requests straight through (`networkFirstApi` would otherwise wait on `blob()` for a body that never ends, and `/shopping/` is on its whitelist prefix).

Happy to change the shape if you would rather see this done differently - a plain client-side poll of `/:listId/items` would be the smaller diff, at the cost of constant traffic from every open tab.

ter9ng:
## Summary

Closes #1108.

Two people in the same shop saw two different lists: what one ticked off stayed unticked on the
other's phone until that page was reloaded. The page already had the rules for a refresh
overtaking a tap (the intent / `settledAt` / `_loadSeq` machinery), but nothing ever triggered a
refresh while the list was open.

The server keeps a change counter per list, maintained by database triggers rather than by the
routes: `shopping_items` is written from six modules (the list itself, meal-plan and recipe
imports, housekeeping, MCP, the CalDAV to-do sync), and a counter every writer has to remember is
a counter one of them forgets. The counter also moves when a list is renamed, when an item moves
to another list (both lists), and its row goes with the list. Every list has a row from the start
(backfill plus an insert trigger), so a client's first sighting is never a change that gets lost.

The page polls `GET /api/v1/shopping/versions` every 10 s while the tab is visible, and at once
on `visibilitychange`/`focus`. It reloads only a list whose number moved, through the same items
request it used to open it, so there is still one read path. Deliberately a poll and not a
stream (review): `/api/v1` is the contract, a versions list works through any proxy, holds none
of the browser's per-origin connections, and can grow into a stream on the same counter later.

A row whose only change is the checkbox is redrawn in place - the same gesture as one's own tap,
no rebuild, no animation, no scroll jump (#276). An added, removed, renamed or moved item rebuilds
the list; a list someone else renamed or deleted updates the tab row, and a deleted active list hands
over to the first remaining one. The poll is bound to the router's abort signal (#976).

Own taps cost no reload: the item write routes answer with `list_change: { list_id, before,
after }`, and the page advances its mark only when `before` is the version it last saw. If
someone else wrote in between, `before` differs and the next poll reloads. The items answer carries
`version`, and the page hands it to the feed as the state it holds, so the order of the two
answers on open does not matter. Own edits also survive a reload: a pending tap stays on top of an
older server answer, and the answer that carries it settles the intent; a row removed locally
stays out of every answer that started before its DELETE was confirmed, and "clear checked" names
the ids it removed. All of that is behaviour-tested.

Not yet exercised on two real devices; I will test it on my own instance and mark the PR ready
once that is done.

Drafted with Claude Code, reviewed and tested by me.

## Changes

- `server/db.js`: migration v196 - `shopping_list_changes(list_id, version)`, backfill for
  existing lists, triggers on `shopping_lists` (insert → row with 0, update → bump, delete →
  cleanup) and on `shopping_items` (insert/update/delete → bump, plus `UPDATE OF list_id` →
  bump the list left; the update trigger compares the columns a client shows, so the outbound
  flags, the nested `updated_at` UPDATE and an unchanged sync rewrite do not count) and on
  `shopping_item_tags`. No foreign key on purpose: the item triggers fire during the cascade, and
  a foreign key would make exactly that insert fail (tested). Mirror in
  `server/db-schema-test.js`. Numbered 196 as agreed (#1055 landed as 194, #1144 takes 195);
  the append-only guard is red on the gap until #1144 lands.
- `server/routes/shopping.js`: `GET /versions` (before `/:listId`, like `/categories`);
  `withListChange()` wraps the item writes (`POST /:listId/items`, `PATCH /items/:id`,
  `DELETE /items/:id`, `DELETE /:listId/items/checked`, `PATCH /:listId/items/reorder`) and adds
  `list_change` to their responses; `GET /:listId/items` carries `version`;
  `DELETE /:listId/items/checked` takes an optional `{ ids }`. OpenAPI entries in
  `server/openapi/paths/shopping.js`.
- `public/utils/live-feed.js`: `startLiveFeed()` - polls while visible, reports a list whose
  version rose, appeared or disappeared, `acknowledge()` for own receipts, `hold()` for the state
  the page loaded, ignores stale answers, replays what came before the first answer.
- `public/pages/shopping.js`: `wireLiveUpdates()` (poll bound to the page controller, wake on
  visibility/focus, one refresh at a time), `refreshFromFeed()` (silent on failure, restores the
  tab row if `loadLists()` recorded an error, hands over when the active list is gone),
  `liveRefreshPlan()` (in-place vs. rebuild), `acknowledgeOwnChange()` at the six write sites.
  `render()` takes `signal` from the router and the retry button passes it on.
- `public/sw.js`: `/utils/live-feed.js` in the precache list (`test:sw-precache`).
- Tests: `test/test-shopping-versions.js` through the real server (401 / 403 member / 403 token,
  every trigger incl. the moved item and a write past the route, first change after the
  migration, backfill against a pre-v196 database, `list_change` on every write route, the
  bookkeeping writes that must not count, an own tap on a mirrored list),
  `test/test-live-feed.js` (poll logic), nine behaviour tests in `test/test-shopping-ux.js`
  (plan, refresh, own pending/settled intents, abort, vanished list, wiring with receipt). Both
  new suites are in the `test` chain and in `docs/test-suites.md`.
- `CHANGELOG.md` (Added), `docs/SPEC.md` (feature bullet + data model).

## Checklist

- [x] `npm test` passes - suites touching this change pass locally; on my Windows machine
  `test:frontend-audit` (ESM `--check` path), `test:openapi-coverage` (sub-router paths) and
  `test:readme-consistency` fail identically on unmodified `main`, so CI is the reference
- [x] UI strings use `t('key')` (no hardcoded text) - no new UI strings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01W7zpL4WCwtXAnj5qJJMzuw


ulsklyc (changes requested):
Thanks for this - the core of it is right, and it is the shape I would want: one counter maintained by triggers, so none of the six writers can forget it (DECISIONS.md entry 2), a signal that only says *that* a list moved, and a reload through the page's existing read path. The handling of your own pending taps against a refresh is careful work, too.

Before you put more into the transport, one decision from my side, and then three things that need to change regardless of it.

## Transport: please poll a versions endpoint instead of holding a stream open

I would rather not make this the first long-lived connection in the app. Concretely:

- `GET /api/v1/shopping/versions` returns the snapshot the feed already sends on connect (`readVersions()`).
- The page asks every ~10 s **while the tab is visible**, and immediately on `visibilitychange`/`focus`. It reloads a list only when its version moved; `liveRefreshPlan()` and `refreshFromFeed()` stay as they are.

The reasons, in order of weight:

1. **Reversibility.** `/api/v1` is the contract (`docs/SCOPE.md`, "What is promised, and what is not"), and with an OpenAPI entry the event format becomes part of it. A polled versions list can grow into a stream later, on the same counter; a published stream cannot be withdrawn without breaking that promise. A versions list is also the shape the change-feed gap in #1002 asks for.
2. **Where it runs.** An install reached without a reverse proxy speaks HTTP/1.1 (`app.listen` in `server/index.js`), and there every open shopping tab holds one of the browser's ~6 connections per origin. The stream also stays open while the tab is hidden: `wireLiveUpdates()` closes it only when the page is left. Behind a proxy, every product buffers differently beyond the nginx and Caddy cases you covered. And per the EventSource spec a non-200 answer (a 502 during a container update, a 429, a 401) fails the connection for good - there is no `error` handling in `connectLiveFeed()`, so the page sits in its 30 s fallback until the next navigation.
3. **No server state.** No subscriber set, no ticker, no service-worker exception for `text/event-stream`.

The cost is latency: up to ~10 s instead of ~1 s. Standing next to someone in a shop, that still reads as "it updated", and it is the trade I would make here.

## Blocking, independent of the transport

1. **An item that moves to another list is not reported to the list it left.** The update trigger bumps only `NEW.list_id` (`trg_shopping_items_change_au` in `server/db.js`), but `server/services/caldav-reminders-sync.js` rewrites `list_id` on existing items when a selection's target list changes. The old list never learns that items left it. The trigger needs to bump `OLD.list_id` as well when it differs. This one matters before a release, because a trigger shipped in a migration can only be corrected by another migration.
2. **The first change to every list is lost.** `shopping_list_changes` starts empty, and `connectLiveFeed()` treats the first version it sees for a list as its baseline - `test/test-live-feed.js` pins exactly that ("die erste Sichtung einer Liste ist ihr Ausgangsstand"). After the update, no device sees the first change another member makes to an existing list, and the same happens for every newly created list. A backfill in the same migration (`INSERT INTO shopping_list_changes (list_id, version) SELECT id, 0 FROM shopping_lists`) covers the existing lists; an insert trigger on `shopping_lists` that writes version 0 covers new ones.
3. **Migration 194 is also claimed by #1055**, which has been in review longer. Please expect to renumber: `server/db.js`, the mirror in `server/db-schema-test.js`, the `v194` test names and the comments in `server/services/shopping-feed.js`. Whichever lands second takes the next free number.

## Worth deciding while you are in the migration

**Should the counter carry list-level changes too** - a list created, renamed or deleted by someone else? Today a list another member deletes stays on every open device. That is the same on `main`, so not a regression. Two more triggers on `shopping_lists` in the same migration would cover it, and `refreshFromFeed()` already redraws the tabs; adding it later costs another migration and mirror. I would include it, but it is your call.

## Smaller points

- **Tests through the real stack.** `test/test-shopping-feed.js` mounts the router on a bare Express app with a hand-set user, so authentication, the module gate and token scopes are never exercised. For the versions endpoint please cover: 401 without a session, 403 for a member whose shopping access is none, 403 for a token without `shopping:read`, a moved item bumping both lists, and the first change after the migration.
- **Reload cost.** Every change makes each open tab reload through `Promise.all([loadLists(), loadItems(listId)])` - two requests per tab, including the tab that made the change. Phones behind one household IP share the per-IP limiter, so skipping the reload when the new version is the one your own write just produced keeps that in bounds.

No need to polish the SSE-specific parts first; I will look again once the transport is switched.

ulsklyc (changes requested):
Thanks for turning this around so quickly, and for taking all of it. The transport switch reads exactly as intended, and I checked the migration against the real schema: a moved item bumps both lists, `SET list_id = list_id` bumps only the list the item stays in, every existing list gets its 0 row from the backfill, a new list starts at 0, and deleting a list (or a user, through the cascade) leaves no orphan row behind. Keeping 194 until one of #1055 and this lands is fine; main now also has a guard that compares existing migrations against the base (#1111), and this branch passes it.

The before/after receipt is a good idea, and it is also what turns two client races into state that stays wrong. Both come from the same assumption: `acknowledge()` treats everything up to `after` as already reflected on screen, but a live refresh can land between the optimistic local change and the receipt.

## Needs a fix before merge

**1. A deleted item comes back during the undo window, and stays after the delete.**
Steps: delete an item (or use "clear checked"), and within the 5 s undo window let the list change on the server so a poll lands inside the window.

- `refreshFromFeed()` -> `loadItems()` replaces `state.items` with the server state (`shopping.js:2323`), which still has the item, because the DELETE is only sent when the window closes. The row reappears under the "deleted" toast.
- When the window closes, the DELETE's receipt has `before` equal to the mark the refresh just set, so the mark advances (`live-feed.js:98`) and the next poll does not reload. The deleted row stays on screen until the list changes again.
- Undo inside the window pushes the snapshot into an array that already contains the item again (`shopping.js:550`, `:2113`), so it shows twice.

What should hold afterwards: a row removed locally does not return while its undo window runs, undo never duplicates, and a receipt never stands in for a reload the local state did not get.

## Should fix

**2. Saving the item dialog after a live refresh loses the edit locally.** `openItemDetails()` captures the item object when the dialog opens (`shopping.js:1607`), and saving does `Object.assign(item, data.data)` (`:1765`). If a poll reloaded the list in the meantime, that object is no longer in `state.items`. With a category change, `updateItemsList()` redraws from `state.items` and shows the old name and aisle at once; otherwise the next rebuild does. The receipt again advances the mark, so no poll heals it. Looking the item up by id at save time would cover it.

**3. The service worker caches the versions poll.** `'/shopping'` is a prefix in `API_CACHE_WHITELIST` (`public/sw.js:35`), so `networkFirstApi()` writes every successful poll into Cache Storage, and when the network drops, the page gets the last snapshot back as a normal 200 through `api.get()`. Your stale-answer rule keeps that mostly harmless, but it is a cache write every 10 s per open tab for data that has no value offline. I would exclude `/shopping/versions` in `isCacheableApiGet()` (`:592`), with a case in `test:sw-api-cache`. Last time I counted "no service-worker exception" as a plus of polling; this small one I missed.

## Smaller points

- `docs/SPEC.md`: the new "Shopping List Changes" section sits between the Shopping Stores table and its two closing paragraphs (price in minor units, `store_id` SET NULL), which now read as part of the counter section.
- CHANGELOG and SPEC say "within a few seconds"; with a 10 s poll, "within about ten seconds" is the honest wording.
- `shopping_item_tags` has no trigger. Nothing is missed today, because the only writer (`setItemTags()` in the CalDAV to-do sync) runs right after an UPDATE of the same item. But `liveRefreshPlan()` compares tags, and a trigger in this migration costs less than a migration later. Your call.
- For information only: the CalDAV to-do sync rewrites every mirrored item on every run, even when unchanged, so each run makes every open tab of a mirrored list reload once. Cheap, fine to leave.

CI is green on `19eed28b`. You mentioned testing on two phones first; the PR is already marked ready, so I will take the next look once these are in.

ulsklyc (changes requested):
Round 3, reviewed at `1cbfd76d`. Each point below was checked by taking the fix out again and running the suite.

## Round 2 is in

- **Deleted rows during the undo window.** `pendingRemovals` is set before the optimistic removal (`shopping.js:554`, `:2153`), `loadItems()` filters it (`:2386`) and `loadLists()` subtracts it (`:2299`). Without the filter two of the new tests go red, without the counter subtraction three.
- **Item dialog.** Save looks the item up at save time (`:1809`). With the captured object put back, the dialog test goes red.
- **Service worker.** `/shopping/versions` is excluded (`sw.js:601`). Without the check, the new `test:sw-api-cache` case goes red.
- **Tag triggers.** Both are in the migration and in the mirror. With them removed from both, `test:shopping-versions` goes red on the tag test.
- The SPEC section sits in the right place now, and the wording is "within about ten seconds".

Green on this head: shopping-ux 60, live-feed 12, shopping-versions 15, sw-api-cache 16, schema-mirror, shopping-routes 57, migrations-append-only, shopping-send, caldav-reminders. CI is green too.

## Needs a fix before merge

**A load that started before the DELETE and answers after it brings the row back, and it stays.**

`commit` clears `pendingRemovals` in its `finally` as soon as the DELETE returns (`shopping.js:569`, and `:2165` for clear checked). A `loadItems()` that is still in flight at that moment carries the server state from before the DELETE. It passes both load guards (`:2357` same list, `:2362` no newer load applied), finds no pending entry any more, and puts the row back. The DELETE's receipt has already moved the mark (`live-feed.js:98`), so the next poll does not reload.

This is the window the feature itself opens: a poll lands, `refreshFromFeed()` starts `loadItems()`, and the 5 s undo window closes while that GET is still on the wire.

Reproduced on this head with the harness from your own test, only with the GET answered after `commit` instead of before it:

```js
test('a GET that started before the DELETE and answers after it', async () => {
  resetShoppingState();
  __test.state.lists = [listRow(0)];
  __test.state.items = [milk(0), bread(0)];
  global.window.yuvomi.showToast = () => {};
  const undo = captureUndo();
  let releaseGet;
  globalThis.__apiStub = {
    get: async () => ({ data: [listRow(0)] }),
    // The GET read the server before the DELETE arrived: the milk is still there.
    getWithSource: () => new Promise((resolve) => {
      releaseGet = () => resolve({ data: { data: [milk(0), bread(0)] }, fromCache: false });
    }),
    delete: async () => ({ data: null, list_change: { list_id: 1, before: 5, after: 6 } }),
  };

  __test.deleteItemUndoable(10, makeNullContainer());
  const refresh = __test.refreshFromFeed(makeNullContainer(), 1, new AbortController().signal);
  await new Promise((r) => setImmediate(r));
  await undo.last.commit({ keepalive: false });
  releaseGet();
  await refresh;
  assert.deepEqual(__test.state.items.map((i) => i.id), [11]);
});
```

Result: `actual: [ 10, 11 ]`. The existing test (`test-shopping-ux.js:1345`) answers the GET before `commit`, which is why it stays green.

The page already solves this for checks: `settledAt` (`shopping.js:350`, `:482`) keeps a confirmed check from being turned back by a response that was sent before the confirmation. A removal needs the same rule. After a successful DELETE, keep the entry, stamped with `_loadSeq`; keep filtering responses from loads that started before that stamp; drop the entry once a load that started after it has applied. After a failed DELETE, clearing at once stays right, because `restore` brings the row back. `loadLists()` needs the matching rule, since a list load that started before the DELETE still counts the item.

What should hold afterwards: a response that started before the DELETE never brings the row back, whichever order the answers arrive in.

## Should fix

**Clear checked can delete an item someone else checked during the undo window, and the receipt then hides it.** The route deletes whatever is checked when the DELETE arrives (`server/routes/shopping.js:1165`), not the items the page removed. If another member checks one more item within the 5 s and a poll brings that into the page, the page shows it checked, the server deletes it along with the rest, and the receipt moves the mark. The row stays on screen until the list changes again; without the receipt, the next poll would have seen the version move and reloaded. The response already carries `deleted`, so acknowledging only when it equals `count` (`shopping.js:2140`) lets the poll reload in the other case. Sending the ids would work too, but changes the API. I found this by reading the code, not with a test.

## Small

- The duplicate check in `restore` (`shopping.js:585`, `:2173`) is not bound by a test: with it removed, the suite stays green. With `pendingRemovals` in place I found no path that reaches it, so it is fine as a safety net. Just noting it.

## Migration number

Unchanged: #1055 lands first and keeps 194, and this branch moves to 195 after that. #1055 still needs one more merge with `main`, so there is nothing to renumber yet.

ulsklyc (changes requested):
Round 4. **Both round-3 findings are genuinely fixed** - I traced the new mechanism rather than trusting the description, and it holds:

- `beginLoad()` is the statement immediately before `try` in both loaders and `endLoad()` is in both `finally` blocks; nothing between them can throw and neither loader passes an abort signal, so `_loadsInFlight` cannot leak in a way that matters. Even a hypothetically never-settling fetch is bounded - `removalHides` only hides a response with `startedAt <= removedAt`, so a stuck entry cannot hide a *later* load.
- The comparison is drawn identically at all three sites: `removalHides` keeps while `removedAt >= startedAt`, `pruneSettledRemovals` drops when `removedAt < oldest`. Exact complements, no off-by-one, and both consult sites go through `removalHides` rather than open-coding it.
- The service-worker window is closed twice over (`/shopping/versions` is in `API_CACHE_EXCLUDE` and tested in both directions; `loadItems` drops a `fromCache` answer once the list is applied).
- The tests reorder for real: the GET is held open, the commit runs, `assert.ok(pendingRemovals.has(10), ...)` fires *between* them, then the GET is released. That is a direct counterproof against the round-3 regression, not a restatement of it.
- `reloadsOnNextPoll()` drives the real feed and asserts both directions (`deleted: 1` -> no reload, `deleted: 2` -> reload).

That was the hard part, and it is done. Three things still block.

## Blocking

**1. `server/db.js:7605` - migration must move to 195.** #1055 keeps 194 (it was promised that number and is merging now); `main` is at 193. Ten sites move together: `server/db.js:7605` and its comment block, `server/db-schema-test.js:1116` plus the key at `:1120`, `server/routes/shopping.js:109`, `test/test-shopping-versions.js:2/:18/:252/:253/:254`, `docs/SPEC.md` ("migration v194", two places), `docs/test-suites.md`.

**2. `public/pages/shopping.js:2630-2634` - the early return leaves the DOM permanently behind `state.items`.** `loadLists()` never rejects (it catches internally and records `state.listsError`), so after `Promise.all([loadLists(), loadItems(listId)])` at `:2619` the items have **already** been written to `state.items` when the `state.listsError` branch returns.

Scenario: member B adds "Eier"; A's poll fires; `GET /shopping` returns 500 while `GET /shopping/1/items` succeeds. A's `state.items` now holds "Eier", the DOM does not, and `updateItemsList` is never called. On the next poll `liveRefreshPlan(previous = state.items, fresh)` compares the already-applied state against identical fresh data, gets `{rebuild: false, changed: []}`, and **the row never appears** - not after ten more polls. The doc comment at `:2580` states the assumption this path breaks ("die Seite haelt beides ohnehin deckungsgleich"). Either restore `state.items = previous` as well, or fall through to the draw.

**3. `public/pages/shopping.js:2583-2591` - `liveRefreshPlan` cannot see a reorder, and the test that says it can does not contain the case.** `shape` omits `sort_order` and both sides are `.sort()`ed, so a pure permutation is byte-identical -> `rebuild: false`, `changed: []`. But `loadListItems` orders by `CASE category ..., is_checked ASC, sort_order ASC` (`server/routes/shopping.js:94-98`), so `PATCH /:listId/items/reorder` *does* change the response order and *does* bump the counter.

Scenario: member B drags "Brot" above "Milch"; A reloads, `state.items` takes the new order, the DOM keeps the old one until some unrelated change forces a rebuild.

The test titled `'liveRefreshPlan: ein neuer, fehlender, umbenannter oder umsortierter Artikel baut die Liste neu'` asserts five cases - `dazugekommen`, `verschwunden`, `umbenannt`, `andere Kategorie`, `andere Menge` - and **no reorder case**. The name promises a guarantee the code cannot keep, which is worse than the gap itself: it is the kind of test that stops the next person from looking. Clean fix: add `i.sort_order ?? ''` to `shape`. That is precise here, because a check-off does not change `sort_order` (only the display order), so it introduces no spurious rebuilds - which is why the `.sort()` was there in the first place.

## Should fix

**4. `server/routes/shopping.js:579` - your own tap still costs a reload on a CalDAV-mirrored list.** `markTodoOutbound` runs *after* `withListChange()` read `after`, and before `res.json()`. It executes `UPDATE shopping_items SET outbound_dirty = 1, ... WHERE id = ?`, which trips `trg_shopping_items_change_au`. One tap is three counter steps (route write, dirty=1, dirty=0 after the push) while the receipt reports only the first: the client acknowledges `after = N+1`, the server sits at `N+3`, and the next poll reloads for every open client including the one that tapped. The CHANGELOG's "Your own taps do not cost a reload" never holds for a mirrored list. Both DELETE routes get this right by calling `queueTodoDeletions` *before* `withListChange` (`:663`, `:1161`).

**5. `public/pages/shopping.js:2619-2634` - `state.listsError` can stick permanently and silently kill live updates.** `Promise.all` short-circuits on `loadItems`' rejection while `loadLists()` is still in flight; the guard at `:2630` is then evaluated before `loadLists` has settled, and the flag it was meant to clear is set afterwards. From there `refreshFromFeed` returns at `:2613` for the rest of the page's life, with no error UI and `state.lists` left at `[]`. The existing test `'refreshFromFeed: ein Fehler beim Nachladen bleibt still'` cannot catch it - both stubs throw in the same microtask, so `loadLists` happens to settle first. `Promise.allSettled` (or awaiting both handles explicitly) before reading the flag.

## Nice to have

- `server/services/caldav-reminders-sync.js:345-350` - `upsertShoppingItem` re-`UPDATE`s every existing mirrored row unconditionally. Each inbound pass moves the counter once per mirrored item, and every open page reloads for a no-op. Not yours, but your trigger makes it user-visible for the first time; a `WHERE ... AND (name IS NOT ? OR is_checked IS NOT ? OR list_id IS NOT ?)` guard settles it.
- `public/utils/live-feed.js:95` - `acknowledge()` is a no-op while `seen` is `null`, so a tap before the first `/shopping/versions` answer costs one pointless reload.
- `public/pages/shopping.js:3105` - wiring the feed before the load narrows the open-time race but does not remove it: if `/versions` is read after `/items` and someone writes in between, `seen` is initialised past that change. The comment at `:3098-3104` says the ordering removes the gap; it moves it.
- `public/pages/shopping.js:2384` - `loadLists()` has no `fromCache` guard and `/shopping` is in `API_CACHE_WHITELIST`, so on a network blip a pre-delete snapshot arrives as a plain 200 and the tab counter is one too high until the next good load.
- `server/routes/shopping.js:1158` - `DELETE /:listId/items/checked` still deletes whatever is checked on arrival, so A's "clear checked" can remove an item B ticked during the undo window, and A's undo restores only A's snapshot. The `deleted === count` guard makes the *mark* correct, not the *deletion*. An explicit `{ ids }` body would make both exact.

Constraint sweep otherwise clean: no `innerHTML`/`outerHTML`, no `require`, no `eval`, no `.toISOString().slice(0,10)`, no new UI strings, migration appended not edited, `db-schema-test.js` mirrors it verbatim, both new suites registered and in the chain and in `docs/test-suites.md`, `/versions` registered before `/:listId`, routes return `{ data }` / `{ error, code }` inside try/catch.

ulsklyc (approved):
Round 5, reviewed at `bdebf6c4`. **Approved.** All three blockers and both should-fix points are fixed, and every fix of this round is bound by a test. I checked each one by taking the fix out again.

## Where I measured

Your branch still needs #1144 (195) for its gap guard, so I built a throwaway merge in the order the PRs will actually land: current `main` (`fb0311ab`), then #1144 (`28752b91`), then this head. The conflicts were mechanical: the test chain in `package.json` (main added `test:housekeeping-ui`), and 195 in `server/db.js` before your 196.

On that merge, each suite on its own:

| suite | result |
|---|---|
| `migrations-append-only` | 3/0 (the gap closes with 195) |
| `schema-mirror` / `schema-reconcile` | 3/0 / 4/0 |
| `shopping-versions` | 19/0 |
| `live-feed` | 18/0 |
| `shopping-ux` | 71/0 |
| `shopping-routes` / `shopping` | 59/0 / 53/0 |
| `sw-api-cache` | 16/0 |
| `openapi-structure` / `openapi-coverage` | 11/0 / 3/0 |
| `caldav-todo-outbound` / `caldav-reminders` / `shopping-send` | 62/0 / 39/0 / 15/0 |
| `changelog` / `suite-chain` / `layer-boundary` | 28/0 / 4/0 / 3/0 |
| `frontend-audit` / `i18n` | 386/0 / 54/0 |
| `test-db.js` | exit 0 |

## Counter-proofs

Each fix taken out on that merge, from a `cp` backup, and restored afterwards (71/0, 19/0, 59/0 again at the end):

| Taken out | Suite | Result |
|---|---|---|
| 2: the `return` back in the `listsFailed` branch | `shopping-ux` | red, 70/1 - `refreshFromFeed: scheitert nur die Listenzeile, kommt der frische Bestand trotzdem auf den Schirm` |
| 3: `i.sort_order` removed from `shape` | `shopping-ux` | red, 70/1 - the `umsortiert` case |
| 4: the `WHEN` removed from `trg_shopping_items_change_au` | `shopping-versions` | red, 17/2 - `Buchhaltung bewegt die Nummer nicht` |
| 5: `allSettled` replaced by `Promise.all` with the round-4 short-circuit (the rejection of the items maps to a result, no throw) | `shopping-ux` | red, 70/1 - `scheitern die Artikel sofort und die Listenzeile SPAETER, bleibt kein Merker zurueck` |
| the `_liveFeed?.hold(...)` call in `loadItems` removed | `shopping-ux` | red, 70/1 - `die Laufnummer der Artikel-Antwort wird der Stand des Feeds` |
| `{ ids }` ignored in `DELETE /:listId/items/checked` | `shopping-routes` | red, 58/1 |

## Also checked

- **The trigger's column list matches what the page draws.** Every item field `public/pages/shopping.js` reads (`id`, `name`, `quantity`, `category`, `is_checked`, `notes`, `url`, `price_cents`, `store_id`, plus `sort_order` and `list_id`) is in the `WHEN`. `tags` live in their own table. The trade-off you wrote into the migration comment and SPEC - a new visible column needs a trigger migration - is the right one to state.
- **`IS NOT` instead of `<>`** is correct for the nullable columns (`quantity`, `notes`, `url`, `price_cents`, `store_id`).
- **Leaving two points for later is fine.** `loadLists()` without a `fromCache` guard is better done without thirty stub edits in this diff. `setItemTags()` rewriting tags on every inbound pass belongs to tasks as much as to shopping.

## On merging (not on you)

- **Order:** #1144 lands first. Then CI on this branch needs a new run against the new `main` - on this head it is still red on the gap, and as with #1095, that guard stops the `&&` chain after step 3.
- **The `package.json` test chain** now conflicts with `main` (`test:housekeeping-ui`). It is mechanical, and I will resolve it when merging.
- **Commit message:** I will squash and write it myself, as announced in round 4.

Issue #548, opened 2026-07-21
feat(shopping): default manually added items to the misc category


schneider-de-com:
## Problem

In **Kitchen → Shopping**, the category selector defaults to **Fruit & Vegetables** when adding an item manually.

For quickly creating a general shopping list, this causes unrelated items to be assigned to the wrong category unless the user changes the selection every time.

## Proposed behavior

Use **Other / Misc** as the default category for manually added shopping items.

The user can still select a more specific category before adding the item. Items imported from meal planning or recipes may continue using their detected or assigned category.

This would make quick manual entry safer and avoid unintentionally grouping most new items under Fruit & Vegetables.

PR #1120, opened 2026-09-11
feat(shopping): duplicate lists and templates, archiving, uncheck/check-all, price totals, move-between-lists and cross-category drag

### LINKED ISSUE #1113: Shopping: a suggested item name brings back the category it was last filed under
From #1103 (@nriel). Picking a previous item name from the suggestions fills in only the name. The category falls back to the default and has to be chosen again, every time. That is most of why a reusable store list feels necessary: the categories are the part that takes effort to rebuild.

## Where it drops

On `1f96c4dc`:

- `GET /api/v1/shopping/suggestions` answers with `SELECT DISTINCT name FROM shopping_items` - names only (`server/routes/shopping.js:311`).
- Choosing a suggestion sets `input.value` and nothing else (`public/pages/shopping.js:1056-1061`, the autocomplete in the item form; the keyboard path a few lines further down does the same).
- `POST /api/v1/shopping/:listId/items` stores `req.body.category || defaultCat` (`server/routes/shopping.js:822`).

## Expected

Choosing a suggestion also preselects the category that name was last filed under. The user can still change it before saving.

## Constraints

- **`/suggestions` is `/api/v1` surface** (listed in `server/openapi/paths/shopping.js`). Its `data` is an array of strings today; replacing the strings with objects would break every client that reads it. The category has to arrive additively.
- **"Last filed under" needs a rule, stated once:** the most recently updated item with that name, matched case-insensitively like the suggestion query already is, household-wide like the query itself.
- A category can be renamed or deleted after the item was filed. A suggestion pointing at a category that no longer exists falls back to the default instead of recreating it.

mclgoerg:
## Summary

Closes #1113. Related to discussion #1103 and discussion #998.

Discussion #1103 asked for a reusable shopping-list template (a store's usual walk-through order, duplicated before each trip instead of rebuilt from scratch every time). Built as proposed and agreed on the discussion, plus the batch of follow-on comfort features discussed in the same thread.

## Changes

**Core (#1103):**
- `POST /shopping/:listId/duplicate` copies a list's items in one transaction, always preserving category and manual per-category order (the actual point of duplicating), with three on-by-default flags to reset checked state and optionally strip quantities/notes. CalDAV sync columns, `added_from_meal`, `price_cents` and `store_id` are never copied - each names something true only of the original item (a synced remote object, a specific meal, a purchase that happened), never of a fresh copy.
- A list can be flagged as a template (badge in the tab bar, "Duplicate to start shopping" as its primary action) while staying fully editable.
- `GET /shopping/suggestions` now returns each name's most recently used category and quantity, not just the name, and orders by recency instead of alphabetically - closes #1113.
- Adding an item whose name already matches an open item on the list updates that row instead of creating a duplicate, merging its quantity: a bare integer increments, no quantity counts as an implicit one, and anything else (units aren't modeled) is left alone. A resubmitted quantity identical to the existing one counts the same as no quantity given, since picking the autocomplete suggestion prefills the field with the last-used value before "add" is even clicked.
- The item POST route's default category (when none is given) now matches `import-pantry`'s existing fallback to the last category, instead of the first - the original intent of #548.

**Also built in the same branch (follow-on comfort features from the same discussion):**
- A list can be archived instead of only hard-deleted, with a lightweight "Archived lists" modal to restore or delete for good.
- "Uncheck all" resets every checked item on a list for reuse on a later trip; "Check all" does the opposite for closing out a trip in one step.
- A checked-items price total appears once at least one checked item has a recorded price.
- An item can be moved to a different list (discussion #998), re-ranking it to the end of its category on the destination and resetting CalDAV sync state so it uploads as a new object if the destination is itself synced.
- An item can be dragged from one category group into another, not only reordered within one.

**Fixes found while live-testing this branch before opening the PR:** the autocomplete dropdown rendered `[object Object]` after suggestions started carrying category/quantity; quick-add's category selector stayed on a previous pick instead of resetting; "uncheck all" could look like a no-op right after a manual toggle (stale optimistic-update state wasn't cleared); duplicating a list copied the shop (`store_id`) even though `price_cents` was already correctly excluded for the same reason.

## Note for review

- **Template flag:** built and included even though the maintainer's follow-up reply on #1103 said it isn't wanted in the first cut ("every other list feature then has to decide how to treat it... easy to add later") - it was already built by the time that reply landed. Happy to drop it if you'd still rather not have it in this cut; nothing else in this PR depends on it.
- **Sequencing with #1109:** the same reply asked for this to be built on top of #1109 (touches the same files). This branches directly from `main` instead, since #1109 hasn't landed yet - rebase-ready whenever it does.

## Test plan

- [x] `npm test` passes (full suite, 300+ suites; the only pre-existing failure, `test:migrations-append-only`, is an unrelated fork-sync artifact of this fork's `origin/main` lagging `upstream/main` - reproduces identically on a clean checkout of `upstream/main` with no branch changes at all)
- [x] New route tests in `test-shopping-routes.js`, UI tests in `test-shopping-ux.js` and `test-frontend-audit.js` for every change above, including regressions for each fix found during live testing
- [x] Translated in all 24 locales (`test:i18n`, `test:i18n-plural`, `test:i18n-translated` all green)
- [x] Live-verified end to end against a deployed instance with real puppeteer screenshots for every feature, including the exact autocomplete-suggestion-prefill flow that caused the quantity bug

## Checklist

- [x] `npm test` passes
- [x] CHANGELOG.md updated


ulsklyc (changes requested):
First: thank you for the two notes at the end of the description. You flagged the template flag and the #1109 sequencing yourself instead of letting me find them, and that is worth more to me than either of those decisions. It also makes this review easier to write honestly.

I am going to ask you to split this, and I want to be clear that this is about **landing your work**, not rejecting it. 2697 lines carrying seven independent decisions means one objection holds all of it up - and I have exactly one real objection, to a part you could cut out in ten minutes.

## What I want to land now

The core of #1103 plus #1113, which is what the discussion actually agreed:

- `POST /shopping/:listId/duplicate` with the three reset flags. Your reasoning about which columns never travel (`carddav_*`, `added_from_meal`, `price_cents`, `store_id` - each naming something true only of the original) is exactly right and worth keeping in the comment.
- `GET /shopping/suggestions` carrying the most recent category and quantity, ordered by recency. That closes #1113.
- The default-category fix (last category instead of first, matching `import-pantry`). That is #548's original intent and a one-line bug fix.
- The four fixes you found while live-testing, insofar as they touch the above.

That is a PR I can review in one sitting and merge this week.

## What should be its own PR, and why

- **The template flag.** You already offered; yes, please drop it. Not because it is bad work - because "is this list a template" is a question every later list feature has to answer (does archiving a template make sense? does duplicating one? does it sync?), and I would rather answer it once, deliberately, than inherit it as a side effect.
- **Archiving, uncheck/check-all, the checked-items price total, move-between-lists (#998), cross-category drag.** Five separate features. Each is defensible, several I want - but each also has its own edge cases, and bundled they are unreviewable as one thing.
- **The same-name merge.** See below; this one has a defect, and it needs its own conversation anyway.

## The one real objection: the same-name merge loses data

`server/routes/shopping.js` (the `POST /:listId/items` block): when an open item with the same name exists, the route updates `quantity` and returns that row - and **silently discards `notes`, `url` and `category` from the request**.

Scenario: the list already has an open "Milch" with no note. Someone adds "Milch" with the note "die laktosefreie, im hinteren Regal". They get a `200` and a row back, the UI shows the item, and the note is gone. Nothing tells them. Same for a URL, and same for a category - if they filed it under a different category on purpose, that choice is dropped.

That is the kind of loss people do not notice until the shop, which is the worst time.

Two further things on the same block:

- **`COLLATE NOCASE` is ASCII-only in SQLite.** "Müsli" and "müsli" do not match, so the merge fires for `Milch`/`milch` but not for the umlaut case - it will look arbitrary to anyone whose language has them, and this codebase already cares about that elsewhere (`test:search-diacritics`).
- **The status code changes from 201 to 200** on an `/api/v1` route we have committed to. That is a contract change, not an implementation detail; it needs to be deliberate and documented in `docs/SPEC.md` and the OpenAPI path, or the route should keep returning 201.

The behaviour itself I am open to - "twice the same open item is not a user error" is a reasonable position, and your quantity heuristic (bare integer increments, no quantity means implicit one, free text left alone) is carefully thought through and well commented. But it is a change to what "add an item" means, it can lose a field, and it deserves a PR where that is the only thing under discussion.

## On sequencing with #1109

Do not rebase onto #1109 yet. It is in round 4 with three blockers open, so its shape may still move. Branch the trimmed PR from current `main` (`beb8e84c`) - #1055, #1149 and #1153 landed today, so it has moved under you again.

Migration numbers, so you can plan: `main` is at **194**. #1109 is spoken for at 195 and #1144 at 196, so take **197** and expect me to renumber it as it lands - that is normal here and does not break the append-only rule.

I know a split is annoying after the work is already done. But your trimmed PR lands in days, and right now none of it is landing at all.

PR #1129, opened 2026-09-11
feat(calendar): add local public holidays for US, CA, UK, AU, NZ (#965)

### LINKED ISSUE #965: [Bug] United States not an option in Holidays
### Yuvomi version

v2.57.4

### Deployment

### Description

Under Public & School Holidays, United States is not an option under country dropdown.

### Steps to reproduce

1. Go to /settings/modules/calendar
2. Scroll to Public & School Holidays
3. Select country drop down
4. No United States option

United States being an option on the drop down.

### Browser / device / OS

Brave on Windows 11

### Logs

```shell

```

### Additional context

_No response_

mclgoerg:
Closes #965.

## Scope note, up front

@ulsklyc's reply on #965 asked specifically for the US, as a local list "in the shape Brazil already has," in one PR. This PR does that, but also extends the same approach to four more countries (Canada, UK, Australia, New Zealand) where the same reasoning applies equally well - each one's public holidays are pure date arithmetic, verifiable against a primary source, with no state/province variation to guess at. If that's more than wanted for a first PR, the four extra countries are easy to split back out; the US half stands on its own. Flagging this proactively rather than assuming the wider scope is welcome.

## What this does

OpenHolidays (the free API behind Settings → Calendar → Holidays) lists 36 countries - mostly Europe plus Brazil/Mexico/South Africa. The US, the reported gap, isn't among them, and neither are Canada/UK/Australia/New Zealand. Rather than add a second live provider for a handful of missing countries (a second dependency that can be down or rate-limited independently of the first - explicitly the concern raised in the issue thread), all five are computed locally, in the same shape Brazil's existing hand-written fallback already used.

- **`server/services/holidays.js`**: replaced the Brazil-only fallback with a small declarative rule engine - `fixed`/`nth-weekday`/`last-weekday`/`easter-offset`/`table` rule kinds, plus per-entry weekend-observance policies (US Sat-back/Sun-forward; Canada Sunday-only; UK/NZ "Mondayisation"; a shared same-weekday lookup table for adjacent-day pairs like Christmas/Boxing Day). Brazil itself now runs through this engine - its own existing test suite passing unchanged is the parity proof for the migration, not a new assumption.
- **Five countries added**, each checked against a primary source during implementation rather than assumed:
  - **US** - 11 federal holidays.
  - **Canada** - 10 federal statutory, EN+FR names (same language cascade Brazil's PT/EN already used). Victoria Day uses its actual legal definition ("Monday preceding May 25", not "last Monday of May" - these differ whenever May 25 itself is a Monday).
  - **United Kingdom** - modelled as three complete subdivisions (`GB-ENG`/`GB-SCT`/`GB-NIR`), not a national baseline plus overlay, because the nations' holidays genuinely diverge: Scotland has no Easter Monday and its own August date; Northern Ireland adds two holidays. `getSubdivisions('GB')`/`getGroups('GB', ...)` are answered locally - GB isn't a real OpenHolidays country, so no wasted network call.
  - **Australia** - 7 national-only holidays, deliberately **no** weekend-observance shift at all: there's no Australian federal substitute-day law, each state legislates its own, so there's no single correct shifted date to compute.
  - **New Zealand** - 11 holidays including Mondayisation (per the 2013 Waitangi/Anzac amendment) and Matariki via the actual government/MBIE-published date table through 2052 - not a lunar-calendar formula. Outside that table, the code returns no date rather than guess.
- **`getCountries()`** appends a synthetic entry per local country only when OpenHolidays doesn't already list it (an API-listed country always wins, never duplicated); each synthetic entry carries `schoolHolidays: false`, since none of these five has school-holiday data.
- **Settings UI**: the school-holidays toggle disables itself with an explanation when the selected country has no school-holiday data, instead of silently syncing nothing.
- **For every other uncovered country**: an ordinary ICS calendar subscription (Settings → Personal → Calendar subscriptions) still works, and that page now says so directly.

## Deliberately not attempted

Moon-sighting or year-by-year decreed holidays (no formula exists - e.g. Islamic holidays, China's announced bridge days), and any state/province-level holidays (US states, Australian states, Canadian provinces) - a national-only list stays correct without guessing at a subdivision that varies.

## Testing

- `test/test-holidays.js`: 70/70 (30 new) - per-country fixture tests for every rule kind and observance policy, hand-verified against real dates. Notably: the same real-world Saturday (Jan 1, 2028) shifts backward to the prior Friday for the US but forward to the following Monday for the UK, both correctly, in the same test file; a documented (not accidental) Canada edge case where an unshifted Boxing Day and a Sunday-shifted Christmas can land on the same date, since Canada has no UK-style pair-avoidance rule; Matariki's exact published 2026 date (not the naively estimated one); the Matariki table's 2052 boundary returning nothing beyond it.
- `test/test-preferences-routes.js` and `test/test-settings-navigation.js` extended for the six-country dropdown addition and the new school-holiday-availability UI helper.
- Full `npm test`: exit 0, zero failures anywhere in the chain.
- Manually verified live against a throwaway local instance: real sync against the live OpenHolidays API merges correctly with all six local countries (confirmed via the actual `/preferences/holidays/countries` response); UK→Scotland subdivision selection; a live "Sync now" for the US completing successfully; and the actual calendar rendering the computed holidays on their correct dates (Labor Day on Sep 7 2026, Veterans Day on Nov 11 2026, Thanksgiving on Nov 26 2026). Checked at 390/768/1024/1440px, light and dark.

No migration - nothing schema-touching.


ulsklyc (changes requested):
The five-country scope is fine - I checked with the original decision and we are keeping all five, so please do not trim it back. US, CA, UK, AU and NZ all stay.

I verified the data rather than reading it: drove the rule engine directly and diffed its output date-by-date against primary sources for 2026 and 2027, plus the Saturday/Sunday boundary years 2022 and 2028. **Almost everything is exact** - all 11 US federal holidays including every observed shift (Independence Day 2026 Sat -> Fri 3 Jul, as decided), the UK tables byte-exact against `gov.uk/bank-holidays.json`, Canada's 10 Labour Code holidays with Victoria Day 2026 = 18 May (and your `lastMondayOnOrBefore` reading is the right one - 2026 is precisely the year where "last Monday in May" gives the wrong answer), and New Zealand line-for-line against employment.govt.nz including Matariki and the pair shifts. Brazil's migration onto the shared engine produces identical dates to the old function. That is careful work.

Two things block.

## Blocking

**1. `server/services/holidays.js:448` - St Andrew's Day is wrong, and the comment asserts the opposite of the statute.**

```js
{ names: "St Andrew's Day", rule: { type: 'fixed', month: 11, day: 30 } }
```

with a comment above saying *"St. Andrew's Day hat keine belegte Verschiebungsregel - anders als die uebrigen schottischen Feiertage bleibt er unveraendert am 30.11."*

The St Andrew's Day Bank Holiday (Scotland) Act 2007 s.1(2) moves it to the following Monday when 30 November falls on a Saturday or Sunday, and GOV.UK's own `bank-holidays.json` - the primary source this PR cites elsewhere - lists it three times: `2019-12-02`, `2024-12-02`, `2025-12-01`.

**This is live, not hypothetical.** 30 Nov 2025 was a Sunday; the Scottish bank holiday was Mon 1 Dec 2025, and the PR writes `2025-11-30`. The sync window is `currentYear-1 .. +2` = 2025..2028, so a Scottish household sees the wrong day today. Next future miss is 2030.

The fix is one word: `observance: 'mondayised'`. I confirmed that adding it keeps all 70 tests green - which is the second problem.

**2. `test/test-holidays.js` - the suite cannot catch a wrong date in the table.**

The new `#965` fixtures assert per-country *counts* plus a handful of hand-picked dates, so most of the ~55-entry rule table is unguarded. Six single-token mutations all survive `npm run test:holidays` (exit 0, 70/70):

| mutation | what it breaks |
|---|---|
| `:444` Scotland Summer `n: 1` -> `n: -1` | first vs last Monday in August |
| `:433` England & Wales Summer `n: -1` -> `n: 1` | the same, inverted |
| `:431`/`:440` Early May `n: 1` -> `n: 2` | first vs second Monday |
| `:518` NZ Labour Day `n: 4` -> `n: 3` | third vs fourth Monday |
| `:516` NZ King's Birthday `n: 1` -> `n: 2` | |
| `:480` Australia Day `day: 26` -> `day: 27` | |

The first one is the exact confusion your own comment at `:442-443` warns about ("keine Tippfehler-Variante derselben Regel"). And the test titled *"...eigener Sommertermin..."* at `:1159` contains no August assertion at all. The only St Andrew's Day assertion (`:1254`) checks the **name** in the cache, never the date - which is why the bug above shipped green.

A `assert.deepEqual` of the full expected date list per country for 2026 and 2027, transcribed from the primary source rather than from the engine, pins the table and makes the data reviewable. That is roughly 66 lines and it is the only thing standing between this feature and a silently wrong date.

## Should fix

- **`server/services/holidays.js:67-68`** - `getCountries()` has no catch, `preferences.js:1335` turns a throw into 502, and `modules-calendar.js:621` falls back to `countries = []`. So when OpenHolidays is unreachable - a self-hosted box with no egress, or the API being down - **the five new countries are unselectable even though they need no network at all**. The whole argument for computing them locally is undone by making their discoverability depend on the live service. Degrade to `Object.keys(LOCAL_COUNTRIES)` instead.
- **`public/settings/pages/modules-calendar.js:468-475`** - `applySchoolAvailability()` only ever sets `showSchool.checked = false` and never restores it. Scenario: a `DE` household with school holidays on opens Settings, looks at `United States`, switches back to `Germany`, saves - `holiday_show_school` is now persisted as `false` and the layer is silently gone, without anyone touching that toggle. Either remember and restore, or only `disabled` the input and leave `checked` alone.
- **`test/test-settings-navigation.js:894`** - `countrySchoolHolidaysAvailable` is tested as a pure function against a hand-built array. The caller (`applySchoolAvailability` and its two call sites) is not exercised at all. The helper being right is the easy half.

## Nice to have

- `holidays.js:478-486` - the deliberate "no weekend observance" choice for Australia is well argued, but its visible consequence is Christmas 2027: the calendar shows Sat 25 / Sun 26 Dec while every Australian state observes Mon 27 / Tue 28. Not wrong, but worth a line in the user-facing text.
- `holidays.js:536-541` - for the five local countries every sync still fires a live `/PublicHolidays` request known in advance to return `200 []`. Short-circuiting before the fetch would also make the offline case work by construction.
- `test/test-holidays.js:1112` - the second operand of the `||` subsumes the first, so the disjunction only tests the weaker one.
- `holidays.js:438` - Scotland's second January is named `'2 January'`; GOV.UK writes `2nd January`.

## The proclamation gap: record it, do not chase it

One-off holidays created by royal proclamation cannot come from any static table - VE Day 2020, the 2022 Platinum Jubilee (which also *moved* that year's Spring bank holiday), the 2022 State Funeral, the 2023 Coronation, and currently live a **Scotland-only World Cup bank holiday on Mon 15 June 2026**, which is inside this sync window and will be missing. New Zealand has the same class.

I am not asking you to solve that - the ICS-subscription hint is the right answer. But please add one sentence to `docs/SPEC.md` next to the "deliberately not attempted" paragraph, so the next reader treats it as a known boundary rather than an oversight, and so the next bug report about a missing UK holiday has an answer waiting.

No migration in this PR, so nothing to renumber. The trial merge onto current `main` (`beb8e84c`) is clean, and everything was green on the merged tree: `holidays` 70/70, `preferences-routes` 66/66, `settings-navigation` 85/85, `i18n` 54/54, `i18n-translated` 92/92, `frontend-audit` 384/384.

ulsklyc (approved):
Approving. Both round-3 blockers are fixed and I verified each fix in the code itself, not from the reply. Every should-fix and nice-to-have from round 3 is in as well. I have one should-fix below: extend the pinned tables you already wrote. You can do it here or in a follow-up. Nothing in this round blocks the merge.

## What I verified

Everything below was measured on a trial merge of `151b0c09` onto current `main` (`f254b57c`), not on the head alone. The head still sits on `a8067497`, 34 commits behind, and `git merge-tree` reports no conflicts.

**St Andrew's Day** - `server/services/holidays.js:464` now carries `observance: 'mondayised'`. The test at `test/test-holidays.js:1232` pins 2019-12-02, 2024-12-02, 2025-12-01, 2026-11-30 and 2030-12-02. To check the test, I removed `observance` from a `cp` backup and ran `npm run test:holidays`: exit 1, 80 pass / 1 fail, and only that test turned red. After I restored the file, `cmp` showed it identical to the backup.

**Mutations** - each one applied alone, with `npm run test:holidays` read from its own exit code:

| mutation | exit | pass / fail |
|---|---|---|
| Scotland Summer `n: 1` -> `-1` | 1 | 80 / 1 |
| England & Wales Summer `n: -1` -> `1` | 1 | 79 / 2 |
| England & Wales Early May `n: 1` -> `2` | 1 | 79 / 2 |
| NZ Labour Day `n: 4` -> `3` | 1 | 80 / 1 |
| NZ King's Birthday `n: 1` -> `2` | 1 | 80 / 1 |
| Australia Day `day: 26` -> `27` | 1 | 80 / 1 |
| *new:* `"Washington's Birthday"` -> `"Presidents' Day"` | 1 | 79 / 2 |
| *new:* Juneteenth `observance: 'us'` removed | 1 | 79 / 2 |
| *new:* CA Boxing Day `observance` removed | 1 | 80 / 1 |
| *new:* NI Battle of the Boyne `day: 12` -> `13` | 1 | 80 / 1 |
| *new:* Victoria Day `maxDay: 24` -> `25` | 1 | 79 / 2 |
| *new:* Matariki `2027: '06-25'` -> `'06-18'` | 1 | 80 / 1 |

**The expected date lists** - I checked every entry of the 2026 and 2027 tables against the primary sources: OPM for the US, gov.uk `bank-holidays.json` for all three UK nations, the Canada Labour Code list, the Australian national holidays, and employment.govt.nz. All of them match. That includes the dates most likely to be mistranscribed: US Independence Day 2026-07-03, Juneteenth 2027-06-18, Christmas 2027-12-24, Victoria Day 2026-05-18, Scotland's 2nd January 2027-01-04, NZ Waitangi 2027-02-08, and Matariki 2026-07-10 and 2027-06-25. I also checked the engine's output for 2025 and 2028 by hand and it is correct there too. US New Year's Day 2028 is observed on 2027-12-31. The engine emits that date while computing 2028, and `getForRange()` reads the cache by `start_date`/`end_date`, so the date appears where OPM lists it.

**The round-3 should-fix items, each checked by breaking the fix:**
- `getCountries()` degradation (`holidays.js:73-79`): I made the catch re-throw. `test:holidays` went to 80 / 1 and `test:preferences-routes` to 65 / 1.
- Local countries skip `/PublicHolidays` (`holidays.js:626`): I disabled the condition with `false &&`. Result 79 / 2 (the BR language test and the US no-fetch test).
- The school toggle's stash/restore (`public/settings/pages/modules-calendar.js:237`): I disabled the restore branch. `test:settings-navigation` went to 87 / 2.

**Locales** - 24 files, one inserted line each and no deletions. `git diff --stat` and `git diff -w --stat` are identical, so nothing was reserialised, and the new line uses the files' nested 4-space indent.

**Invariants, over the whole PR diff** - no `innerHTML`/`outerHTML` assignment and no `.toISOString().slice(0,10)`. The new hint text goes through `t()` and uses the existing `.form-hint` class, with no new design values. There is no migration. The CHANGELOG entry sits under `[Unreleased]` and contains no em dash. The proclamation boundary is recorded in `docs/SPEC.md:785`.

**Suites on the merged tree** - every one exited 0:
- `holidays` 81/81
- `preferences-routes` 66/66
- `settings-navigation` 89/89
- `i18n` 54/54
- `i18n-plural` 17/17
- `i18n-translated` 92/92
- `frontend-audit` 386/386
- `changelog` 28/28
- `migrations-append-only` 3/3

## Should fix

**`test/test-holidays.js:1277-1480` - the pinned years stop one year short of what users see.** The sync window is `currentYear - SYNC_YEARS_BACK .. currentYear + SYNC_YEARS_AHEAD` (`holidays.js:18-19`, `:804`), which today is 2025..2028. The tables pin only 2026 and 2027. Three single-token mutations stay green (exit 0, 81/81):

| mutation | why 2026/2027 miss it | first wrong year |
|---|---|---|
| `holidays.js:397` Thanksgiving `n: 4` -> `n: -1` | November 2026 and 2027 have only four Thursdays, so "last" equals "fourth" | 2028: 30 Nov instead of 23 Nov, inside the window today |
| `holidays.js:518` Matariki `2028: '07-14'` -> `'07-07'` | a lookup-table value outside the pinned years | 2028, inside the window today |
| `holidays.js:426` CA Remembrance Day `observance` removed | 11 Nov falls on a weekday in 2026 and 2027 | 2029 (a Sunday) |

All current values are correct, so this is a gap in the test, not a wrong date. It is the same class as the round-3 blocker, though: the Thanksgiving mutation is the historic "last Thursday" confusion. Adding a 2028 table for at least US and NZ closes the first two. A 2028 table for GB would also pin a year where 1 Jan falls on a Saturday and 2 Jan on a Sunday, which is the pair shift through the new year.

## Nice to have

- **`test/test-settings-navigation.js:911`** - the comment says these tests drive the real caller path, the country-change handler. They actually drive `createSchoolAvailabilityUpdater` with stub objects. I reverted the call site at `modules-calendar.js:511` to the round-3 inline code, which clears the toggle and never restores it: `test:settings-navigation` stays at 89/89. Now that the call site is a one-line delegate the risk is small, but the comment claims more than the test shows. Please reword it, or add a small text guard that `applySchoolAvailability` delegates to the updater. Likewise the AU hint at `modules-calendar.js:518` is untested: forcing `hidden = true` also stays 89/89.
- **`holidays.js:73-79` - the degraded country list is silent in the UI.** Previously an OpenHolidays outage returned 502 and the settings page showed the discovery error. Now it returns 200 with six countries and no hint. From reading the code, a household with `DE` saved keeps its setting: `isHolidayCountryResolved` returns false, so `resolveHolidayLocation` falls back to the saved value. The dropdown, however, shows the placeholder and `DE` is missing from the list, with no explanation. A `degraded: true` flag in the response would let the page keep its hint.
- **`holidays.js:617-626` - the short-circuit also covers BR, which OpenHolidays does list.** On 13 Sep, `/Countries` returned 36 countries including `BR`, and `/PublicHolidays?countryIsoCode=BR` for 2026 returned `[]`, so nothing changes today. If upstream ever adds BR data, though, sync keeps ignoring it while `getCountries()` says the API entry wins. Either limit the short-circuit to local countries the API does not list, or say so in the comment.
- **`docs/SPEC.md:785-789`** - the new sentence uses two em dashes. The rest of the repo's own text uses "-". Trivial.

Issue #1137, opened 2026-09-11
Add month navigation to housekeeping reports


nriel:
## Problem

The Housekeeping Reports tab displays only the automatically loaded month and has no month selector or previous/next navigation.

Historical visits are accessible only indirectly: users must switch to Staff, select an individual worker, and then change the month. This makes the aggregate report history unreachable from the report flow itself.

## Expected behavior

Users can inspect aggregate housekeeping reports for a chosen month directly from the Reports tab.

## Acceptance criteria

- The Reports tab provides an accessible month selector or previous/next month controls.
- Changing the month reloads visits and totals for that month.
- The selected month remains consistent while the report view is being used.
- Empty historical months show a clear empty state.
- Tests cover month changes, totals, empty months, and locale-sensitive month labels.

ulsklyc:
Confirmed, and most of it is already there.

`GET /housekeeping/visits` accepts `?month=YYYY-MM` and totals that month, and the visit log in the Staff tab already passes it from its month field. The Reports tab calls the same endpoint without the parameter (`loadData()` in `public/pages/housekeeping.js`), so it can only ever show the current month.

One trap for whoever builds it: every action on the page, marking a visit paid or saving an edit, reloads through `loadData()`. If the chosen month lives only in the control, the report jumps back to the current month after the first action - your third criterion. The selected month has to be page state that the reload reads.

Which control it gets is still open on our side: the Staff log uses a plain month field, other modules a previous/next stepper, and the Reports tab should not become a third variant.

Issue #1135, opened 2026-09-11
Hide or explain housekeeping visit actions that the current user cannot perform


nriel:
## Problem

The staff visit log always renders edit and delete actions, including for paid/settled visits.

The backend intentionally permits only administrators to edit, delete, or pay an already settled visit. A non-admin member can therefore select visible actions that inevitably fail with `403 Permission denied`.

## Expected behavior

The UI should reflect the effective permission before the user starts the action.

## Acceptance criteria

- Paid visits do not expose edit/delete actions to users who lack permission, or the actions are disabled with a clear explanation.
- Administrators retain the permitted actions.
- Unpaid visits continue to expose the actions allowed to regular members.
- Server-side authorization remains authoritative.
- UI tests cover admin/member behavior for paid and unpaid visits.

ulsklyc:
Confirmed at the code, thanks.

The server rule is `assertMayTouchSettled()` in `server/routes/housekeeping.js`: once a visit is paid, editing it, deleting it and paying it again need an admin. The page does not know that rule. In the Staff tab's visit log every row renders edit and delete whatever the payment state and whoever is looking; only the pay button reacts, and only to `paid_at`. The recent visits on the Overview offer edit the same way. A member who opens a paid visit fills in the form and learns at save that it was never allowed.

The direction: the server keeps the rule and tells the page per visit what the current user may do, the way the calendar already sends capability fields such as `can_override_occurrence` next to an event, rather than the page copying the admin check. Otherwise the two drift apart the first time the rule changes. Your acceptance criteria stand as written, including tests for both roles on paid and unpaid visits.

PR #1144, opened 2026-09-12
fix(contacts): use the misc category key for auto-created member/guest contacts

### LINKED ISSUE #1140: [Bug] Contacts header is in german despite rest of app in english
### Yuvomi version

v2.65.2

### Deployment

TrueNAS SCALE (catalog app)

### Description

On the "Contacts" Page the header reads "SONSTIGES" Which I believe is misc in German.  This should be an english equivalent for english users.

### Steps to reproduce

Click more -> contacts, look at the header of the div for people/contacts.

Actual: "Sonstiges"
Expected: English equivalent

### Browser / device / OS

IOS 26

### Logs

```shell

```

### Additional context

_No response_

mclgoerg:
## Summary
- Household member and guest accounts get an auto-created/mirrored contact
  (`syncFamilyMemberArtifacts` in `server/auth.js`, `syncGuestArtifacts` in
  `server/routes/split-expenses.js`). Both fell back to the legacy German
  literal `'Sonstiges'` for the contact's category instead of the canonical
  `misc` key.
- Since `'Sonstiges'` isn't a key in `contact_categories`, the contacts page
  couldn't resolve a translated label for it and rendered the raw key
  verbatim — showing "Sonstiges" regardless of the user's locale.
- Added migration 194 to backfill existing contacts still carrying the
  literal `'Sonstiges'` category over to `misc` (mirrors the CardDAV-only
  backfill added in migration 91, extended to cover this path too).
- `misc`'s label (`contacts.categoryOther`) is defined in all 24 locale
  files, so the fix applies to every supported language, not just en/de.

Fixes #1140

(Reopened from #1142, which used a `claude/...` branch name by mistake.)

## Test plan
- [x] `node --experimental-sqlite --test test/test-family-contacts.js test/test-carddav.js`
- [x] `node --experimental-sqlite --test test/test-split-expenses-routes.js test/test-family-routes.js test/test-contact-categories.js`
- [x] `npm run test:migrations-append-only`
- [x] `npm run test:schema-reconcile`


ulsklyc (changes requested):
Thanks — the diagnosis is right and the fix is in the two places that write the mirrored contact. Three things before this can land.

**1. The migration number has to move: 194 is already promised to #1055.**

`main` is on 193, and three open PRs all reach for 194. The order we committed to publicly is:

| PR | migration |
|---|---|
| #1055 (recurring occurrence overrides) | **194** — promised to land first and keep this number |
| #1109 (shopping live updates) | 195 |
| **this PR** | **196** |

Please renumber to 196. (Migrations are append-only here — see CLAUDE.md — so this is purely about picking the next free slot, nothing else changes.)

**2. The migration draws a wider line than the one it cites.**

The comment names migration 91 as the precedent, but 91 deliberately restricts itself:

```sql
UPDATE contacts SET category = 'misc'
  WHERE carddav_uid IS NOT NULL AND category = 'Sonstiges';
```

with the comment *"Nur Sync-Kontakte, damit manuell angelegte Kontakte unberührt bleiben."* This PR heals every row instead:

```sql
UPDATE contacts SET category = 'misc' WHERE category = 'Sonstiges';
```

I think the wider heal is actually defensible — the column carries `DEFAULT 'Sonstiges'`, and the contact form only offers keys from `contact_categories`, so a user cannot type that raw value themselves. But then it should *say* that, rather than cite 91 as precedent and quietly go further. Either add a sentence explaining why the restriction from 91 no longer applies, or narrow it to the rows this issue is actually about:

```sql
UPDATE contacts SET category = 'misc'
  WHERE family_user_id IS NOT NULL AND category = 'Sonstiges';
```

Your call which — just make the choice visible in the comment.

**3. No test.**

Three files, none of them a test. Nothing holds the two mirrored write paths (`server/auth.js:492`, `server/routes/split-expenses.js:185`) to writing `'misc'`; the next refactor of `syncFamilyMemberArtifacts` can put the German literal back and every suite stays green. A small one that creates a household member and asserts the resulting contact row carries `misc` would cover both paths.

One thing I checked and want to rule out explicitly, so nobody "fixes" it later: `server/db-schema-test.js:448` still writes `'Sonstiges'`, and that is correct — it mirrors historical migration 23 for test databases and must not be touched.

ulsklyc (changes requested):
Thanks, all three points are answered, and the comment on the migration now says what it does. Two small things before this can merge, one of them a change of plan on our side.

## Change of plan: please take 195, not 196

When I gave you 196, 195 was held for #1109. #1109 still has open blockers from its last round, and its schedule is not yours to wait on: as long as 195 is missing on `main`, `test:migrations-append-only` stays red here, and #1140 is a visible bug in every non-German household. So the numbers swap: **this PR takes 195, #1109 moves to 196.** I am telling ter9ng on #1109 at the same time.

Concretely: `version: 196` -> `version: 195` in `server/db.js`, nothing else references the number. With that, the gap guard goes green on the merge with current `main` (`f254b57c`), which merges cleanly with your head.

## Should fix

**A CHANGELOG entry.** The PR changes what users see (the contacts header and every mirrored member or guest contact), so it needs an entry under `[Unreleased]` / `### Fixed` (CONTRIBUTING, "Changelog"), with the bolded lead-in `test:changelog` checks. Something along the lines of: household members and guests created as contacts now show the translated "Other" category instead of the German "Sonstiges", and existing ones are corrected on update (#1140). Hyphens, not dashes.

## Verified

- **The migration's boundary.** I checked that nothing can still write the raw value. The remaining inserts into `contacts` either pass a category through `FALLBACK_CATEGORY = 'misc'` (`server/routes/contacts.js:19`) or resolve it through the CardDAV category mapping, and with this PR the two mirror paths write `'misc'`. The column default `'Sonstiges'` is never reached by an insert without a category. The wide heal is safe as argued.
- **Both write paths are bound.** `test:family-contacts` 3/0 and `test:split-expenses-routes` 72/0 on your head. You showed the counter-proof for the member path. I ran the guest path: `'Sonstiges'` put back into `syncGuestArtifacts` (`server/routes/split-expenses.js:185`) turns the new guest test red (71/1), and restoring it turns it green.
- `server/db-schema-test.js:448` stays on `'Sonstiges'` as the mirror of migration 23, correct.

Nice to have, not needed here: a test that runs the migration over a row with `category = 'Sonstiges'` and checks it comes out `'misc'`. The heal itself is currently unbound.

ulsklyc (approved):
Thanks @mclgoerg - all of it holds on `28752b91`, approved.

## Verified

- **195 is the right slot.** `main` ends at 194 (#1168, #1170 and #1169 brought no migration), and nothing else in the branch names the number.
- **On a merge with current `main` (`fb0311ab`)**: `test:migrations-append-only` 3/0 (the gap guard is green now), `test:contact-categories` 19/0, `test:family-contacts` 3/0, `test:split-expenses-routes` 72/0, `test:carddav` 147/0, `test:family-routes` 4/0, `test:schema-reconcile` 4/0, `test:changelog` 28/0, `test-db.js` exit 0.
- **The heal is bound for real.** My own counter-proof on that merge: narrowing the v195 `UPDATE` to `... AND name <> 'Gast'` turns `test:contact-categories` red for the right reason (`Gast: erwartet misc, war Sonstiges`, 18/1); restored, 19/0. One thing to know, no change needed: the probe runs over the v1 + v84 schema only, so a mutation that names a later column (I first tried the migration-91 narrowing `carddav_uid IS NOT NULL`) goes red on `no such column`, not on the assertion.
- CHANGELOG bullet reads right, hyphens only, `git diff --check` clean.

## The one conflict left

`main` gained its own `## [Unreleased]` since your rebase (#1154, #1137 under `### Added`, #1135 under `### Fixed`), so `CHANGELOG.md` conflicts again. It is purely mechanical: your bullet goes under the existing `### Fixed`, after the #1135 bullet. I will resolve that when merging, nothing to do on your side.

PR #1145, opened 2026-09-12
fix(housekeeping): show an error when a visit deep link fails to open

### LINKED ISSUE #1139: Show an error when a housekeeping visit deep link cannot be opened
## Problem

Opening Housekeeping with `?editVisit=<id>` fetches the requested visit. If it is missing, invalid, or unauthorized, the error is silently ignored and the user lands on the normal dashboard without any explanation.

This makes stale links, permission problems, and malformed URLs indistinguishable from successful navigation.

## Expected behavior

The application should explain that the requested visit could not be opened while keeping the Housekeeping page usable.

## Acceptance criteria

- A missing visit produces a localized “not found” message.
- An unauthorized visit produces an appropriate access message without leaking visit details.
- Other request failures offer a retry where useful.
- The invalid `editVisit` state is removed or neutralized so subsequent renders do not repeat the failed request.
- Tests cover valid, missing, invalid, and unauthorized deep links.

mclgoerg:
## Summary

- `?editVisit=<id>` (used by reminder/calendar notifications) fetched the target visit but swallowed every failure in an empty `catch`, leaving the user on the plain dashboard with no way to tell a stale/broken link from a working one.
- A missing (404) or inaccessible (403) visit now shows a localized toast with no retry (retrying the same id would just fail again the same way), without leaking any visit details in the access-denied case.
- Any other 4xx (e.g. a malformed id, 400) is likewise treated as a terminal state: a localized, generic message with no retry - not the raw, unlocalized server text, and not a retry loop that would fail identically forever.
- A genuine 5xx or network failure offers a retry that re-fetches the same visit.
- The broken `editVisit` id is cleared from the URL as soon as the fetch fails, so a reload or back-navigation doesn't repeat the failed request.

## Test plan

- New `test/test-housekeeping-editvisit.js` (`npm run test:housekeeping-editvisit`, 10 tests): the 404/403/400/5xx/network classification, toast wiring, URL cleanup, and that a retry click re-fetches the same id. The success path (opening the edit modal) goes through the shared modal system, which needs a real DOM - consistent with how other modal flows in this repo are verified live rather than unit-tested.
- Full `npm test` green.
- Live-verified in a throwaway browser instance against all four cases:
  - Old code vs. new code, same missing-visit link: before, nothing happens and the broken `?editVisit=999999` stays in the URL; after, a "The entry was not found." toast appears and the URL is cleaned.
  - A malformed id (`?editVisit=not-a-number`, server 400): a generic, localized "An error occurred." toast, no retry button, URL cleaned.
  - A valid deep link still opens the edit modal correctly (unaffected).
  - The unauthorized (403) case is covered by the unit tests; a real end-to-end repro wasn't reachable because the client router already redirects away from a fully module-denied page before this code runs (an existing, separate guard) - the fix still handles it defensively, e.g. for a race between page load and access revocation.

Closes #1139


ulsklyc (changes requested):
Thanks @mclgoerg - the core of the fix is right: the empty `catch` is gone, `err.status` is the correct field (`ApiError` in `public/api.js`), a network failure without a status gets the retry, and the toast wiring matches `showToast(message, type, duration, action)`. `common.errorGeneric` and `common.retry` exist in all 24 locales. I also ran two counterproofs against the suite: making 403 retryable turns 2 tests red, dropping the URL cleanup turns 3 red. So the tests do bind.

A few things before this can go in.

## Blocking

**1. Rebase onto main, and the CHANGELOG bullet has to move.** The PR conflicts in `package.json`, `docs/test-suites.md` and `CHANGELOG.md` (`housekeeping.js` itself merges cleanly). The first two are mechanical: take main's `"test"` chain and add `npm run test:housekeeping-editvisit` after `test:housekeeping-routes`; keep main's rewritten `test:housekeeping-routes` line in `docs/test-suites.md` and add yours below it. The CHANGELOG is not mechanical: v2.66.0 shipped this morning and moved everything that was under Unreleased into `## [2.66.0]`, so git puts your bullet inside the already-published 2.66.0 notes. It belongs under `## [Unreleased]` with its own `### Fixed` heading.

## Should fix

**2. The handler keeps running after the user has left the page.** `render(container)` (`housekeeping.js:1278`) ignores the second argument, but the router passes `{ user, signal }` and aborts that signal when the route is replaced (`public/router.js:1575`). The visit fetch is awaited after `loadData()`, so a user who taps away while it is in flight is already on another page when it fails. Then `history.replaceState(null, '', location.pathname)` runs against *that* page's URL, dropping its query and the router's `{ path }` state, and the error toast appears over the wrong page. On a 5xx the retry toast stays for 6 s and `onClick` calls `openVisitFromDeepLink()` again, which can open the housekeeping edit modal on top of an unrelated page. Please take `context`, pass the signal down, and bail out when `signal?.aborted`: before the cleanup and toast, and again inside the retry `onClick`.

**3. 429 is treated as permanent.** `const isTransient = status == null || status >= 500;` (`housekeeping.js:1249`) makes a 429 final: no retry, generic message, link removed from the URL. It is a real case here: `apiLimiter` in `server/index.js` allows 300 requests per minute per IP on `/api/`, and every device of a household behind one NAT shares that IP. The reasoning in the comment ("a retry only returns the same error") does not hold for 429, and `api.js` even carries `Retry-After` on the `ApiError`. Treat 429 like 5xx, and add a test for it: right now flipping it keeps all 10 tests green.

**4. The URL cleanup wipes more than the deep link.** `history.replaceState(null, '', location.pathname)` drops every query parameter, the hash and `history.state`. The two other places that strip callback parameters remove only their own and keep the state: `public/settings/pages/sync-calendar.js:1472` and `public/settings/pages/documents-storage.js:558`, both `history.replaceState(history.state, '', url.pathname + url.search + url.hash)` after deleting their keys. Housekeeping only reads `editVisit` today, so nothing breaks yet, but the test at `test-housekeeping-editvisit.js:90` (`[null, '', '/housekeeping']`) locks the wipe in. Please delete just `editVisit` from a `URL` and keep `history.state`, and assert that.

**5. The CHANGELOG describes behaviour the code does not have.**
- "A reminder or calendar notification's `?editVisit=<id>` link": the only places that build this link are the two calendar event taps (`public/pages/calendar.js:3775` and `:4264`). No reminder or notification builds it. The code comment at `housekeeping.js:1233` says the same and needs the same fix.
- "a visit you no longer have access to says so": `GET /housekeeping/visits/:id` has no per-visit access check. It returns 400 for an invalid id, 404 for a missing visit, or 500. The only 403 on that path is the module gate in `server/index.js`, and as your PR body says, the client router already redirects away before this code runs. Handling 403 defensively is fine, but the changelog should not promise it as a user-visible case.
- "any other failure offers to try again": since the second commit every 4xx is final, including the "invalid" link the same bullet mentions. After point 3 it is "a server error, a network problem or rate limiting offers to try again".

## Nice to have

- `test-housekeeping-editvisit.js:60`: `assert.ok(!message.includes('42'), ...)` can never fail. The id is never passed to `describeDeepLinkError()`, and the `friendlyError` stub returns a fixed string. The exact-message check in the `openVisitFromDeepLink` 403 test is what actually guards against leaking the id, so this line can go.
- `test-housekeeping-editvisit.js:29-33`: the 403/404 messages come from a local stub, so the real mapping in `router.js` (`common.errorForbidden` / `common.errorNotFound` / `common.errorServer`) is not exercised. Worth a sentence in the file header so nobody reads it as covered.
- `test-housekeeping-editvisit.js:166`: the test named "erfolgreiche Antwort" stubs `{ data: null }`, which is the empty-payload branch, not a successful open. A name like "leere Antwort" would match what it checks.

ulsklyc (changes requested):
Thanks @mclgoerg - all five points and the three nice-to-haves are in, and they hold on `240fb478`: the abort checks sit before the URL cleanup, inside the retry `onClick` and before the modal; 429 offers the retry; only `editVisit` leaves the URL and `history.state` stays; the CHANGELOG and the comment now match the code.

One thing happened after your push, though: #1170 (the #1135/#1137 housekeeping work) was merged into main, and it touches the same deep-link code. The PR conflicts again, in `CHANGELOG.md`, `docs/test-suites.md`, `package.json` and `public/pages/housekeeping.js`, and two of those conflicts are not mechanical.

## Blocking

**1. The deep link now has two targets, and taking your side drops one without a sound.** On main, `render()` opens the edit form only when the server allows it, and the visit report otherwise (`housekeeping.js:1469-1470` on main):

```js
if (visit.can_edit) openVisitEditModal(visit, content);
else openVisitReportModal(visit, null, { onRefresh: () => renderCurrentTab(container) });
```

Your side replaces that block with `await openVisitFromDeepLink(...)`, whose success path still calls `openVisitEditModal` unconditionally. I resolved the conflict that way on a scratch copy and ran `test:housekeeping-editvisit`, `test:housekeeping`, `test:housekeeping-routes`, `test:housekeeping-ui`, `test:frontend-audit`, `test:changelog` and `test:suite-chain`: all green (15/13/63/16/386/28/4). Nothing catches it, and a member tapping a paid visit in the calendar would get the form that fails at save again - the bug #1135 just fixed. Please move the branch into the success path of `openVisitFromDeepLink()`, after the `signal?.aborted` check. With that, the same seven suites are green too.

Please also add a test for that branch, so it does not depend on someone reading the diff. The file header says the success path needs a real DOM, but under the loader `openModal` is a no-op stub (`test/test-browser-loader.mjs:106`). If it handed off to `globalThis.__openModal?.(...args)` - the way `promptModal` and `confirmOverModal` in the same stub already hand off to a global - the test could assert the title key: `housekeeping.editVisit` for `can_edit: true`, `housekeeping.visitReportDetails` for `can_edit: false`. I have not run this: `openVisitEditModal` reads `state.workers` and `visit.check_in` before it reaches `openModal`, so the fixture has to provide what those need. If it works out, the sentence about the real DOM in the header and in `docs/test-suites.md` goes.

**2. `__test` would exist twice.** Main already ends `housekeeping.js` with an `export const __test = { ... }` for `test-housekeeping-ui.js`. Keeping both declarations is a syntax error when the module loads, so please add `describeDeepLinkError` and `openVisitFromDeepLink` to main's object instead.

**3. CHANGELOG: one `### Fixed`, not two.** `## [Unreleased]` on main now carries `### Added` (#1154, #1137) and `### Fixed` (#1135). Your bullet goes under that existing `### Fixed`, after the #1135 bullet.

`package.json` and `docs/test-suites.md` conflict the same way as last time and resolve the same way. Still no migration, nothing to renumber.

ulsklyc (approved):
Thanks @mclgoerg - all three points hold on `33621371`, approved.

## Verified

- **The branch is where it belongs.** `openVisitFromDeepLink()` checks `signal?.aborted` first and only then chooses between `openVisitEditModal` and `openVisitReportModal` by `visit.can_edit`; `render()` just awaits it. One `__test` object, and the CHANGELOG bullet sits under the existing `### Fixed`.
- **On a merge with current `main` (`fb0311ab`, merges cleanly)**: `test:housekeeping-editvisit` 17/0, `test:housekeeping` 13/0, `test:housekeeping-routes` 63/0, `test:housekeeping-ui` 16/0, `test:frontend-audit` 386/0, `test:browser-loader-stubs` 3/0, `test:changelog` 28/0, `test:suite-chain` 4/0, `test:i18n` 54/0.
- **The new tests bind.** My counter-proof on that merge: the `can_edit` branch replaced by an unconditional `openVisitEditModal(visit, content)` turns `test:housekeeping-editvisit` red on exactly the `can_edit=false` case (title `housekeeping.editVisit` instead of `housekeeping.visitReportDetails`, 16/1); restored, 17/0. The `globalThis.__openModal` hand-off in the loader stub is the right shape.
- No dashes in the added lines.

Nothing left on your side. When merging I will squash and write the commit message myself, since the single commit's subject now describes the rebase rather than the fix.

PR #1146, opened 2026-09-12
Waste page UX: labelled add-type entry point, unified row actions, curated type colors

mclgoerg:
> [!IMPORTANT]
> **Stacked on #1095** (`feat/waste-collection`) — merge that first. This branch is one commit on top of its current head (`478e6488`) and touches only `waste.js`/`waste.css`, tests, locales, and the changelog. No migrations.

Follow-up UX pass over the Waste page from an app-wide UI/UX review, focused on making the page actionable on a fresh install and consistent with the rest of the app.

### What changed

- **A visible, labelled way to add a type.** The header actions were one unlabeled `⋯` menu; the FAB is "Add pickup", which on a fresh install dead-ended in an "add a type first" toast. Now: a labelled **Add type** `btn--secondary` sits in the toolbar next to the docked FAB (the FAB stays the page's one primary — a `btn--primary` was tried first and reverted, since the docked FAB already renders as a filled primary at ≥1024px and two identical violet pills side by side violate the one-primary rule; the `toolbar-new-btn` mechanism was also evaluated and rejected because it *hides* the FAB at desktop widths, and Add type ≠ Add pickup). The FAB's no-types branch now opens the type modal directly, with the explanatory toast kept.
- **Sources empty state has a CTA** (opens the import flow), matching the Types empty state.
- **Section hierarchy fixed.** The three `<h2>`s rendered smaller and greyer than the rows they govern; they now carry the shared `u-section-title` role.
- **One trailing-action grammar.** Type cards, schedule rows, and source rows each had their own mix of bare icons; all four row types now use the standard `⋯` popover menu, flush at the row's trailing edge and vertically centered. This also fixed a real markup bug: the rows used `.list-row__main--interactive` *without* the base `.list-row__main` class, which left the menus floating mid-card and stacked each row's icon above its name (alignment pixel-measured before/after at 320/360/768/1280px).
- **`.waste-badge--paused` finally has a CSS rule** (it was emitted but unstyled, so a paused schedule looked informational).
- **Curated color palette** replaces the raw `<input type="color">` for waste types: 10 named swatches with a shared luminance band (the old picker allowed pure white/black), presets aligned to it. The 11 new `waste.color*` keys are translated in all 24 locales — every name harvested from the color-name sets already shipped in this app (calendar/notes/budget), none invented.

The read-only member path is preserved throughout: the toolbar button, both CTAs, the FAB branch and every row menu stay behind the existing read-only gates (omitted from markup, not hidden).

### Verification

`test:waste-ui` 48/48 (18 new: popover conversions incl. boundary-arrow omission, paused badge, palette invariants, primary-button absence contract, read-only conditioning), all other `test:waste-*` suites, `test:frontend-audit` 383/383, `test:typography`, `test:i18n`/`-plural`/`-translated`, `test:changelog` — all green. Browser-verified (seeded + fresh-install instances) at 390/768/1024/1440 in light and dark.


ulsklyc (changes requested):
Thanks @mclgoerg - this is a real improvement to a page that was hard to start with. Reviewed at `83d623b9`, which is `main` plus your commit plus my Node 22 test fix from the rebase. Two small things need fixing before merge. Everything else holds.

## Should fix

**1. Every existing waste type opens with a duplicate "Current colour" swatch** (`public/pages/waste.js`, `swatchColors` in `openTypeModal()`)

Types saved before this PR came from `<input type="color">`, and per the HTML spec that input's value is **lowercase** hex. The server's `color()` validator (`server/middleware/validate.js:143`) only checks `/^#[0-9A-Fa-f]{6}$/` and stores the value unchanged, so those rows hold `#16a34a`, not `#16A34A`. This applies to preset-based types too, because the preset set the input's value and the browser lowercased it.

`WASTE_TYPE_COLORS.includes(selColor)` is case-sensitive. When such a type is edited:
- none of the ten palette swatches is marked active;
- an eleventh swatch labelled "Current colour" appears, and it looks exactly like one of the ten.

That is every type anyone created before this change. Please compare case-insensitively, for example by normalising `selColor` with `toUpperCase()` before the `includes` and the `c === selColor` checks, so the palette's spelling wins. A test with `'#16a34a'` would bind it.

**2. Two radios with the same name in `fa` and `ko`** (`public/locales/fa.json`, `ko.json`)

`waste.colorTeal` and `waste.colorCyan` are both `فیروزه‌ای` in fa and both `청록` in ko. They become the `aria-label` of two different radios in the same group, so a screen reader announces two identical choices. Other locales keep them apart, for example ru `Бирюзовый`/`Голубой` and zh `青绿色`/`青色`.

## Nice to have

- **The chosen swatch on save is not bound by a test.** Counter-proof on the PR head: I changed the save body to always send the colour the dialog opened with (`color: selColor`), and `test:waste-ui` stays at 49/0. Notes and the calendar read their swatch the same way without a test, so this is not specific to your PR. Still, a tiny exported reader (`activeSwatchColor(panel, fallback)`) would make it bindable with a stub panel.
- **Row menu triggers.** `popoverMenuHtml()` gives its button `aria-haspopup="menu" aria-expanded="false"`. The three new row buttons lack both, and so does the pickup row's existing button on `main`. Adding them to all four would announce a closed menu as a menu.
- `WASTE_TYPE_COLOR_NAMES()` builds the whole map once per swatch. Build it once before the `map`.
- **Colour names still in English:** "Teal" in cs, fil, sv and tr. This is allowed and `test:i18n-translated` passes, but cs "Modrozelená", sv "Blågrön" and tr "Turkuaz" exist.
- **CHANGELOG:** the new sentences now extend the long Waste bullet, and they mix "colour" and "color". A separate bullet would read more easily.

## Verified

- **Suites on the PR head:** all nine `test:waste-*` (`waste-ui` 49/0), `frontend-audit` 386/0, `i18n` 54/0, `i18n-translated` 92/0, `i18n-plural` 18/0, `changelog` 28/0, `suite-chain` 4/0, `typography` 17/0. CI is green on Node 22 and 24.
- **Counter-proofs**, each from a `cp` backup and restored to 49/0:

| Mutation in `waste.js` | `test:waste-ui` |
|---|---|
| move-up entry offered at every position | red, 47/2 |
| FAB without a type no longer opens the type dialog | red, 48/1 |
| sources empty state without its CTA | red, 48/1 |
| type row menu no longer behind the read-only gate | red, 48/1 |
| save always sends the opening colour | **green, 49/0** (see Nice to have) |

- **Menus close before the action runs.** `installPopoverMenus(_container)` hides the enclosing `.popover-menu` in a capture-phase click handler. That covers re-rendered rows, and `add-schedule`/`delete-schedule` never open a dialog under a still-open menu.
- **Palette as a JS constant** matches `NOTE_COLORS`, `EVENT_COLORS` and `ACCOUNT_COLORS`. The CSS uses tokens, `.waste-badge--paused` stays neutral as argued, and the removed `.waste-occurrence-row__icon` is gone from both JS and CSS.
- **Escaping:** type and source names, the schedule summary, icons and colour values go through `esc()`. No `innerHTML`.
- **i18n:** 11 new keys in all 24 locales, only those lines touched, 4-space indent kept. `waste.colorCurrent` exists everywhere.

I read the layout claims (row alignment, tap targets) in the CSS but did not check them in a browser myself.

ulsklyc (approved):
Re-review on `64f21159`: approving.

All seven points from the round at `83d623b9` are in, and I checked each one against the code path, not just the tests:

- **Swatch match.** `resolveSwatchColors()` compares in upper case and leaves the palette's own spelling in `data-color`/`style`; a stored `#16a34a` opens on its palette swatch without an eleventh one. Saving that type writes the palette spelling back, which is fine.
- **fa/ko.** `colorCyan` and `colorTeal` are distinct in both files now.
- **`aria-haspopup`/`aria-expanded`.** The static `aria-expanded="false"` is the right starting value: `installPopoverMenus(_container)` (`waste.js:1949`) catches `toggle` in the capture phase, and `onToggle()` in `utils/popover-menu.js` sets `aria-expanded` on the `[popovertarget]` trigger of any `.popover-menu`, so all four row menus report open and closed correctly.
- **CHANGELOG.** The new bullet sits under `[Unreleased]/Added`, and the `## [2.66.0]` section is byte-identical to the `v2.66.0` tag. I compared it by hand because this branch's base (`8b298415`) predates #1183, so the new frozen-section guard has not run here yet. It will run on the merge commit.

Isolated run on the head, 16 suites, all green: `waste-domain` 56, `waste-migrations` 38, `waste-routes` 57, `waste-ui` 55, `waste-import` 28, `waste-dashboard` 23, `waste-calendar` 31, `waste-url-source` 20, `waste-reminders` 11, `frontend-audit` 386, `i18n` 54, `i18n-translated` 92, `i18n-plural` 18, `changelog` 27, `suite-chain` 4, `typography` 17.

One note, not blocking: the save test binds `activeSwatchColor()` itself, not the call site. Changing `color: activeSwatchColor(panel, selColor)` back to `color: selColor` leaves `test:waste-ui` at 55/0. A source assertion that the save body calls `activeSwatchColor(` would close that. Your call whether it is worth a line here.

PR #1147, opened 2026-09-12
Schedule UX: stable Today card, consolidated add affordances, sticky Compare headers

mclgoerg:
> [!IMPORTANT]
> **Stacked on #1099** (`fix/schedule-audit-findings`) — merge that first (which itself follows #1095, per its own PR body). This branch is one commit on top of #1099's current head (`910914ec`). No migrations, no new strings.

Batch 4 of an app-wide UI/UX review — hierarchy and action cleanup for the Schedule page, on top of the audit rewrite in #1099.

### What changed

- **The Today card no longer jumps.** It rendered only on the Shift-types/Planning tabs, as the first block inside the per-tab body — so switching to Statistics or Compare removed the page's first block and everything reflowed. It now lives in a persistent slot between the tab rail and the tab content and renders on every tab. The `inUse` guard is unchanged (a fresh household still sees no double empty states — the deliberate behavior from the earlier audit), and the S-07 default-tab logic is untouched. Placement note: below the tab rail rather than above it, because the tab rail is a required direct child of the collapsing `.page-toolbar` chrome — a data card inside that header would collapse/stick like navigation.
- **One durable add path on Planning.** Overrides and Extras each carried an always-visible header add-button *in addition to* their own empty-state CTA and the page FAB — an empty section showed two identical buttons stacked. The header buttons are gone; the FAB remains the durable path and each empty-state CTA the contextual one, exactly how the Patterns section above them already worked.
- **Quick-start presets are buttons, not a segmented control.** The three one-shot template actions used `.segmented`, which implies mutually-exclusive state (none ever carried `is-active`/`aria-pressed`). Now plain `btn--secondary` buttons in a wrapping row; the household template-visibility filter is unchanged.
- **Compare day headers are sticky on the block axis**, same technique the gutter already uses on the inline axis, so a block mid-scroll can always be attributed to its day. The week/day density toggle next to them (a real state toggle) stays a `.segmented`.

### Verification

`test:schedule` 108/108 (4 new structural tests, 2 updated), `test:schedule-feed`, `test:schedule-extras`, `test:schedule-reminders`, `test:frontend-audit` 383/383, `test:settings-navigation` — all green. (`test:changelog` has a pre-existing failure on this branch's base — the 2.65.2 release tag — unrelated to this change and reproducible without it.) Browser-verified across all four tabs at 390 and 1440, light theme, zero console errors.


ulsklyc (approved):
Review on your commit (`1f3796a7`, rebased locally onto #1099's `8ca21847`).

**The rebase.** Two conflicts, both from #1099's last round:

- `CHANGELOG.md`: #1099 moved its second-pass paragraph out of `[2.66.0]`, so your bullet moves to `[Unreleased]/Fixed`, right after #1099's. `[2.66.0]` stays identical to the tag.
- `renderShell()`: #1099 gave `.schedule-body` its tabpanel wiring (`id`, `role="tabpanel"`, `tabindex`). Resolved as your slot followed by that body.

Your Today-slot test then needed two adjustments. Neither changes what it protects. The first regex expected a bare `<div class="schedule-body"></div>`. The second check ("nothing named schedule-body before the slot") now tripped over #1099's `aria-controls="schedule-body"` on the tab buttons, so it looks for `class="schedule-body"` instead. Counter-proof: moving the slot below the body turns the test red again.

After the rebase, 17 suites are green in isolation: `schedule` 112, `schedule-preferences-gate`, `tablist`, `dashboard` 112, `schedule-feed`, `schedule-extras`, `schedule-reminders`, `schedule-custom-fields`, `frontend-audit` 386, `i18n`, `i18n-translated`, `i18n-plural`, `changelog`, `suite-chain`, `typography`, `migrations-append-only`, `schema-mirror`.

**The change itself, checked at the code path:**

- Printing Statistics is unaffected. The print sheet already hides `.schedule-today`, so the card does not come back on paper now that it renders on every tab.
- Removing the section-head add buttons leaves no gap. `updateScheduleFab()` shows the FAB on Shift types and Planning for everyone who could use those buttons, and hides it for read-only members, as the buttons were hidden for them too.
- The quick-start row and the sticky Compare head use tokens only.

**The Today card on Statistics and Compare.** The old page left it out on those two tabs on purpose, as redundant next to a week grid and a statistics summary. We are fine with reversing that: a first block that comes and goes with the tab moves everything below it, and "what is today" does not depend on the tab you are on. So no change needed.

Approving. #1099 has landed, so I rebased your commit onto current main and force-pushed it to this branch (the same resolution as above, nothing else changed). **Before you push to this branch again, please reset to the pushed state:**

```
git fetch origin && git reset --hard origin/fix/schedule-ux
```

PR #1148, opened 2026-09-12
Split-expenses: one primary action, section heading when embedded, destructive delete styling

mclgoerg:
From an app-wide UI/UX review; against current `main`, no migrations, no new strings.

The "Today"/"Current" period-navigation half of the original PR has been removed as requested in review, with the original placement — and the comment recording that decision — restored verbatim. The position question now lives in #1164, with the honest answer to the review's question (it is chiefly about the inconsistency between the three modules, not a bad-reading arrangement).

### What changed

- **Split-expenses: four triggers for one action, reduced to one primary.** `TAB_CAPS['split-expenses'].add` carried an action, so Budget's generic toolbar button *and* Budget's FAB both stayed enabled for the tab and delegated (`.click()`) into Split-expenses' own header button — on top of that page's own FAB. Budget's generic add is now off for this tab (same as Reports), the embedded header button is demoted to secondary, and the tab's FAB is the single primary.
- **No second page title.** Split-expenses re-rendered its own `<h1>` under Budget's page title; embedded rendering now emits it as an `<h2>` section heading (the `embedded` parameter was already passed, just never consumed).
- **Deleting a group looks destructive.** `#split-delete-group` was styled identically to edit/archive; now `btn--danger-outline`, matching the existing app-wide precedent.

### Findings re-verified as already fixed on main (no change)

Four items from the same review batch did not reproduce against current `main` and were left alone: the Meals week-board clipping (a `wireScrollFade` scroller already exists), the dashboard FAB overlapping the Rewards widget (`--fab-safe-zone`/`--shell-tail` already reserves clearance — re-confirmed in a live browser at true scroll-bottom), the Documents filter-chip clipping (both scrollers already wired), and Budget's `+ Entry` placement (already in the toolbar actions slot).

### Verification

Rebased onto current `main` (v2.65.3). `test:budget-ui` 71/71 (3 new regression tests), `test:split-expenses`, `test:budget-structure` 5/5, `test:meals` 48/48, `test:calendar` 25/25, `test:frontend-audit` 385/385, `test:changelog` 28/28 — all green. `public/pages/meals.js` and the period-nav section of `public/pages/budget.js` are byte-identical to `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Split Expenses no longer shows three ways to add an expense at once, or a second page title
+  under Budget's own heading.** Viewed as Budget's Split Expenses tab, the tab used to offer its own
+  header button and its own floating button for adding an expense, on top of Budget's own generic
+  toolbar button and FAB - both of the latter only ever repeated the tab's own button under the
+  hood. Budget's generic add action is now switched off for this tab, the same way it already is for
+  Reports; the tab's own floating button is the one primary action, and its header button steps back
+  to a secondary one. The tab's own `<h1>` - a second page title stacked under Budget's - is now a
+  section heading instead, matching how it already looked in a lighter type size. Deleting a group
+  now gets the same restrained red treatment used for a destructive action elsewhere in the app,
+  instead of looking identical to editing or archiving it.
+
 - **A housekeeper can check out again, and work a second session on the same day** (#1133, #1138).
   The one button that carries both directions was disabled while someone was checked in, and it is
   the only thing that triggers the check-out path - so that path was unreachable: a household could
--- a/public/pages/budget.js
+++ b/public/pages/budget.js
@@ -249,7 +249,16 @@ const TAB_CAPS = {
   'subscriptions':  { month: false, note: 'budget.periodNoteSubscriptions', add: 'subscriptions.add' },
   'loans':          { month: false, note: 'budget.periodNoteLoans',         add: 'budget.newLoan' },
   'reports':        { month: true,  range: true, add: null },
-  'split-expenses': { month: false, note: 'budget.periodNoteSplit',         add: 'splitExpenses.addExpense' },
+  // `add: null` wie Berichte: Split-Ausgaben bringt seine eigene Primaeraktion
+  // mit (Kopfknopf + FAB in split-expenses.js). Vorher stand hier derselbe
+  // Aktionsname wie im eingebetteten Kopf, und der generische Kopfknopf UND
+  // der generische FAB dieser Seite delegierten beide per Klick an
+  // #split-add-expense - macht mit dem eigenen Kopfknopf und dem eigenen FAB
+  // der Unterseite VIER Ausloeser fuer dieselbe Handlung (Cross-Modul-Review:
+  // "drei violette Add-Knoepfe zugleich"). Split-Ausgaben ist das einzige
+  // Sub-Tab mit eigenem Primaerknopf/-FAB; die anderen sechs teilen sich
+  // Budgets generische Knoepfe, weil sie keinen eigenen mitbringen.
+  'split-expenses': { month: false, note: 'budget.periodNoteSplit',         add: null },
 };
 
 // Sentinel für „keine eigene Farbe" im Kontofarb-Wähler: der echte Wert ist der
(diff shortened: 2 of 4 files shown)


ulsklyc (changes requested):
Two changes in one PR, and they have very different standing with me. The Split Expenses half I want; the "jump to now" half asks me to reverse a decision I made on purpose, and I would rather we talk about it than merge it quietly.

## The Split Expenses half: yes, please

`TAB_CAPS['split-expenses'].add = null` is right, and your reading of why is right. Four triggers for one action - the tab's own header button, the tab's own FAB, Budget's generic toolbar button and Budget's generic FAB, with the latter two only ever clicking the first under the hood - is exactly the kind of thing that reads as "the app is unsure what it wants". Reports already does it this way, so this is the tab joining an existing rule rather than inventing one.

The `<h1>` becoming a section heading is the same story: a second page title stacked under Budget's own was a structural bug, not a styling preference, and it already *looked* like a section heading. Good catch on the destructive-action colour for deleting a group, too.

## The "Today"/"Current" half: this reverses a decision, and the reason is still in the file

The comment you replaced said it:

> „Aktuell" ist ein Reset, kein Navigationsschritt: hinter dem Stepper statt zwischen Pfeil und Wert.

That was not drift - it was the decision, written down where the next person would find it. Your PR moves the button to the other side and cites Calendar as the reference, which turns "Calendar does it differently" into "Calendar is right". Both readings are defensible:

- **Yours:** one position app-wide beats three, and Calendar is the most-used module, so it sets the house style.
- **The existing one:** a reset is a different *kind* of control from a step, so putting it inside the stepper group reads as a third step. Position carries that difference.

I lean toward keeping the current arrangement, but I am genuinely open - what I am not willing to do is flip it without noticing that I am flipping it. So:

**Could you split this PR?** Land the Split Expenses part now - it is uncontroversial, it fixes a real redundancy, and it should not wait on a taste question. Open the "Today"/"Current" position as its own PR or, better, as an issue first, and make the argument there. If you make it well, I will take it; it is a small change either way, and then the decision is recorded in one place instead of being reversed in a comment diff.

One thing that would help me decide, if you want to make the case: is there a module where the *current* arrangement actually reads badly, or is this purely about the inconsistency between the three? The first would move me; the second is real but cheaper to fix in the other direction.

Nothing else in the diff needs work - the code is clean, `test:budget-ui` covers the change, and the comments you wrote explain themselves well.

ulsklyc (approved):
Thanks for the split, this is exactly the half that could land on its own. Approving.

**Merge conflict: resolved on your branch.** The branch had gone DIRTY after #1161, #1162 and #1163 landed on `main` overnight, all three with a `### Fixed` entry at the same spot as yours, so CI could not run. Since maintainer edits are enabled, I merged `origin/main` into `fix/cross-module-ux` and pushed it as `7d7cdbe3`. The only conflict was `CHANGELOG.md`: your entry first, then the two from `main`, nothing reworded. No code file was touched.

**What I checked on that merge:**

- `add: null` for `split-expenses` in `TAB_CAPS` plus the removed `addHandler` case: Budget's generic toolbar button and FAB are hidden on this tab the same way as on Reports, and nothing else reached `#split-add-expense` through the budget page.
- `embedded: true` is the only call that exists today (`budget.js:720`), so the `h1` to `h2` switch is the live path.
- The split title is styled by class only (`typography.css:74`, `:133`, `split-expenses.css:38`), so the tag change does not lose any styling.
- The FAB is still the primary action on desktop: `.page-fab--docked` in `glass.css` turns it into the labelled dock button there, so demoting the header button to `btn--secondary` does not leave a wide viewport without a primary action.
- `btn--danger-outline` exists (`layout.css:4891`) and `deleteGroup()` already goes through `confirmModal({ danger: true })`.
- Suites on the merge: `test:budget-ui` 71/0, `test:frontend-audit` 386/0, `test:changelog` 28/0, `test:split-expenses` green.

**Nice to have, not for this PR:**

- With the tab title now an `h2`, the group name (`split-expenses.js:368`, `<h2>${esc(group.name)}</h2>`) sits at the same level as the tab title rather than under it, and Balances / Recent expenses / Activity stay at `h3`. One level down for the group name (and `h4` for its sections) would make the outline read Budget > Split Expenses > group > section. Only when embedded, so it follows the same `embedded` switch.
- The three new tests match source text (`const TitleTag = embedded ? 'h2' : 'h1'`, the class strings). They hold the current spelling, but a renamed variable turns them red without a behaviour change, and a dead branch would keep them green. The budget-ui suite already renders through the real modules in places; a render with `embedded: true` that counts `h1` elements and primary buttons in the container would bind the behaviour instead. Fine as a follow-up.

The "Today"/"Current" question is now in #1164, I will answer there.

Issue #1003, opened 2026-09-02
Shopping: a price on the item, and the history that follows from it


ulsklyc:
Promised in #714 ("price per item, yes - it is on the list") and never given a place. Filing it so the promise has one.

## The gap

`shopping_items` carries 18 columns - `name`, `quantity`, `category`, `notes`, `url` plus sort/sync ones - and **none of them is a price or a shop** (re-measured against the live schema, not the migration text). So "what did I last pay for this, and where" cannot be answered from data the household already enters, even though every ingredient of the answer is already there.

## Why this one earns its place, and the catalogue next to it does not

A price is a fact about **a purchase**: you paid it once, at one shop, on one day, and it stays true forever. Nothing has to be tended for it to remain correct, and an old price is still a useful old price. That is the test a field has to pass. Nutrition, package sizes and a product catalogue fail it - those are facts about a *product*, they go stale when a manufacturer changes a recipe, and maintaining them by hand produces a half-filled table that looks like an answer. #714 has the full reasoning, including the correction that this is not a privacy argument.

## Shape

- A price on the shopping item, optional, entered when you tick the item off rather than when you add it - that is the moment the number is known.
- Which shop. The item has no store field either; a price without one answers half the question.
- The history follows from the items already ticked off, with no new concept: same name, earlier purchases, what it cost each time.

## Decided (answered by @LEON-BROTHER in the thread)

1. **Per unit, with a choice - but per item ships first.** Per-unit is what actually compares two shops, and it is what he would look up. It only computes if `quantity` is machine-readable, and it is free text today (`2 kg`, `2kg`, `zwei Kilo` are one column). So: **price per item first**, then **quantity as number + unit**, and per-unit follows from the second for free. Sequencing it the other way makes the price field wait for the harder half.
2. **A managed list of shops**, not free text. A household visits few enough shops that maintaining them is cheap, and free text is messy from the first week (`REWE`, `Rewe`, `rewe City`).
3. **Budget counting is an option, not automatic** - and the pattern exists: a subscription either materialises a budget entry (`budget_entry_id` on the record) or does not. Caveat to settle in the implementation: a shopping price is a plan until the shopping is done, so the option should post **when the item is ticked off**, not when the price is typed. Otherwise the budget fills with money nobody has spent.

## Still open

- **Matching for the history** - same list item text, or something more stable?

Origin: #714.

ulsklyc:
[...]

PR #1165, opened 2026-09-12
feat(shopping): duplicate a list, recency-aware suggestions, last-category default (#1103, #1113)

### LINKED ISSUE #1113: Shopping: a suggested item name brings back the category it was last filed under

mclgoerg:
This is the trimmed core of #1120, cut down to exactly what your review asked to land now. New branch and new PR rather than a force-push, because the trimmed change is a different thing from what #1120's title and body describe; #1120 is closed with a pointer here, and everything split out stays parked on `feat/shopping-list-templates` to be opened one at a time after this lands.

## What's in

- **`POST /shopping/:listId/duplicate`** (#1103). One transaction; category assignment and the manual per-category `sort_order` are always carried over — `sort_order` copied explicitly rather than left to the `AFTER INSERT` trigger, which would append every copied row to the end of its category. The three flags (`resetChecked`, `keepQuantities`, `keepNotes`, all defaulting to true) control the rest.

  What a copy never carries, independent of the flags — each of these names something true only of the *original* item:
  - the CalDAV sync columns (`external_uid`, `external_source`, `external_account_id`, `external_object_url`, `outbound_dirty`, `outbound_attempts`): they are a statement about one specific mirrored remote object. Copied verbatim, an edit to the copy would PUT onto the same remote VTODO the original mirrors, and deleting the copy would queue deletion of that shared remote object — the same class of hazard you spelled out on #998. A copy is a new, local item; if the destination list is itself a sync target it uploads as a brand-new VTODO on the next pass.
  - `added_from_meal`: the copy comes from this action, not from the meal that added the original.
  - `price_cents` and `store_id`: both are facts about a purchase actually made — "paid once, in this shop" (#1003). A copy has not been bought yet.

- **`GET /shopping/suggestions` carries category + quantity** of the most recently used row per name, ordered by most recently used (`MAX(created_at)`, tie-broken by id) instead of alphabetically. Picking a suggestion pre-fills category and quantity in quick-add. Closes #1113.

- **Default-category fix** (#548): `POST /:listId/items` without a category falls back to the *last* category, matching `import-pantry`'s existing fallback and the issue's original intent. One line, plus the test that pins it.

- **The live-testing fixes that touch the above**: the suggestion pick renders the name instead of `[object Object]` (the client half of the suggestions change, now with its own tests); quick-add's category selector resets to the default after every added item instead of keeping a previous suggestion's category; and the duplicate route never copies `store_id` (same reasoning as the price, folded into the provenance list above).

- Docs: `CHANGELOG.md` entries for the kept features only; `docs/SPEC.md` and the OpenAPI path for duplicate + suggestions only.

## What's out (split per your review)

The template flag is **dropped**, as you asked and we had offered — the "is this list a template" question stays yours to answer once, deliberately. The code is parked on `feat/shopping-list-templates`. The rest follows as separate PRs after this lands, in order: archiving, uncheck/check-all, checked-items price total, move-between-lists (#998), cross-category drag, and the same-name merge in its own PR that addresses the three defects you named (silent notes/url/category loss, ASCII-only `COLLATE NOCASE`, and the 201→200 contract change). In this PR, `POST /:listId/items` keeps plain insert semantics and returns 201, unchanged.

## No migration

This PR is schema-free: `server/db.js` is untouched (`git diff main -- server/db.js` is empty). `is_template` (194) and `archived_at` (195) left with their features, so no migration number needs holding for this — the 197 you provisionally offered can stay with #1095.

## Sequencing with #1109

Branched from current `main` (`ac78c8f7`), not stacked on #1109, per your instruction — rebase-ready whenever #1109 lands.

## Test plan

Run on this branch, all green:

- `npm run test:shopping-routes` (65) — duplicate options matrix, sort_order preservation, CalDAV/`added_from_meal`/`price_cents`/`store_id` never copied, 404/400 validation; suggestions recency order and category+quantity payload; default-category fix.
- `npm run test:shopping-ux` (48) — suggestion pick fills name/category/quantity, unknown category leaves the field alone, renderer never passes the suggestion object to `esc()`; quick-add category reset.
- `npm run test:migrations-append-only` (confirms no new migration), `test:openapi-structure`, `test:openapi-coverage`, `test:frontend-audit`, `test:i18n`, `test:i18n-translated`, `test:changelog`.

The duplicate dialog, the suggestion dropdown and the two quick-add fixes were also clicked through live on my test instance while this was still part of #1120; the code here is the same, minus the dropped features.

Closes #1113. Core of discussion #1103; the follow-ups above cover the rest of it.


ulsklyc (changes requested):
Thank you for cutting this down exactly along the lines of the #1120 review. The scope holds: no `is_template`, no archive, no same-name merge, `server/db.js` untouched, `POST /:listId/items` still answers 201. The duplicate route itself is solid, and I measured it rather than taking the description's word for it. Nothing below blocks the merge on correctness. What I would like before merging is mostly about the tests binding what the description promises, plus one default-category question.

All numbers below are from a throwaway merge of this branch with current `main` (`f254b57c`).

## Should fix

**1. "Last category" is not "Sonstiges" once a household adds a category** (`server/routes/shopping.js:933`)

`POST /categories` appends new categories at `MAX(sort_order) + 1` (`server/routes/shopping.js:141-147`), so the last category by aisle order becomes whatever the household added most recently. Measured: before adding a category, an item without `category` lands in `Sonstiges`; after `POST /categories { name: 'Baumarkt' }`, the next item lands in `Baumarkt`. The comment above the line says "LETZTE (#548, 'Sonstiges')", but the two only coincide on a fresh install.

The codebase now carries three definitions of "the default category": the item route takes the last by `sort_order` (this PR), the quick-add client takes the name `DEFAULT_CATEGORY_NAME = 'Sonstiges'` (`public/utils/shopping-categories.js:3`, reset at `public/pages/shopping.js:1327-1329`), and the category delete fallback takes the first by `sort_order` (`server/routes/shopping.js:213-216`, while its header comment says "zu Sonstiges"). My own #1120 review suggested "last, matching import-pantry", so this is partly on me. Suggestion: prefer the name when it exists and fall back to the last category only when it does not, for example `validNames.includes('Sonstiges') ? 'Sonstiges' : validNames.at(-1)`. That matches the client and #548's intent. `import-pantry` (`:1170`) has the same gap today; fixing it in the same line or leaving it for a follow-up are both fine with me, as long as the two stay in step.

**2. The recency tests pass without recency** (`test/test-shopping-routes.js:75`, `:86`)

The query itself is correct. It avoids the bare-column trap by reading category and quantity in a separate `ORDER BY created_at DESC, id DESC LIMIT 1` per name, and my own cases pass against it:
- same name, the row with the later `created_at` has the *smaller* id: category and quantity come from the later row;
- names inserted as `Ananas` then `Apfel`: result `['Apfel', 'Ananas']`;
- name A with the later `created_at` but the smaller max id: A comes first.

The PR's tests do not pin this down, though. Counterproofs, each run against a `cp` backup and restored afterwards:

| Mutation in `server/routes/shopping.js` | `test:shopping-routes` |
|---|---|
| name order `ORDER BY latest DESC, latest_id DESC` -> `ORDER BY name ASC` (the old alphabetical behaviour) | **green, 65/65** |
| latest row `ORDER BY created_at DESC, id DESC` -> `ORDER BY id DESC` (ignores `created_at`) | **green, 65/65** |
| latest row -> `ORDER BY created_at ASC, id ASC` | red, 1 failure (`:86`) |

The order test inserts `Zwiebeln` then `Zucchini`, but `Zucchini` is also first alphabetically, so the test cannot tell recency from the alphabet. Both suggestion tests also insert within the same second, so only the id tiebreak is ever exercised, never `created_at`. Please pick names whose alphabetical order is the opposite of their insertion order, and add one case that sets `created_at` explicitly so the later timestamp has the smaller id. With those two cases in my probe, both green mutations above turn red.

**3. The suggestions response shape changes on `/api/v1`, and the OpenAPI path does not say so** (`server/openapi/paths/shopping.js:28`)

`GET /shopping/suggestions` goes from `string[]` to `{ name, category, quantity }[]`. That is a contract change on the committed API surface. The only in-repo consumer is `public/pages/shopping.js`, which is updated, but token clients are not in the repo. `docs/SPEC.md` documents the new shape. The OpenAPI entry is unchanged (the description says the OpenAPI path was updated for suggestions too, but the diff touches only `/duplicate`), and the CHANGELOG lists it under "Added" without telling API users that the shape changed. Please add a `description` to the suggestions op and one sentence in the CHANGELOG entry saying that the response items are now objects.

## Nice to have

- **The quick-add reset test is a text guard** (`test/test-shopping-ux.js:1244`). Changing the condition at `public/pages/shopping.js:1327` to `if (false && ...)` keeps `test:shopping-ux` green (48/48), so the line can be dead and the guard still passes. `applyAutocompleteSuggestion` got a real behaviour test. The reset could get one too, or at least a check that the assignment is reachable.
- **Flags accept anything but a literal `false`** (`server/routes/shopping.js:758-760`). `resetChecked: 'false', keepQuantities: 'false'` duplicates with the checkbox reset and the quantity kept (measured: `is_checked 0`, `quantity '2'`). A 400 for non-booleans would make API misuse visible.
- **`POST /:listId/duplicate` without a body returns 500** (Express 5 leaves `req.body` undefined, and `:755` reads `req.body.name`). `POST /` and `PUT /:listId` share the pattern, so this is not new with this PR; `req.body?.name` would be enough here.
- **Tags are not copied.** `shopping_item_tags` rows on the source do not reach the copy (measured: `tags: []`). That is consistent with the rule, since tags are mirrored VTODO CATEGORIES, but it is not in the "never copied" list in the route comment, SPEC or CHANGELOG. One word there saves the next reader the question.
- **Suggestions run up to 8 extra queries.** Harmless with synchronous SQLite. A single `ROW_NUMBER() OVER (PARTITION BY name ORDER BY created_at DESC, id DESC)` query would do the same work.
- `test/test-shopping-ux.js:1255`: `git diff --check` reports a new blank line at EOF.
- `shopping.duplicateDefaultName` builds `"{{name}} – Kopie"`, while recipes build `"{title} (Kopie)"` from `recipes.copySuffix`. Not wrong, just two spellings for the same idea.

## Verified, no action needed

- **Sort order.** It is copied explicitly. The only sort trigger (`trg_shopping_items_sort_order`) fires `WHEN NEW.sort_order = 0`, so a copied non-zero rank is never overwritten, and the other `AFTER INSERT` triggers only touch the search index. I measured five items across two categories, hand-reordered, one checked, `resetChecked: false`: the copy matches the source exactly on `name:sort_order:is_checked`. Counterproof: passing `0` instead of `item.sort_order` turns `test:shopping-routes` red (`:157`).
- **Sync and purchase columns.** They are never copied. Counterproof: copying `external_uid`, `external_source`, `external_object_url`, `price_cents` and `store_id` turns `test:shopping-routes` red (`:204`).
- **Default category.** Counterproof: reverting to `validNames[0]` turns `test:shopping-routes` red (`:262`).
- **Access.** Shopping lists are household-wide with no per-list visibility, so neither duplicate nor suggestions can reach a list the caller may not see. Module rights and token scopes apply through the `/api/v1` gates (`POST` requires write).
- **CalDAV.** The new list cannot be a sync target at creation. If someone later selects it as a target, `pendingShoppingCreations()` picks up its `external_source = 'local'` rows, so the copies upload then instead of disappearing silently.
- **404 and 400.** Unknown list id and `/abc/duplicate` return 404, an empty name returns 400, and the insert runs in one transaction.
- **Client.** No `innerHTML`, all user data goes through `esc()` (the modal title is escaped inside `openModal`), all text goes through `t()`, and no new CSS values are introduced. Locales: all 24 files +10/-1, no reserialisation noise.

## Test runs (merge with main)

`shopping-routes` 65/65, `shopping-ux` 48/48, `migrations-append-only` 3/3, `openapi-structure` 11/11, `openapi-coverage` 3/3, `frontend-audit` 386/386, `i18n` 54/54, `i18n-translated` 92/92, `i18n-plural` 17/17, `changelog` 28/28, `shopping` 53/53, `shopping-send` 15/15, `shopping-order-migration` 6/6, `caldav` 56/56, `caldav-recurrence` 20/20, `caldav-outbound` 98/98, `caldav-todo-outbound` 62/62, `caldav-reminders` 39/39, `caldav-object-urls` 21/21, `caldav-event-target` 9/9, `search-index-duplicates` 9/9, `search-permissions` 11/11. All exit 0.

## Heads-up: overlap with #1109

Not a blocker here, but whichever PR lands second will have to rebase. A three-way merge of this branch with `pull/1109/head` (merge base `7a4d18f6`):
- Textual conflicts involving this PR:
  - `public/pages/shopping.js`: one small hunk in the `__test` export.
  - `test/test-shopping-ux.js`: both PRs append at the end of the file; keeping both sides resolves it.
  - `docs/SPEC.md`: both add a section at the same spot.
- The conflicts in `server/db.js`, `server/db-schema-test.js` and `package.json` are between #1109 and `main`, not this PR.
- `server/routes/shopping.js` merges without conflict, but there is one semantic point. #1109's version counter is trigger-based, so duplicated items bump the new list's version automatically. However, #1109's item routes also return `list_change`, which the client uses to acknowledge its own writes, and `/duplicate` does not. Whoever rebases second should decide whether `/duplicate` needs one.

ulsklyc (changes requested):
Thanks @mclgoerg - the round holds on `950d5b9b`. One small thing left, and it is one I should have named last round instead of filing it under spelling.

## Should fix

**The dash in `shopping.duplicateDefaultName`.** 23 of the 24 locales build the default name with a dash: `"{{name}} – Kopie"`, `"{{name}} – Copy"` and so on (en dash), and `ru`/`uk` with an em dash (`"{{name}} — копия"`). CONTRIBUTING ("Hyphens, not dashes") covers locale values explicitly: `-`, never `–` or `—`. No guard checks this outside the READMEs, so nothing went red. `ja` already writes `"{{name}} - コピー"`. Please make it `"{{name}} - Kopie"` in the other 23 files - one value per file, the 4-space indent stays as it is.

The `(Kopie)` vs `- Kopie` question from last round stays out of scope, as you said. Only the dash is a rule.

## Verified

All numbers from a throwaway merge of `950d5b9b` with current `main` (`fb0311ab`). The only conflict is `CHANGELOG.md`: `main` gained its own `[Unreleased]` in the meantime. Your two `### Added` bullets go after #1137's, your `### Fixed` bullet after #1135's. I will resolve that when merging if nothing else changes, so there is no need to rebase just for it.

- **Suites:** `shopping-routes` 70/0, `shopping-ux` 50/0, `shopping` 53/0, `openapi-structure` 11/0, `openapi-coverage` 3/0, `frontend-audit` 386/0, `i18n` 54/0, `i18n-translated` 92/0, `i18n-plural` 18/0, `changelog` 28/0, `migrations-append-only` 3/0, `shopping-send` 15/0, `search-index-duplicates` 9/0, `caldav-todo-outbound` 62/0.
- **Counter-proofs**, each against a `cp` backup and restored afterwards (70/0 and 50/0 again at the end):

| Mutation | Suite | Result |
|---|---|---|
| name order -> `ORDER BY name ASC` | `shopping-routes` | red, 69/1 |
| latest row -> `ORDER BY id DESC` | `shopping-routes` | red, 69/1 |
| latest row -> `ORDER BY created_at ASC, id ASC` | `shopping-routes` | red, 68/2 |
| item POST default -> plain `validNames.at(-1)` | `shopping-routes` | red, 68/2 |
| `resetQuickAddCategory` condition -> `if (false && ...)` | `shopping-ux` | red, 49/1 |
| the call in the submit handler removed | `shopping-ux` | red, 49/1 |
| flag check -> no type check | `shopping-routes` | red, 69/1 |

  The two tests that passed without recency last round now fail as they should.
- **OpenAPI and CHANGELOG** name the `string[]` -> `{ name, category, quantity }[]` change. Tags are listed as never copied in the route comment, SPEC and CHANGELOG.
- Leaving the 8 extra queries and the `Kopie` spelling for later is fine with me.

## Order with #1109

Pairwise, this branch and #1109 still conflict in `docs/SPEC.md`, `public/pages/shopping.js` (the `__test` export), `server/openapi/paths/shopping.js` and `test/test-shopping-ux.js`. It stays as agreed: whichever lands second rebases, and if #1109 is first, your rebase decides on `list_change` for `/duplicate`.

ulsklyc (approved):
Thanks @mclgoerg - all of it holds on `8edd14ef`. Approved.

## `list_change` for `/duplicate`: leave it out

Agreed, no acknowledgement needed. The copy is created by `INSERT`, which the #1109 trigger already counts, so other open pages pick it up on their next poll. The only cost is one extra reload on the duplicating client itself, and that is not worth more code in this PR.

## Verified

- **The dash** is now ` - ` in `shopping.duplicateDefaultName` in all 24 locales (parsed, not grepped).
- **The rebase onto #1109 only moved things.** The range-diff against my last review (`950d5b9b`) shows placement changes, and the content of both commits is unchanged:
  - CHANGELOG: both `Added` bullets after #1108, the `Fixed` bullet after #1135, all under `[Unreleased]`;
  - SPEC: "Duplicating a list" after the change-counter table;
  - OpenAPI: `/suggestions` and `/versions` as separate entries;
  - `__test` and the test blocks: both sides kept.
- **On a merge with current `main` (`8b298415`, which also has #1095; merges cleanly):**
  - `shopping` 53/0, `shopping-ux` 77/0, `shopping-routes` 72/0, `shopping-send` 15/0;
  - from #1109: `shopping-versions` 19/0 and `live-feed` 18/0;
  - `openapi-structure` 11/0, `openapi-coverage` 3/0, `i18n` 54/0, `i18n-translated` 92/0, `i18n-plural` 18/0, `changelog` 28/0;
  - `frontend-audit` 386/0, `suite-chain` 4/0, `migrations-append-only` 3/0, `search-index-duplicates` 9/0, `caldav-todo-outbound` 62/0, `test-db.js` exit 0.
  - CI is green on Node 22 and 24.
- **The earlier counter-proofs still bind after the rebase.** Re-run on that merge, each from a backup and restored to 72/0 and 77/0:

| Mutation | Result |
|---|---|
| suggestion order -> `ORDER BY name ASC` | `shopping-routes` red, 71/1 |
| item POST default -> plain `validNames.at(-1)` | `shopping-routes` red, 70/2 |
| `resetQuickAddCategory` condition -> `if (false && ...)` | `shopping-ux` red, 76/1 |

I will merge it from here.

PR #1169, opened 2026-09-13
refactor(notes): share category name length limit

PechJiri:
## Summary

I followed up on the reviewer's nice-to-have from #1044 and made the Notes category name limit a single shared contract.

I exported `NOTE_CATEGORY_NAME_MAX_LENGTH` from the existing isomorphic `public/utils/note-category-name.js` utility and used it for:

- server-side Notes category validation;
- the validation error text;
- `NoteCategory`, `NoteCategoryInput`, and `NoteCategoryRenameInput` in OpenAPI;
- the Notes category picker `maxlength`;
- the Notes category manager `addMaxLength` configuration.

I left the generic category manager default at 60 characters, so other modules retain their existing behavior.

This is an internal refactor only. It does not change behavior, the API contract, the database, design, localization, dependencies, or the build. It does not require a changelog entry.

## TDD evidence

I wrote the contract and call-site tests before changing production code.

RED on the pre-refactor implementation:

- `test/test-note-categories.js`: 10 passed, 2 failed because the shared export and server wiring did not exist. The existing 80/81 behavior test already passed.
- `test/test-note-category-filter.js`: 23 passed, 1 failed because OpenAPI was not tied to the shared contract.
- `test/test-category-manager.js`: 17 passed, 1 failed because both Notes UI consumers were still numeric literals.
- `test/test-category-manager-lifecycle.js`: 15 passed, confirming the generic 60-character default and configurable override before production changes.

GREEN after implementation:

- `npm run test:notes-routes`: 60 passed.
- `npm run test:category-manager`: 18 static checks passed and 15 lifecycle tests passed.
- `npm run test:openapi-structure`: 11 passed.
- `npm run test:openapi-coverage`: 3 passed.
- `npm run test:frontend-audit`: 386 passed.
- `npm run test:layer-boundary`: 3 passed.
- Syntax checks for all seven changed JavaScript files passed.
- `git diff --check origin/main...HEAD` passed.
- Fresh full `npm test` passed.

The full baseline `npm test` also passed on `da9ea6ecc10e642a36c29d05075aaabb43f907b7` before the change.

## Review

I ran two independent Superpowers code review rounds over the complete candidate diff. Both reported no Critical, Important, or Minor findings and returned `REVIEW_PASS`. No review-driven changes were required.

Drafted with OpenAI Codex.


ulsklyc (approved):
Thanks @PechJiri - this is exactly the follow-up from #1044, and it is clean.

Checked on `f4def1c3` in a separate worktree:

- `test:notes-routes` 60/0, `test:category-manager` 18 + 15/0, `test:openapi-structure` 11/0, `test:openapi-coverage` 3/0, `test:layer-boundary` 3/0.
- Counterproof: putting the literal `80` back into `validateCategoryName()` turns `test-note-categories.js` red (1 failing), so the wiring tests bind.
- The server importing from `public/utils/note-category-name.js` follows the existing pattern (`categoryNameKey` already comes from there), and the generic manager default of 60 stays as it is for the other modules.
- No CHANGELOG entry is right: nothing a user or an API client can observe changes.

Nothing blocking. One small note, take it or leave it: the three source-text assertions (the regexes over `notes.js`, `note-categories.js` and `schemas.js`) check how the code is spelled, so a harmless reformat - say `maxlength` moving in front of `id` on the picker input - would turn them red with nothing actually wrong. The value assertions (80/81 on the server, `maxLength` on the three schemas) are what carry the contract. Fine to keep them as they are.

Issue #1173, opened 2026-09-13
Health: add a permission-gated fasting tracker


PechJiri:
## Problem
Yuvomi Health currently has no fasting journal. Users (me 🙂) may need a private, self-hosted alternative to paid trackers for starting and ending fasts, correcting history, and seeing restrained progress information.

## Proposed scope
- Add a permission-gated `Fasting` Health tab at `/health/fasting` and a self-only dashboard widget.
- Follow existing Health module access, role/member capability overrides, visibility defaults, caregiver grants, API-token scopes, and optimistic revision conventions.
- Track UTC start/end instants with a captured IANA time zone, editable date/time, goal presets 12:12, 14:10, 15:9, 16:8, 18:6, 20:4, and user-defined whole-hour goals up to 14 days.
- Provide active/inactive timer, multi-day dial, history, backfill/edit/delete with undo, CSV export, current/longest streak, all-time/year/30-day count/total/average, and a seven-day actual-versus-captured-goal view.
- Add optional per-user goal-reached and planned-next-start reminders. A next-start reminder is scheduled at the end plus the remaining eating window for goals under 24 hours; fasts of 24 hours or more have no next-start reminder. Notifications are owner-only and off by default.
- Require an explicit first-use acknowledgement that this is a tracking tool, not medical advice; include neutral safety copy and no diagnosis, eligibility scoring, or metabolic certainty claims.
- Keep offline display/resume possible, but do not implement offline writes or calendar completion badges.

## Delivery proposal
1. `feat(health): add fasting journal` - capability, schema, safety, privacy/caregiver rules, timer, history, edit/delete, CSV.
2. `feat(health): add fasting insights` - streaks, summaries, and chart.
3. `feat(health): add fasting reminders` - settings, reconciliation, and notification origins.
4. `feat(dashboard): add the fasting widget` - self-only controls and picker permission.

Each PR will include focused tests, all locale keys, OpenAPI/docs/changelog updates, and the full test suite. This issue is intentionally product/architecture scope only; deployment and private acceptance details are not part of the upstream issue. Drafted with Codex assistance; final merge remains maintainer-controlled.

PechJiri:
## Implementation update: final pre-submission review

I have an integrated implementation, but the final requirements review found remaining gaps. I am holding the upstream PR submission until the first slice is independently complete rather than presenting the whole feature as ready.

The main remaining product work is:

[...]

PR #1177, opened 2026-09-13
feat(health): add fasting journal

PechJiri:
## Summary

Add a first-class fasting journal to Health for people who want to track their
fasts in Yuvomi without a separate paid service. This is the journal slice of
the four-part proposal in #1173.

Refs #1173. Insights, reminders and the dashboard widget follow in separate PRs;
this PR does not close the full feature issue.

## Stack and merge order

1. **[This PR — journal](https://github.com/ulsklyc/yuvomi/pull/1177)** (`feat/fasting-journal`)
2. **[Insights — #1178](https://github.com/ulsklyc/yuvomi/pull/1178)** (`feat/fasting-insights`)
3. **[Reminders — #1179](https://github.com/ulsklyc/yuvomi/pull/1179)** (`feat/fasting-reminders`)
4. **[Dashboard widget — #1180](https://github.com/ulsklyc/yuvomi/pull/1180)** (`feat/fasting-dashboard`)

Merge in this order. The later branches are commit-stacked so reviewers can inspect
the complete proposal now. Each successor will be refreshed onto current `main`
after its predecessor lands; do not merge a later PR first.

## Changes

- Permission-gated Health tab using existing role/member capability resolution,
  Health read/write access and private/family visibility defaults. Fasting ships
  enabled for every member; an admin can explicitly deny it for a role or person.
- UTC start/end records with captured IANA zones, whole-hour goals up to 14 days,
  personal first-use acknowledgement, interval validation and optimistic revisions.
- Active/inactive timer, selectable elapsed/remaining display and bounded
  multi-day dial. Actual recorded duration is not capped. Explanatory help is
  available from keyboard- and touch-accessible info controls.
- Earlier starts, manual history, editing, end/delete undo, cursor pagination,
  recorded-zone date filters and visibility-scoped CSV export.
- Family reading. MVP controls modify only the signed-in person's records;
  existing directed caregiver authorization remains supported at the API, but a
  caregiver cannot acknowledge safety for another person.
- Plain ES modules and existing design components/tokens, locale coverage,
  append-only schema, OpenAPI and contributor documentation. No new dependencies.

## Maintainer review follow-up

- Capability defaults are per capability: fasting defaults to `allow`, notes stays
  `none`, and explicit role/member `none` wins. Admin summaries and invite presets
  show the effective fasting state.
- Safety acknowledgement is owner-only. Unknown/malformed subjects, caregiver
  grants and private foreign record IDs now keep their distinct 400/401/403/404
  contracts, with no admin bypass of a missing caregiver grant.
- Family-reader state and export tests cover private running and finished records;
  overlap tests cover edits and backfills against completed intervals.
- Immediate Start/Finish uses server time. Implausible years are rejected and the
  unneeded expanded-year browser path was removed.
- Duration units and extra-day plurals use i18n in all 24 locales; unused journal
  keys were removed. Unexpected route failures are logged before a 500 response.
- Generic fasting visibility mutations now enforce the same capability, including
  explicit denials inherited from a role or set for a person.
- The schema is append-only migration **209** throughout the implementation,
  schema mirror, tests and documentation.

## Verification

- `npm test`: passed on final candidate `66ae6b2d`, based on current `main` at
  `b7ec5322`.
- `npm run test:fasting-browser`: 11/11 passed, no skipped cases, on the same
  candidate. It covers desktop/mobile, Czech/German/Arabic, keyboard-accessible
  help, lifecycle, date/time editing, permission boundaries, offline display and
  undo flows.
- Focused fasting, permissions, OpenAPI, locale/plural, timezone, migration/schema
  mirror and frontend-audit suites passed.
- Two independent Superpowers code reviews were completed and all validated
  findings were corrected. A separate final traceability audit checked every
  blocking and should-fix item from the maintainer review before publication.

The browser harness disables service workers, so this is not an installed-PWA
end-to-end claim. The full `test:document-guards` release handrail was not run for
this PR-only submission and remains required before an interface release.

API test setup follows the adjacent Health suites: imported routers, an isolated
in-process Express loopback HTTP fixture and in-memory SQLite. This is a narrow
exception to CONTRIBUTING's literal no-HTTP test guidance; it tests real decoding
and routing without an external service, separate server process or production data.

## Checklist

- [x] `npm test` passes on this standalone branch
- [x] Focused journal browser acceptance passes on this standalone branch
- [x] `CONTRIBUTING.md` and `DESIGN.md` followed, with the test-fixture exception disclosed above
- [x] No new frontend framework, build tool or dependency
- [x] New visible strings included in every locale
- [x] OpenAPI, feature/test documentation and changelog updated

Implementation and reviews are Codex-assisted. Final product acceptance and
merge remain with the human maintainer.


ulsklyc (changes requested):
Thanks for splitting this into slices and for the thorough PR body. This review covers the journal (#1177) only. #1178-#1180 wait until this one has landed and been refreshed onto main, as you planned.

## Decisions on the open product questions

1. **Fasting is available to everyone by default.** Right now only admins get it, and not because of anything fasting-specific: `resolvePermissions()` gives every capability the one global `CAPABILITY_DEFAULT = 'none'`, and the only other capability, `notes_manage_household_categories`, is a management right where "off" is the right start. Please give the fasting entry its own default of `allow` (for example a `default` field on the `PERMISSION_CAPABILITIES` entry that `resolvePermissions()` prefers over the global constant), so every member can use the tab unless an admin switches it off for a person or a role. The admin permissions sheet and the invite presets should show that state. Tests: a member without an override gets `allow`, an explicit `none` still denies, and the notes capability keeps its `none` default.
2. **The safety acknowledgement is personal.** Only the person who fasts can acknowledge it. A caregiver creating a fast for someone who has not acknowledged should be refused instead of acknowledging on their behalf (`acknowledgeSafetyFor(database, actorId, subjectId)` in `server/services/fasting.js`).
3. **Fasting in the caregiver model while Cycle stays out of it** is fine as it is.

## Migration number

Main is at **208** now (#1099 landed). Please take **209**. The renumber touches more than the `version:` line:
- `server/db-schema-test.js`: the `197:` key. Main has its own `197:` there, read by `test/test-search.js`. It is an object literal, so if a conflict fix keeps both, one silently replaces the other.
- `test/test-health-fasting-migration.js` and `test/test-health-fasting-service.js` look the migration up by `version === 197`, so they would pick up main's waste migration.
- The `v197` test titles, `docs/SPEC.md` (three mentions) and the PR body.

## Blocking

- `test/test-fasting-journal-browser.js:32`, `:38`, `:407` assign `innerHTML`. That is blocked everywhere in the repo, tests included. `replaceChildren()` plus `insertAdjacentHTML('beforeend', ...)` does the same.

## Should fix

Each of these I checked at the code, not only in the tests.

- **The privacy filters are right, but nothing would notice their removal.** I replaced `visibleFilter(...)` with a plain `user_id = ?` in the export query (`getAllFastingHistory`) and, separately, in the `active:` query of the state. `test:health-fasting` stays at 11/0 and 23/0 both times. So a family reader's CSV or state could include private fasts without a red test. Please add tests: a family reader's export and state leave out a private finished fast and a private running fast, and an edit or backfill that overlaps a finished interval returns `FASTING_OVERLAP`.
- **The "admin cannot bypass" assertion passes for the wrong reason** (`test/test-health-fasting-service.js:88-90`). Line 88 removes subject 1's capability, so the throw on line 90 comes from the missing capability, not from the missing grant.
- **Someone else's private record can be detected by id.** `loadWritable()` answers 404 only when the row does not exist and 403 otherwise. The existing Health routes answer 404 in both cases.
- **An unknown `user_id` signs the user out.** `capabilityAllowed(subjectId)` goes through `userRow()`, which fails with `401 FASTING_ACTOR_REQUIRED`, and `public/api.js` treats every 401 as an expired session. The member list on the page is loaded once, so picking a person who was removed meanwhile is enough. Please validate the subject (400) and answer 404 for an unknown one; 401 belongs to the actor only.
- **A phone clock slightly ahead breaks Start and Finish.** `public/components/fasting-controls.js:126` and `:157` send `new Date()`, and `validateRange()` (`fasting.js:149-152`) rejects anything later than server time with no tolerance. Letting the server fill in "now" when the client means "now" avoids that.
- **Dates:** `endAt <= startAt` compares ISO strings, and there is no lower bound for `start_at`, so a typo year like `0202` saves as an 1824-year fast that the insights slice would then average. Please reject implausible years on the server. If the expanded-year handling added to the shared `public/utils/timezone.js` only exists for such inputs, it can go with that.
- **i18n:** `public/utils/health-fasting.js:101-103` hard-codes `d`, `h` and `min`. `health.fasting.extraDays` takes `{{days}}` and renders "+1 days"; please pass `count` and add `extraDays_one`. I found no reference to `health.fasting.safetyHint`, `finishConfirm`, `notePrompt`, `deleteConfirm` or `deleteDetail`; if they are for later slices, they belong there.
- **Silent 500s:** `sendError()` in `server/routes/health/fasting.js` answers 500 without logging. The other Health routers log first.

## Nice to have

- `public/index.html` loads `fasting-controls.css` on every page. In this slice only Health uses it, and the router already loads `health.css` there.
- In `docs/SPEC.md` and `docs/test-suites.md` the new sections sit above the files' own intros.

## What is fine

Module gate and token scopes cover every new route, including the export. Read and write rules match the existing Health model. The client never derives date keys from `toISOString()`. `APP_RELEASE` is untouched. Locales are complete in all 24 files with real translations and no em-dash. The CHANGELOG entry sits under `[Unreleased]`.

**Migration order as of today:** #1177 = 209, #1181 (cycle tracker v2) = 210-213. Whichever lands first takes the next free number.

ulsklyc (changes requested):
Thanks, this round addresses everything from the last review, and the fixes hold: I ran the mutations again, this time on `66ae6b2d`.

## Counter-proofs on 66ae6b2d

| Mutation | Result |
|---|---|
| history/export query: `visibleFilter(...)` -> `user_id = ?` | red (`family state and export omit private active and finished records`) |
| `active:` query of the state: same | red |
| private-id 404 block in `loadWritable()` removed | red (4 tests) |
| `resolvePermissions()` ignores the per-capability default | red (`test:permissions`) |
| sparse role rows compared against the global default | red (`test:permissions`) |
| admin shortcut in `ensureSubject()` right after the self check | **green**, see Should fix |
| permissions sheet: `item.default ?? 'none'` -> `'none'` | green, see Nice to have |

Also confirmed at the code: the `allow` default for fasting while notes keeps `none`, the personal acknowledgement (`FASTING_ACK_FORBIDDEN`), migration 209 appended last, 404 for foreign private ids, 400/404 for bad subjects with 401 left to the actor, server-filled start and finish, numeric date checks with the 1900 bound, `count` plurals, logging before 500, and no `innerHTML` assignment left. `66ae6b2d` does not loosen the import guard: the call-site assertion still requires `parsePermissionGroup(opt.dataset.group)`.

## Should fix

- **The "admin cannot bypass" test still binds another branch.** `test/test-health-fasting-service.js:92-94` expects `FASTING_NOT_FOUND` on a `private` row. That now comes from the private-id check in `loadWritable()` (`server/services/fasting.js:301-304`), before `ensureSubject()` is reached, so an admin shortcut in `ensureSubject()` keeps the suite green (measured above). Please keep the assertion under its real name (private ids stay hidden) and add one that reaches the grant check, for example an admin without a grant creating a fast for subject 1, or updating a `family` row, expecting `FASTING_SUBJECT_FORBIDDEN`.
- **SPEC still describes the old capability model.** `docs/SPEC.md:3697-3701` says `PERMISSION_CAPABILITIES` registers `notes_manage_household_categories` and that capability rows default to `none`. With this PR there are two entries with their own defaults (`server/permissions.js:117`, `:123`, `:232`), and role rows equal to that default are not stored (`:474`). `permissionCatalog()` still reports `defaults.capability: 'none'` (`:407`); the catalog text in `server/openapi/paths/permissions.js:35` should say that the per-item `default` wins.
- **"Add completed fast" still takes its end from the phone clock.** `public/components/fasting-controls.js:142-144` seeds `end_at` with `new Date()`. If only the start is changed, that value reaches `validateRange()`, and a server clock a minute behind the phone answers `FASTING_FUTURE`. Start and Finish no longer have this problem. Either let the server treat an `end_at` a few minutes ahead of its clock as now, or seed the default from a server timestamp in `/fasting/state`.

## Nice to have

- `public/locales/cs.json`: Czech needs `extraDays_few` (`"+{{count}} dny"`); the file already uses `_few` elsewhere.
- The CHANGELOG entry does not say that fasting is on for every member by default and can be switched off per role or person. For an admin reading the release notes, that is the one permission fact in it.
- Every fasting test fixture inserts an explicit `allow` row, so no route test runs the real default. One API test with a member without any row would pin it end to end.
- The sheet's default display is only guarded by spelling: `'none'` instead of `item.default ?? 'none'` in `public/settings/pages/admin-permissions.js:129` and `:253` stays green. It cannot write a wrong row, but the sheet would show fasting as blocked.
- Still open from the last round: `fasting-controls.css` loaded on every page (`public/index.html:71`), and the new SPEC and test-suites sections above the files' own intros.

#1178-#1180 follow once this one has landed, as planned. Migration order unchanged: #1177 = 209, #1181 = 210-213.

ulsklyc (changes requested):
Thanks, this round holds up. I re-ran the mutations on `5b17bcc4`:

| Mutation | Result |
|---|---|
| admin shortcut in `ensureSubject()` after the self check | red (`admin cannot bypass the caregiver grant for a reachable subject write`) |
| `server_now` removed from the state | red |
| completed-entry default back to `new Date()` | red (browser scenario) |
| fasting default `allow` -> `none` | red (`fasting API is usable through the shipped default without an override row`) |
| rendered capability row ignores `item.default` | red (`test-permission-sheet-ui.js`) |
| shared `effectiveCapabilityAccess` falls back to `none` | red (both suites) |

SPEC, OpenAPI, CHANGELOG, the Czech `_few` form, the lazy stylesheet and the section order are all fixed. `test:fasting-browser` passes 12/12 locally, and the branch merges cleanly into current main.

## Should fix

- **The submit check still reads the phone clock, now in the other direction.** `public/components/fasting-controls.js:263` rejects `endAt > Date.now()`. The default end now comes from `server_now`, so a phone whose clock is behind the server can't save "Add completed fast" with its default values.
  - I reproduced it with the page clock 3 minutes behind for the whole interaction: the modal stays open with `rangeError` and nothing is saved.
  - With the old `new Date()` default, the same scenario saves.
  - The new browser test restores `Date` before `requestSubmit()`, so it never reaches this check.
  - Possible fix: when the creator opens, take the offset `Date.parse(state.server_now) - Date.now()` and check against `Date.now() + offset`. Alternatively, drop the client-side future check and rely on the server's `FASTING_FUTURE`.
  - Please keep the mocked clock through submit in the test and add a phone-behind case.

## Nice to have

- `effectiveCapabilityAccess()` in `public/settings/pages/admin-permissions.js:130`, which feeds the deviation marker at `:314`, is still untested: forcing it to `'none'` stays green.
- `docs/test-suites.md:193` could mention that `test:permissions` now also runs the rendered sheet test.

Migration order is unchanged: #1177 = 209, #1181 = 210-213.

PechJiri:
Addressed review 5208425160 in `2e87d5d8`.

- Completed-entry validation now uses an advancing server-derived clock, so both ahead and behind device clocks are handled while the mocked clock remains active through submit.
- Added browser coverage for +10 min and -3 min skew.
- Finished-review/history-style editors without trusted `server_now` no longer use the device clock as authority for future timestamps; server validation remains authoritative while local chronology checks remain.
- Added a unit guard proving the derived server clock advances with elapsed device time.
- Added behavioral coverage for capability deviation markers (sparse default vs explicit denial).
- Updated `docs/test-suites.md` to document the rendered permission-sheet coverage.

Two review rounds were completed. The second round found the finished-review caller and a missing advancing-clock mutation guard; both were fixed, then the delta review passed with no Critical, Important, or Minor findings.

Final pre-push audit re-read this review and mapped every requested item to implementation and tests.

Focused verification:
- `npm run test:fasting-browser`: 13/13
- `npm run test:health-fasting`: pass
- `npm run test:permissions`: 31/31
- `npm run test:frontend-audit`: 387/387
- `npm run test:sw-precache`: 7/7
- `git diff --check`: pass

The standalone branch's full `npm test` also reaches an unrelated date-sensitive tasks assertion that turned stale after 2026-09-14; current `main` already fixes that exact assertion, and this branch merges cleanly with current `main`. GitHub's merge-ref CI is the integration check for that combination.

ulsklyc (approved):
Thanks, the clock skew fix holds up. I re-ran the mutations on `2e87d5d8`:

| Mutation | Result |
|---|---|
| submit check back on the device clock (`const now = Date.now()`) | red (both new browser tests: creator with the phone 3 min behind, and the finished review) |
| only editors without `server_now` fall back to the device clock | red (finished review only) |
| derived server clock frozen at capture | red (`derived server clock advances by device elapsed time after capture`) |
| derived server clock returns absolute device time | red (same unit test) |
| `effectiveCapabilityAccess` forced to `'none'` | red (`sparse role capability defaults do not render a deviation marker`) |
| deviation marker never rendered | red (same test) |

The mocked clock now stays active through submit, the phone-behind case is covered, and catching the finished-review path as a second instance of the same bug was a good find. `docs/test-suites.md` is updated too. Locally `test:fasting-browser` passes 13/13, `test:health-fasting`, `test:permissions`, `test:frontend-audit` and `test:sw-precache` are green, and migration 209 is still the next free number after main.

## Before merge

- **`package.json` now conflicts with main.** Since your last check, main added six suites to the `"test"` chain (`test:parallel-runner`, `test:bind-address`, `test:household-members`, `test:household-member-guard`, `test:people-pickers`, `test:document-owner`). It's a one-line resolution: keep main's chain and append `test:health-fasting`. Everything else merges cleanly.
- **No CI run exists for `2e87d5d8`.** The push did not create one, neither waiting nor finished. The push that resolves the conflict will, and that run is the integration check against current main.

## Nice to have

- **Future timestamps on existing entries now get a generic message.** Editors without `server_now` (active start edit, history edit, finished review) no longer check "not in the future" on the client, which is right. But the server's 400 comes back as `reason: 'FASTING_FUTURE'`, and `fastingError()` in `public/components/fasting-controls.js:12` only knows offline and 409, so the user sees "Fasting could not be saved." instead of `health.fasting.rangeError`. Mapping `error.data?.reason` of `FASTING_FUTURE` and `FASTING_RANGE_INVALID` to `rangeError` would restore the specific message.
- `fastingServerClock` is captured in `openFastingEditor` (`fasting-controls.js:183`), after the visibility-defaults request and the safety confirmation. The derived clock therefore lags the server by however long that dialog stays open. The defaults are unaffected, but a hand-entered end time close to "now" can be rejected on the client although the server would accept it. Capturing the clock right after the state request and passing it along would match the JSDoc ("after a state response is received").
- Dropping `server_now` from the creator row, which disables the client-side future check, still passes 13/13. A small case that enters a future end in the creator and expects `rangeError` would pin the check itself, not just its absence of false rejections.
- `capabilityDeviationHtml` is tested with an injected `view`. The `state` fallback that `deviationChips()` actually uses (`admin-permissions.js:316`) stays untested: passing `{ draft: {} }` there is still green. Optional.

PechJiri:
Resolved the required pre-merge items from review 5214671585: package.json now keeps the current main test chain additions and preserves the fasting health, browser, permissions, and document-guard suites. Pushed reviewed merge commit e73a88c5. CI run 35015516455 is green on Node.js 22.x and 24.x; PR is CLEAN/MERGEABLE. The explicitly deferred nice-to-have items remain out of scope.

ulsklyc:
Merged as 546eb3ee, thanks for the careful rounds. Before merging I ran the full chain on this branch merged with current main, including the admin predicate guard that landed today: green. Next are #1178, #1179 and #1180 in that order; please bring them up to date with main when you get to them.

PR #1178, opened 2026-09-13
feat(health): add fasting insights

PechJiri:
## Summary

Add completed-record fasting insights to the Health journal introduced by #1177. Refs #1173.

## Review and merge order

1. [#1177 — fasting journal](https://github.com/ulsklyc/yuvomi/pull/1177) was merged as `546eb3ee`.
2. **[#1178 — fasting insights](https://github.com/ulsklyc/yuvomi/pull/1178)** targets `main`. It is one insights commit plus a review-fix commit, based on current `main` (`fabab86b`). The diff has no migration or schema changes.
3. [#1179 — fasting reminders](https://github.com/ulsklyc/yuvomi/pull/1179) should be rebased onto this branch and use it as its base. Remove the already-merged journal migration; assign the reminders migration number when rebasing against then-current `main`.
4. [#1180 — dashboard widget](https://github.com/ulsklyc/yuvomi/pull/1180) should be rebased onto #1179 and use it as its base; the widget itself needs no migration.

Review and merge in that order.

## Changes

- Count, total duration and average duration for all time, this year and the rolling last 30 days.
- Current and longest goal-qualified streaks, including bounded interval handling for very long fasts.
- Seven-day actual-versus-captured-goal chart, including explicit missing-record and partial-goal-coverage states.
- Statistics, history and CSV date filters consistently use the household display time zone; `start_tzid` remains record provenance.
- OpenAPI/SPEC contracts, 24-locale insight strings, and API coverage for date-line boundaries, cursor pages, the `/fasting` alias, CSV, and visibility.

## Verification

- `npm run test:health-fasting` — passed.
- `npm run test:health-fasting-stats` — passed.
- Focused browser chart test for partial goal coverage — 1/1 passed.
- Date-line API regressions for household-zone history/CSV — passed.
- `test:openapi-structure`, `test:changelog`, `test:i18n-translated`, and `test:household-timezone` — passed.
- Final independent Superpowers review against the reviewer checklist — no actionable findings.
- `git diff --check` — clean.

The browser harness disables service workers, so this is not an installed-PWA end-to-end claim. The full `test:document-guards` release handrail remains a separate pre-release requirement.

Implementation and reviews are Codex-assisted; final merge is maintainer-controlled.


ulsklyc:
#1177 is merged as `546eb3ee`, so this stack now carries code that is already on main. That is why all three are conflicting: each PR contains its predecessor plus the whole fasting journal.

Please bring them onto main one at a time, so each PR shows only its own step:

1. **#1178 (insights)** onto `main`. The journal part drops out. Its only migration, `version: 197`, is the journal migration that landed as `209` on main, so after the rebase #1178 should carry no migration of its own.
2. **#1179 (reminders)** onto the rebased `feat/fasting-insights`, and set its base branch to it. Of its two migrations, `197` is again the journal one; the reminders migration currently numbered `198` is the one that stays.
3. **#1180 (dashboard widget)** onto the rebased `feat/fasting-reminders`, base set to it. Both migrations it shows come from the two PRs below it; the widget needs none.

On the number: `MIGRATIONS` only ever grows at the end, so it is assigned at merge time. main holds 209 and #1181 holds 210 to 213, so the reminders migration takes **214** unless something lands before it. Please renumber when you rebase, not now.

I will review in the same order, starting with #1178 once it is on main.

PechJiri:
Done. #1178 is now a single insights-only commit directly on current `main` (`08887fb6`) and has no migration/schema changes. I also removed the unused journal strings from the locale diff, clarified the OpenAPI summary shape, aligned the aggregate date semantics to the household display zone (history/CSV filters remain recorded-zone), and added date-line plus family-visibility API regression tests.

Verification is green: full `npm test`, fasting browser suite 16/16, stats/OpenAPI/i18n checks, and two independent read-only review rounds. The PR description now records the requested #1177 → #1178 → #1179 → #1180 order and expected bases.

ulsklyc (changes requested):
Round 1 on `cd322b21`. Thank you for restacking this properly: 43 files instead of 73, one commit, and I verified the migration-free claim in the diff rather than taking it on trust. Visibility is sound too, which matters most here: `getAllFastingHistory` applies the same `visibility = 'family'` clause as the history route, so the insights never widen what another member can see. All 24 locales carry the 14 new keys with placeholders intact, and no orphaned journal strings are left.

Four of your six claims hold up exactly as written. Two need a word each, and one needs a decision from me, which I have made below.

## Blocking

**The CHANGELOG bullet will land inside a released section.** This is my doing, not yours: you branched off `08887fb6`, and `fabab86b` (`chore: release v2.67.0`) landed afterwards, moving the whole `[Unreleased]` content down into the new release block. On your branch the bullet sits correctly; in the merge result it ends up at line 97, **inside `## [2.67.0] - 2026-09-16`**, which is already published.

Nothing signals this: git merges the hunk without a conflict, GitHub reports `CLEAN`, and `test:changelog` checks structure, not placement. Please merge current `main` and move the bullet back under `## [Unreleased]`. After any rebase over a release, this check is worth a look:

```bash
```

## Should fix

1. **`test/test-health-fasting-stats.js:16` passes an option nothing reads.** The call uses `targetZone`, but `fastingStreaks` destructures `{ today, timeZone }` (`server/services/fasting-stats.js:48`). I counter-proved it three ways on the merged tree: as written, with the option deleted, and with the option set to garbage, all three return `{current:2,longest:2}` - while the real name, `timeZone: 'America/Los_Angeles'`, returns `{current:3,longest:3}`. The test currently passes through the `row.start_tzid` fallback, so it is green without exercising what it claims to. It is also the test your PR text cites as proof of the zone alignment. One word.

2. **The zone seam, and my decision on it.** Aggregates key on the household display zone (`server/services/fasting.js:259`) while history and CSV filter on the recorded zone (`server/services/fasting-dates.js:39`). Your own API test seeds exactly the case where they disagree: a fast completed `2026-12-31T16:00Z` with `start_tzid: Pacific/Kiritimati` in a Los Angeles household is pinned to the chart bucket `2026-12-31`, but a history filter for that same day will not return it, because its recorded-zone key is `2027-01-01`. One screen, two answers.

   **Decision: both follow the display zone.** A chart and the list under it must not disagree, however well the difference is documented. That matches `todayKey()` and the rest of the app. Whether you do the history/CSV side here or in a follow-up is your call - it only has to be settled before #1179 and #1180 land on top.

3. **`server/services/fasting-stats.js:19` re-implements `shiftDateKey()`** (`server/utils/timezone.js:158`), which this PR already imports three lines away in `server/services/fasting.js:256`. The arithmetic is correct and the guard allows this spelling, but it is one rule written twice.

4. **`public/pages/health-fasting-insights.js:18` passes no `count` to `t()`**, so no plural form can be chosen and English renders "Goals recorded for 1 of 1 fasts". The repo has been here before: `test/test-i18n-plural.js:106` pins `settings.addressbooksEnabledOfTotal` being called with `count`, precisely because a `_one` form was lost once.

## Nice to have

- `server/services/fasting-stats.js:48` and `:85` default to `dateKeyInZone(new Date())`, which resolves to the **UTC** day. Only tests reach it today, but it makes the default "today" exactly the value CLAUDE.md names as the trap.
- `test/test-health-fasting-stats-api.js:47-48` asserts `count >= 1` and then `count === 1`; the floor is dead weight.
- `public/pages/health-fasting-insights.js:18` nests a ternary inside `esc(...)` on an already long template line.

## On the remaining claims

The OpenAPI summary shape is exactly as you describe it and pinned by two description assertions plus a no-`now`-param check. The family-visibility test genuinely drives `visibleFilter`. I ran `test:health-fasting-stats` (17/17) and `test:health-fasting` (33/33) green on the merged tree; the browser suite sits outside `npm test` by design, so I took your 16/16 as reported rather than re-running it.

PechJiri:
Thanks for the detailed review. I pushed `6ed21973` on top of current `main` (`fabab86b`) and updated the PR description.

All requested items are addressed:
- Moved the fasting-insights changelog entry back under `[Unreleased]`.
- Corrected the stats tests to use the explicit `timeZone` option and removed implicit UTC date defaults from the helpers.
- Aligned aggregate dates, history (including `/fasting` and cursor pages), and CSV filters on the household display time zone. `start_tzid` remains unchanged as record provenance; date-line regressions cover history and CSV.
- Reused `shiftDateKey`, pass `count` into the pluralized goal-coverage translation, removed the redundant count assertion, and extracted the weekly chart markup helpers.

Verification on this candidate: `npm run test:health-fasting`, `npm run test:health-fasting-stats`, the focused partial-goal-coverage browser test, the focused household-zone date-line API tests, OpenAPI/changelog/i18n/time-zone checks, and `git diff --check` all pass. A final independent Superpowers review against this review found no actionable issues. No schema or migration changes.

ulsklyc (changes requested):
Round 2. I measured the current head `6ed21973` against the head I reviewed in round 1
(`cd322b21`), then re-measured everything on the merge with current `main`. Every round-1
finding is resolved, and the two that mattered are now backed by tests I could turn red.

## Round 1, resolved

- **CHANGELOG position.** The bullet sits under `## [Unreleased]` at the head, and it stays
  there in the merge result: after resolving the conflict, the bullet is at line 69,
  `## [Unreleased]` at line 8, `## [2.67.0]` at line 73. `main` has since grown its own
  `[Unreleased]` entries, so this now surfaces as a real conflict instead of a silent slide.
- **The zone test that measured nothing.** `test/test-health-fasting-stats.js` passes
  `timeZone` now, and the helpers reject a missing zone outright. I mutation-tested it:
  putting `row.start_tzid || timeZone` back into `weeklyFastingSeries` and `fastingStreaks`
  turns two tests red. The test earns its keep.
- **`addDays()`.** Gone, replaced by the shared `shiftDateKey()`.
- **`t()` without `count`.** Fixed, and the coverage line is unreachable below two records
  anyway.
- **The zone seam.** `history` and CSV now filter on the household display zone like the
  aggregates do. Counter-proof: reverting `rowMatchesFastingDateRange` to the recorded-zone
  key makes `test-health-fasting-api.js` fail on exactly the new test. This matches the
  decision we settled on.

On the merged tree I ran 15 suites (`health-fasting-stats`, `health-fasting`, `changelog`,
`openapi-structure`, `openapi-coverage`, `suite-chain`, `i18n`, `i18n-plural`,
`i18n-translated`, `household-timezone`, `frontend-audit`, `health-structure`, `health-api`,
`display-timezone`, `docs-landing`) and they are all green. The new browser probe passed
3 runs out of 3 with no flake.

## Please fix before merge

1. **Rebase.** The PR is `CONFLICTING` against `main` (`ba79cbe4`). Both conflicts are
   mechanical: `CHANGELOG.md`, where `main` added `### Added` bullets under `[Unreleased]`
   and you added `### Improved` (keep both sections, yours after), and the `test` chain in
   `package.json`, where `main` appended `npm run test:display-account` and you appended
   `npm run test:health-fasting-stats` (keep both). Nothing semantic collides.

2. **`public/pages/health-fasting-insights.js` belongs in `public/components/`.**
   The module exports one pure-markup function and owns no route, no mount and no layout.
   It only carries

   ```js
   // Delegated panel composition: dashboard, supplied by the Health page.
   // data-composition="dashboard"
   ```

   so that frontend-audit PAGE-001, which requires every file under `public/pages/` to declare
   exactly one composition mode, lets it through. I checked: delete those two comment lines and
   PAGE-001 goes red while everything else stays green. That is a guard being satisfied by a
   declaration the file does not own. The three sibling markup helpers for this same feature
   live in `public/components/` (`fasting-controls.js`, `fasting-dial.js`, `fasting-help.js`)
   and are precached in `sw.js` `APP_SHELL`. Moving the file there lets you drop the marker and
   move the `sw.js` entry from `PAGE_MODULES` to `APP_SHELL`.

3. **The history list now contradicts its own filter, and nothing tells the user.**
   The date filter selects household-zone completion dates, but each row is still stamped in
   its captured zone (`public/pages/health-fasting.js:178`, `fastingStamp(row.start_at,
   row.start_tzid)` plus the zone name). Your own test proves the divergent case: filtering
   `from=2026-09-12&to=2026-09-12` returns a record whose visible stamp reads 13.09. in
   Pacific/Kiritimati. The behaviour is the one we decided on, but the filter labels at
   `public/pages/health-fasting.js:61` say only "completed from / to". Please add the zone to
   that form the way the chart does it, for example a `fastingHelpHtml` note or a hint naming
   the household calendar, so the result set stops looking like a bug to the person reading it.

## Smaller things, your call

- `server/services/fasting.js:256`: `const today = fastingDateKey(now, zone)` re-derives what
  `todayKey(database, now)` already returns. Same answer here, but the house helper is the one
  place this invariant is supposed to live.
- `server/services/fasting.js:190` and `:217`: `const zone = range.from || range.to ?
  householdTimeZone(database) : null` only works because `rowMatchesFastingDateRange` returns
  early on an empty range. Since `fastingDateKey` now throws on a missing zone, that is an
  implicit ordering between two files. One unconditional `sync_config` read is cheaper to read
  than the coupling.
- `test/test-fasting-insights-browser.js:68`: the seeded 2025-01-01 record falls outside the
  seven-day window, so the assertions at lines 71 and 72 would hold with no seed at all.
- `test/test-fasting-insights-browser.js:51`: `assert.match(weeklyText, /2 h/)` reads the whole
  chart's text content and would also match a "12 h" in any other bucket. Scope it to the day
  element.
- `public/pages/health-fasting.js:79`: the stats request fires on every `refresh()`, including
  after each save and delete, and `getFastingStats` reads the subject's full completed history
  each time. Fine at family scale, just noting it doubles the round trips on a page that
  refreshes often.
- `### Improved` appears once in the whole CHANGELOG. This bullet bundles a new panel with a
  behaviour change to existing history and CSV filters; `### Added` or `### Changed` would read
  more like the rest of the file.

Everything from round 1 is genuinely done, and the tests behind it hold up under mutation.
Push the rebase with 2 and 3 folded in and I will take another look.

PechJiri:
Addressed every item from round 2, including all of the smaller suggestions, and force-pushed the rebased branch as `01bb5206`.

What changed:

- rebased directly onto current `main` (`ba79cbe4`), preserving both `test:display-account` and `test:health-fasting-stats`; the fasting changelog entry now lives under the existing `Added` section
- moved the pure insights renderer from `public/pages/` to `public/components/`, removed the synthetic composition marker, updated imports, and moved its SW entry from `PAGE_MODULES` to `APP_SHELL`
- kept history rows in their captured zone but added a localized hint naming the household time zone used by completion-date filters
- switched the stats window to `todayKey(database, now)` and resolve the household zone unconditionally in both history readers
- corrected the weekly browser fixtures to seed a real in-window record and assert the exact value on the exact date bucket
- restored the missing stats/browser npm entry points so every owned test file is exercised by the canonical chain
- stopped routine filter/preference refreshes from re-reading full statistics; record changes and page resume still invalidate them

I also ran the requested two independent Superpowers review rounds. They found a cache-retry edge case and an overlapping-refresh race; both are fixed with regression tests, then both reviewers re-reviewed `01bb5206` and returned PASS. The exact chart assertion was tightened from a substring match to an exact direct-value comparison as well.

Verification:

- reviewer’s exact 15-suite chain: all green
- `test:fasting-browser`: 20/20
- `test:health-fasting-stats`: 18/18
- `test:frontend-audit`: 397/397
- `test:sw-precache`: 7/7
- `test:suite-chain`: 8/8
- final requirements audit and `git diff --check`: PASS

GitHub’s Node 22/24 jobs are running on the pushed head now.

ulsklyc (approved):
Round 3 against head `75f7e6ab`, re-measured on the merge with current `main` (`b592a43f`).

Every finding from round 1 and round 2 is resolved - I checked each one at the code rather than against the commit list, and the two that carried a decision are backed by tests I could turn red.

## Round 1 and 2, verified resolved

- **The zone decision is in the code.** `fastingDateKey` now requires an explicit zone (`server/services/fasting-dates.js:2`), `rowMatchesFastingDateRange` takes it as a parameter (`:40`), and both history and CSV read `householdTimeZone(database)` unconditionally (`server/services/fasting.js:190`, `:217`). Counter-proof on the merged tree: putting `row.start_tzid || timeZone` back turns exactly one test red in `test-health-fasting-api.js` - which is in the `npm test` chain - and one in `test-health-fasting-stats-api.js`. Restored, both green again.
- **`public/components/health-fasting-insights.js`** carries no composition marker any more and sits in `APP_SHELL` (`public/sw.js:107`) next to its three sibling markup helpers.
- **The filter now names its calendar.** `public/pages/health-fasting.js:61` renders `settings.timezoneLabel` plus `display_tzid`, guarded by a new browser test.
- **Insights are no longer refetched on every `refresh()`**: the `loadStats` gate at `public/pages/health-fasting.js:86` plus three browser tests that count the requests, including the in-flight and the failure path.
- `todayKey()`, the unconditional `sync_config` read, the browser seed inside the seven-day window, the scoped `2 h` assertion and `### Added` instead of `### Improved` are all done.
- CHANGELOG placement holds in the **merge result**, not only at the head: bullet at line 54, `## [Unreleased]` at 8, `## [2.67.0]` at 165.
- No migration in this diff. `main` is at 217, so 218 is free for #1179.

On the merged tree I ran `health-fasting-stats` (18/18), `health-fasting` (32/32), `changelog`, `i18n`, `i18n-plural`, `i18n-translated`, `suite-chain`, `openapi-structure`, `health-structure`, `frontend-audit`, `typography`, `household-timezone` and `display-timezone` - all green.

## Blocking

None.

## Should fix

1. **`test/test-health-fasting-api.js:247` - the rewrite deleted the only guard on the CSV export contract.** The zone fix is correct, but the old body ended with two assertions the new one does not replace: the exact nine-column header line, and a match on the `goal_minutes`/`goal_reached` pair in a data row.

   After this PR, `git grep -n 'duration_minutes\|goal_reached' -- test/` returns **nothing** in the whole repo; on `main` it returns that one line. `server/services/fasting-export.js:3` still defines `HEADER`, and both `docs/SPEC.md` and the OpenAPI description (`server/openapi/paths/health.js:134`) still promise those exact nine columns - so the column set and the `goal_reached` computation are now pinned by prose only. The new test at line 284 only counts lines and greps for a note substring. Please fold both assertions back in; the Kiritimati record has `goal_minutes: 60` and a reached goal, so a match on `,"60","true",` works against the new seed.

2. **`test/test-health-fasting-stats.js:21-29` and `:31-42` are the same test twice.** "streak and weekly completion dates use one explicit display zone across records" and "insights can use one display zone consistently across records" seed the same two records at the same instants, in the same zone, and assert the same `{current:1,longest:1}` and the same `series.at(-1)`. The only difference is the `id` fields, which neither helper reads. Drop one.

## Nice to have

- **`server/services/fasting-dates.js:7` builds an `Intl.DateTimeFormat` on every call, and `getFastingStats` now calls it three times per record** - once in the year filter (`server/services/fasting.js:261`), once in the 30-day filter (`:262`) and once per row inside `weeklyFastingSeries`. Measured here, a per-call formatter is 13.4x the cost of a hoisted one (93.5 ms vs 7.0 ms for 2000 calls), so a 500-record history spends around 70 ms per stats request building formatters. `fastingStreaks` already does this right - `server/services/fasting-stats.js:62` hoists one formatter and the comment above it says work must not grow per record. Same treatment for the completion-key path, or accept a `formatter` argument in `fastingDateKey`.
- **`public/pages/health-fasting.js:87` sends `from`/`to` to `/health/fasting/stats`.** `q` is `query(view)`, which carries the history filter; the endpoint declares only `user_id` (`server/openapi/paths/health.js:116`) and `getFastingStats` calls `getAllFastingHistory` with no options, so they are silently dropped. Harmless today, but it makes the URL look like the statistics honour the filter. A `user_id`-only query would say what is actually true.
- **The CHANGELOG bullet has no `(#1178)`.** All three sibling bullets under `[Unreleased]` / `### Added` carry their reference.
- **`public/pages/health-fasting.js:61` names the zone but not what it does.** The hint renders the household zone beside two inputs labelled "completed from / to"; it states the zone without connecting it to the completion dates. The chart solves the same problem one section up with a `fastingHelpHtml` note - a sentence like "completion dates are counted in this calendar" would close the loop the round-2 note was about.
- **The regression cover for both round-2 UI fixes lives outside `npm test`.** The filter-zone hint and the `loadStats` gating are asserted only in `test/test-fasting-journal-browser.js`, which hangs off `test:document-guards`. That matches the documented convention, so this is an observation rather than a request - just worth knowing that CI will not notice if either regresses.
- **`test:fasting-browser` now runs three puppeteer files in parallel**, around 230 s wall time. I ran it three times on the merged tree: 20/20 twice, and once all three files - including the two this PR does not add - were cancelled together at about 31 s with "Promise resolution is still pending". Individually each file is green (insights 3/3, journal 15/15), and the two untouched files failing alongside points at my machine rather than at the new tests, but a third parallel browser instance is what changed.

## Verdict

approve - everything from rounds 1 and 2 is genuinely done and holds up under mutation; please fold the two deleted CSV assertions back in before merge.

PechJiri:
Implemented all items from review 5247955531 and rebased the final candidate onto current `main` (`8e809128`).

### Should fix

- Restored the exact nine-column CSV header assertion and the `goal_minutes` / `goal_reached` data-row assertion.
- Removed the duplicate display-zone statistics test.

### Nice to have

- Reused one completion-date formatter per history, export, and statistics aggregation path.
- Split history and statistics query construction; `/health/fasting/stats` now receives only `user_id` when needed.
- Added `(#1178)` to the Unreleased changelog entry.
- Made the filter hint explicitly connect the selected time zone to completion dates, with unit regression coverage in the regular `npm test` chain.
- Serialized all three Puppeteer fasting files with `--test-concurrency=1`.
- During self-review, corrected partial goal-coverage pluralization in all 24 locales and replaced the non-standard long dash in the new hint.

### Verification

- `npm test` - pass (full chain before the final `main` movement)
- `npm run test:fasting-browser` - 20/20
- Post-rebase: `test:kanban-columns`, `test:health-fasting`, `test:health-fasting-stats`, `test:i18n`, `test:i18n-plural`, `test:i18n-translated`, `test:changelog`, `test:suite-chain`, and `test:frontend-audit` - pass
- `git diff --check` - pass
- Two independent reviews: reviewer-compliance PASS; standards re-review PASS

Final head: `e387ec9f79b4cfa8792a032e74f80e2ed806717a`

PR #1179, opened 2026-09-13
feat(health): add fasting reminders

PechJiri:
## Summary

Add optional owner-only fasting reminders through Yuvomi's existing notification pipeline. Refs #1173.

## Merge order

1. #1177 — fasting journal (merged)
2. #1178 — fasting insights (merged)
3. **#1179 — fasting reminders (this PR)**
4. #1180 — fasting dashboard widget (next)

This branch is rebased directly onto current `main` (`5b44508c`). Its review delta is one commit.

## Changes

- Independent goal-reached and next-start preferences, both off by default.
- Unavailable controls for no-goal and 24-hour-or-longer goals, while retaining the saved preference.
- Per-fast scheduling with atomic lifecycle/settings reconciliation, cancellation after a later fast, immediate permission reconciliation and periodic repair.
- Owner-only delivery through the existing poll, browser and push channels with localized neutral text and `/health/fasting` deep links.
- Append-only migration 218 preserving existing reminder rows/types and the task/event cleanup triggers introduced by migration 217.
- Focused service, API, migration, transaction, localization and browser coverage; OpenAPI, spec, test docs and changelog updated.

## Reviewer follow-ups addressed

- The changelog entry is under `[Unreleased]`, not the published 2.67.0 block.
- Immediately before publication, current `main` still ended at migration 217; this PR therefore uses the next contiguous number, 218.
- The migration regression test starts from the migration-217 trigger state and verifies preservation of `waste_pickup` plus functional task/event cleanup after the rebuild.

## Verification

- `npm test` — passed on exact commit `5a76ed13`.
- `npm run test:fasting-browser` — 21/21 passed, 0 skipped, on the same commit.
- `npm run test:fasting-reminders` — 10 server + 3 browser-loader tests passed.
- `npm run test:reminders-routes` — 32/32 passed.
- `npm run test:notifications` — 46/46 passed.
- `npm run test:household-member-guard` — 47/47 passed.
- `npm run test:permissions-routes` — 14/14 passed.
- `git diff --check origin/main` — clean.

The browser harness disables service workers, so this is not an installed-PWA end-to-end claim. The full `test:document-guards` release handrail remains a separate pre-release requirement.

## Checklist

- [x] Rebased onto current `main`
- [x] Full repository test chain passes
- [x] Focused fasting browser suite passes
- [x] Follows `CONTRIBUTING.md` and `DESIGN.md`
- [x] No new frontend dependency
- [x] Visible strings are localized and documentation/changelog are updated

Implementation and reviews are Codex-assisted; final merge is maintainer-controlled.


ulsklyc:
v2.67.0 shipped today (`fabab86b`), and it moved the whole `[Unreleased]` section down into the release block. Your CHANGELOG bullet still merges without a conflict, but it lands **inside** the published `## [2.67.0]` section. I measured that on the merge tree rather than on your branch, where it sits correctly.

When you rebase this onto the restacked #1178, please move the bullet back under `## [Unreleased]` and check the result with:

```bash
```

Nothing signals this on its own: git merges the hunk cleanly, GitHub reports the PR as mergeable, and `test:changelog` checks structure rather than placement. It is not something you could have seen from your side.

ulsklyc:
Nachtrag zur Migrationsnummer: **214 ist inzwischen belegt** - ich hatte dir am Vormittag die 214 fuer diesen PR angesagt, und seitdem ist #1239 ("who did a task", Issue #1205) auf main gelandet und hat sie genommen. Auf main steht jetzt `version: 214` als letzter Eintrag.

**Deine Migration nimmt beim Neuaufsetzen die 215.** Eine Luecke ist keine Option: `test/test-migrations-append-only.js` meldet jede fehlende Nummer zwischen zwei Eintraegen, 215-mit-Loch-bei-214 waere also rot.

Sonst aendert sich an der Absprache nichts - #1178 zuerst, danach dieser hier auf `feat/fasting-insights` (base umhaengen), zuletzt #1180. Entschuldige den Zusatzaufwand; weil der Stapel ohnehin neu aufgesetzt werden muss, faellt es hoffentlich nicht ins Gewicht.

ulsklyc:
Correction to my last comment, and an apology for two things: it went out in German by mistake, and the number in it is already stale. Here it is properly.

**The migration number is 217, not 215.** After #1239 took 214, the display account PR (#1241, issue #1208) landed and took 215 and 216. `main` is at `ba79cbe4` and its last migration entry is `version: 216`, so the next free one is 217.

I have now told you three different numbers, which is my fault and not a coincidence, so let me give you the rule instead of a fourth one. **Do not fix the number until your final push.** Read the last `version:` in the `MIGRATIONS` array of `server/db.js` on current `main` immediately before you push, and take the next one. A gap is not an option, since `test/test-migrations-append-only.js` fails on any missing number between two entries, and anything I promise in advance is only true until the next PR merges. That is a property of an append-only list with several open branches, not something either of us can plan around.

Nothing else about the sequence changes: #1178 first, then this one on `feat/fasting-insights` with the base moved over, then #1180. And the CHANGELOG hazard from my earlier comment still applies, with one difference that makes it easier: `main` now has its own entries under `## [Unreleased]`, so your bullet will conflict there rather than sliding silently into the published block. Keep both sides when you resolve it.

PechJiri:
Prepared this PR for final review and rebased it onto current `main` (`5b44508c`).

The three existing reviewer notes are addressed:

- the fasting-reminders changelog entry is under `[Unreleased]`;
- the migration number was selected from current `main` immediately before push: main ends at 217, so this PR uses contiguous migration 218;
- the reminder-table rebuild preserves the existing `waste_pickup` type and restores the task/event cleanup triggers from migration 217. The migration regression now verifies both row preservation and trigger behavior.

The restack review also caught and fixed two merge-sensitive issues: the polling toast deep link is attached to the toast rather than the dismiss action, and role permission reconciliation now uses the canonical household-member predicate so service/display accounts are not treated as family members. Next-start notification copy uses the dedicated localized reminder label (including Czech `Připomínka půstu`).

Verification on exact commit `5a76ed13`:

- `npm test` — passed
- `npm run test:fasting-browser` — 21/21 passed, 0 skipped
- focused reminders — 10 server + 3 browser-loader tests passed
- reminders routes 32/32, notifications 46/46, household-member guard 47/47, permissions routes 14/14
- `git diff --check origin/main` — clean
- GitHub Actions — Node.js 22.x and 24.x both passed

This leaves #1179 as a single commit directly on current main. #1180 remains next in the agreed merge order.

ulsklyc (changes requested):
Thanks for the quick rebase, and for turning this into a single commit on top of `main`. I re-reviewed `5a76ed13` against its base `5b44508c`; line numbers below refer to `5a76ed13`.

### Resolved

- **Trigger restore.** `server/db.js:8708-8737` drops and re-creates `trg_reminders_tasks_ad` / `trg_reminders_events_ad` with the v217 bodies. Measured: `main` at `5b44508c` through 217, then your `up`: it applies, both triggers are back, all four `reminders` indexes survive.
- **`waste_pickup` preserved.** Same run with an existing waste reminder: the row survives and the type stays insertable.
- **CHANGELOG placement.** `CHANGELOG.md:59` sits under `[Unreleased]` / `### Added`, right after the insights bullet.
- **Quadratic next-start scan.** `rows.some(...)` is replaced by a single `latestFastId` (`server/services/fasting-reminders.js:94`, `:104`).
- **Role reconciliation.** `server/routes/permissions.js:90` now selects users through `householdMemberSql('u')`, so staff, split guests and display accounts are skipped. The periodic sync still covers everyone with saved preferences, so nothing falls through.
- **Toast deep link.** The link only fires from the toast body (`public/reminders.js:349`); the dismiss button keeps its own handler, and the URL guard (`/` but not `//`) is in place for both the toast and the browser notification.

### Blocking

- `server/db.js:8699` - **migration number and type list have to move again, and that is on us, not on you.** We merged #1255 at 16:51Z, after your push, and it took migration 218 with its own `reminders` rebuild, which adds `document_expiry`. Please take the number by the same rule as before: the last `version:` in `MIGRATIONS` on `main`, plus one, read immediately before your final push. The number alone is not enough, though. Your CHECK list does not contain `document_expiry`, which gives the same failure `waste_pickup` would have caused. Measured on `main` at `67ecd75a`, full chain through 218, then your `up`:
  - with an existing document reminder: `CHECK constraint failed`, and the server does not start;
  - without one, it applies, but `document_expiry` can no longer be inserted. Your block renumbered to 219 as-is makes `test:document-expiry` fail 13 of 20.

  Copy the list from the last `reminders` rebuild on `main` at push time and append your two types. The trigger drop/re-create can stay as it is; #1255's rebuild uses the same bodies.

  When you renumber, move the literals with it: the `218:` key in `server/db-schema-test.js:1343`, `test/test-fasting-reminders.js:30`, `test/test-fasting-reminders-migration.js:12` (and the test name at `:52`), `test/test-fasting-reminders-transactions.js:33`, `test/test-health-fasting-service.js:34`, and `docs/SPEC.md:2980` / `:3019`. On `main`, `version === 218` now finds the documents migration, so a leftover literal applies that one instead of yours without any error. Expect conflicts in `server/db.js`, `server/routes/reminders.js`, `server/services/notifications.js`, `public/reminders.js`, `package.json` and `CHANGELOG.md`, where `main` added `document_expiry` next to your types; keep both sides.

### Should fix (still open from the first round)

- `test/test-fasting-reminders.js:19`, `test/test-fasting-reminders-migration.js:26`, `test/test-fasting-reminders-transactions.js:24`, `test/test-health-fasting-service.js:24` - the fixtures still build `reminders` by hand. They now include `waste_pickup` and the triggers, but the copy is exactly why the `document_expiry` loss above stays invisible to your suites. Building the predecessor from `MIGRATIONS.filter((m) => m.version < N)` keeps the test in step with `main`, and then an existing row of every type can be asserted to survive.
- `server/services/fasting-reminders.js:90` - the permission gate of the sync is still untested. With `userCanUseFasting(database, userId)` replaced by `true`, everything stays green: `test:fasting-reminders` 10/10 and 3/3, plus `test-health-fasting-service.js`, `test-reminders-routes.js` and `test-permissions-routes.js` together 59/59. A test that sets `health_use_fasting` (and once the Health module) to `none` and expects the pending row to be gone would pin what the SPEC promises.
- `server/services/fasting-reminders.js:49-50` - deleting a fast still removes only pending reminders. One that already fired but was not dismissed comes back from `GET /reminders/pending` and now even links to `/health/fasting` for a fast that no longer exists. Delete all reminder rows of the fast on delete, or filter the two fasting types in `/pending` with `EXISTS (SELECT 1 FROM health_fasts WHERE id = r.entity_id)`.
- `server/routes/reminders.js:276-281`, `public/reminders.js:230` and `:286` - the polled toast still gets its fasting text in the household locale, while the rest of the toast uses the device locale through `t()`. Building the body on the client, as `cycleReminderBody()` does, keeps push and channels on the server text and the toast in one language.

### Nice to have

- `server/services/notifications.js:225` - the next-start push now reads "Fasting" / "Fasting reminder" (`health.fasting.remindNext`); the previous "Ready for your next fast" said more. The checkbox labels at `public/components/fasting-controls.js:50-51` have the same gap as before.
- `server/services/fasting-reminders.js:93`, `:117` - the sync still loads the full history every minute and keeps syncing every user who ever had a fasting reminder. Linear now, so not urgent.
- `public/reminders.js:349` - the toast body is clickable only with a pointer; keyboard users reach the dismiss button but not the new deep link.
- `server/services/notifications.js:464`, `public/reminders.js:428-431`, `docs/SPEC.md:1700` and `CHANGELOG.md:59`: the hard-coded English titles, the test-only exports, the missing `entity_type` entries in SPEC and the internal CHANGELOG wording are unchanged from my first round.

### Verdict

Request changes, for the migration number and the `document_expiry` type list only; everything else was either fixed or can come with the same push.

PechJiri:
Review 5250387973 is fully addressed in fe7944dc, rebased as one commit on current main 4135d8a6.\n\nBlocking:\n- Migration is v219 (current main ends at v218).\n- The reminders rebuild copies the full v218 type list, including document_expiry, and appends fasting_goal / fasting_next_start.\n- The migration fixture derives its predecessor from MIGRATIONS below v219 and verifies that every predecessor reminder type and delivery state survives.\n\nShould-fix items:\n- Capability and Health-module revocation remove pending fasting reminders; permission writes and cleanup are atomic.\n- Deleting a fast removes all of its reminder rows, including pushed and dismissed states, in the same transaction.\n- Polling builds fasting copy in the device locale; push and external channels use the household locale.\n- Household Health disablement and account role/family-role changes reconcile immediately and atomically.\n\nNice-to-have items:\n- More descriptive goal/next-fast copy and labels were added in all 24 locales.\n- Periodic reconciliation considers opted-in settings plus pending rows and reads only the latest fast.\n- Toast deep links are keyboard accessible and keep dismiss separate.\n- Server-side fasting titles are localized; test-only reminder exports were removed.\n- SPEC and the user-facing changelog document the entity types, ownership, reconciliation and locale behavior.\n\nVerification:\n- Two Superpowers review rounds completed; all findings were fixed.\n- Final reviewer-checklist audit completed against the exact review body.\n- Full repository suite: npm run test-parallel — 343/343 green.\n- Targeted suites also pass for migration/reminders, permissions, preferences, auth role transitions, polling, push, notifications, document expiry and Inventory reminder compatibility.\n\nThe final full-suite pass caught and removed an over-broad disabled-module filter before push: only fasting reminders opt into the new household-Health boundary, so existing Inventory reminder behavior remains unchanged.

PechJiri:
Follow-up after `main` advanced again:

- Rebased the single PR commit onto current `main` `e986e98e` (new head `fdc9c619`).
- Reconciled the overlap with #1285 by registering `fasting_goal` / `fasting_next_start` in the new shared `ORIGIN_MODULE` map and using its common disabled-module filtering for polling and delivery; the previous PR-specific filter was removed. This supersedes my prior note that only fasting opted into that boundary—current `main` now applies the shared rule to all reminder origins.
- Rechecked the migration immediately before push: `main` still ends at v218, this PR is v219, and preserves `document_expiry` plus every predecessor type.
- Reran the complete suite after conflict resolution: `npm run test-parallel` — **344/344 green**. Focused fasting, disabled-module, reminder-route, notifications, preferences, permissions and push suites are also green.
- `git diff --check origin/main...HEAD` is clean and the PR commit parent is exactly current `main`.

The review 5250387973 checklist remains fully covered, including all should-fix and nice-to-have items.

ulsklyc (changes requested):
Thanks for this round, and for working through the whole list rather than only the blocking part. I re-reviewed `fdc9c619` against its base `e986e98e`; line numbers below refer to `fdc9c619`.

### Resolved, and checked against a broken copy

For each point I switched your fix off in a copy of the branch and ran the suites again. Every one of them went red, so the tests hold the behaviour, not just the text:

- **Migration 219 and the type list.** `server/db.js:8851` carries the full v218 list plus the two fasting types. With `document_expiry` taken out of it, the predecessor-derived migration test in `test:fasting-reminders` fails with `CHECK constraint failed`, and `test:document-expiry` fails 13 of 20.
- **Fixtures.** `test/helpers/migrated-database.js` plus one row per predecessor type (`test/test-fasting-reminders-migration.js:35`) is exactly what makes the check above work.
- **Permission gate.** With `userCanUseFasting()` replaced by `true`, five suites fail: fasting-reminders, notifications, permissions-routes, preferences-routes and admin-password-reset. Removing only the Health-disabled check, or only one of the sync calls in `permissions.js`, `auth.js`, `preferences.js` or the scheduler, turns its own test red as well.
- **Deleting a fast.** Without the `DELETE` at `server/services/fasting.js:420`, `test/test-health-fasting-service.js:78` fails.
- **Toast language.** With `fastingReminderBody()` returning `null`, `test/test-reminders-browser-notification.js` fails.
- **The origin map from #1285.** `fasting_goal` and `fasting_next_start` are in `ORIGIN_MODULE` (`server/services/reminder-origins.js:36-37`); without them `test:disabled-module-reminders` and `test:reminders-routes` fail.
- The nice-to-haves are all in: labels, the next-start text, the latest-fast-only sync, localized server titles, the removed test exports, SPEC and the CHANGELOG entry.

`npm test` passes on your head and on local merges with `main` at `59c02232`, `624ab72d` and `f3c0bbee`; against the current `aedf80d8` I ran the database, migration and fasting suites, all green.

### Blocking

- `public/reminders.js:324-325` and `:363-370` - **tapping a reminder toast no longer dismisses it, for every type except fasting.** On `main`, the whole toast is the dismiss target (`toast.addEventListener('click', ...)` at `public/reminders.js:335` on `main`, with `cursor: pointer` on `.toast--reminder`). On this branch the click handler only exists when there is a `targetUrl`, and `cursor: pointer` is gone from `public/styles/reminders.css:33`, so a task, event or document toast can only be closed with its button. I checked it with the same fake DOM your test uses: a task toast has no click handler here, and has one on `main`. Your keyboard change is right for the fasting link. Please keep tap-to-dismiss for toasts without a target, for example by making the text block a button for every type (with a target it navigates and dismisses, without one it only dismisses), and add the task case to `test/test-reminders-browser-notification.js`. The keyboard point came from my last review, so sorry for the extra lap.

### Should fix

- `server/services/notifications.js:410-412` - **a failing fasting sync now stops every reminder in the household.** The `try/catch` around `syncAllFastingReminders()` is gone, and `test/test-notifications.js:787` pins the consequence: a due *task* reminder is not delivered when the fasting sync throws. Every sync next to it (birthdays, pantry, cycle, schedule, waste) is caught and logged, so a bug or one bad row in one module cannot silence all the others. I understand the intent, a member whose access was revoked must not get a stale fasting reminder, and that can be kept without the global stop: catch and log as before, and for that run drop `fasting_goal` and `fasting_next_start` from the due rows. Fasting then fails closed and everything else still goes out. The test would flip accordingly: the task reminder is sent, the fasting one is not.

### Nice to have

- `server/services/fasting-reminders.js:15-24` - `healthDisabled()` is a private copy of `householdDisabledModules()` from `server/services/household-modules.js`, which #1285 added precisely to replace four such copies. `householdDisabledModules(database).has('health')` does the same thing.
- `server/routes/preferences.js:916-919` - the fasting sync runs inside the same transaction for any change to `disabled_modules`, so switching off an unrelated module would also fail if the fasting sync fails. Running it only when `health` enters or leaves the list keeps the atomicity where it matters.
- `test/test-fasting-reminders-migration.js:35` - the predecessor rows are inserted without `pushed_at`, `dismissed` or `assigned_from`, so the delivery state is not actually asserted to survive. Setting them on a few rows would cover it.

### On our side, no action needed

- `main` moved after your push, and all of it is ours: #1286, #1292, #1294 and #1288. The only conflict is the `test` line in `package.json`, and I will resolve that when merging.
- Migration number: 219 is still the next free number on `main`, but #1256 claims the same number and also rebuilds `reminders`. Whichever of the two lands second gets renumbered by me, with both sets of types in the CHECK list. Your predecessor-derived test will flag it if a type goes missing.

### Verdict

Request changes for the toast regression; the scheduler point would be good to fix in the same push. Everything from my last review is done and holds up when the fix is switched off.

PechJiri:
Addressed the complete review in `38e8d48b` (single amended PR commit).

- Restored tap/keyboard dismissal for every reminder toast by making the text block a native button for all types. Target-bearing reminders dismiss + navigate; task and other non-target reminders dismiss only. The browser regression test covers the task path and verifies that it does not navigate.
- Restored scheduler isolation: a failed fasting reconciliation is logged, only `fasting_goal` / `fasting_next_start` are suppressed for that run, and unrelated reminders continue. The regression test verifies the task is sent/pushed while the fasting row remains unsent.
- Replaced the local Health-disabled parser with `householdDisabledModules(database).has('health')`.
- Limited fasting reconciliation in the preferences transaction to actual Health membership transitions. The trigger-backed test now proves unrelated module changes succeed, while a failing Health transition rolls back atomically.
- Extended the v219 migration regression fixture/assertions for preservation of `pushed_at`, `dismissed`, and `assigned_from`. I also mutation-checked this test by temporarily dropping those copied values; it failed on the expected assertion before the production migration was restored.

Verification on the final candidate:
- `npm run test-parallel` — 344/344 steps green, 0 red
- focused reminder / notification / preferences / disabled-module / frontend suites — green
- `git diff --check` — clean
- independent final code review — no Critical, Important, or Minor findings

Per your “On our side, no action needed” note, I did not restack this branch onto the newer `main`; migration 219 and the package test-line conflict remain for maintainer-side merge handling.

ulsklyc (approved):
Thanks - this one is done. I reviewed `38e8d48b` and ran the suites against a local merge with current `main` (`aedf80d8`), because `test:module-write-access` and `test:calendar-reminder-after-start` do not exist on your base yet. Line numbers refer to `38e8d48b`.

### Resolved, and checked against a broken copy

As last time, I switched each fix off in a copy of the branch and reran the suites. All three went red, so the tests hold the behaviour rather than the text.

- **Tap-to-dismiss.** `public/reminders.js:324-326` makes the text block a `<button type="button">` for every type, and the handler at `:363-368` dismisses unconditionally and navigates only when there is a target. The button resets and the focus ring sit on the existing rule in `public/styles/reminders.css:56-76`, all from tokens. Replacing `public/reminders.js` with the `fdc9c619` version under your new test fails at `test/test-reminders-browser-notification.js:103` with `'span' !== 'button'` - and that assertion is the *task* path, not the fasting one. That is exactly the case that was broken, so the guard sits where it needs to.
- **Scheduler isolation.** `server/services/notifications.js:409-419` catches and logs, `:547-548` drops only `fasting_goal` and `fasting_next_start` for that run. The block now sits in the row of the other five syncs, each with its own try/catch, all before the due-row read at `:455`. Removing the filter turns `test/test-notifications.js:787` red; the test pins both halves, the task gets `pushed_at` and the fasting row does not.
- **The preferences gate.** `server/routes/preferences.js:916-921` runs the fasting sync only on an actual Health transition. Without the `healthChanged` condition, `test/test-preferences-routes.js:402` fails on the unrelated-module case. Reading the current value synchronously before the transaction is the right shape here.
- The other two nice-to-haves are in as well: `server/services/fasting-reminders.js:17` now uses `householdDisabledModules(database).has('health')`, and `test/test-fasting-reminders-migration.js:36-38` / `:52-56` set and assert `pushed_at`, `dismissed` and `assigned_from`.

Everything from the earlier rounds still holds: the predecessor-derived fixtures (`test/helpers/migrated-database.js`), the permission gate (`server/services/fasting-reminders.js:16-22`), the full reminder cleanup on delete (`server/services/fasting.js:420-421`) and the device-locale toast (`public/reminders.js:292-301` against `server/services/notifications.js:233-237`). On the delete point I also checked the remaining path: `DELETE FROM health_fasts` exists only at `server/services/fasting.js:418`, and because fasting is own-only, `health_fasts.user_id` and `reminders.created_by` cascade away together when a user is removed. No orphan left.

`fasting_goal` and `fasting_next_start` are in `ORIGIN_MODULE` (`server/services/reminder-origins.js:36-37`), in `VALID_ENTITY_TYPES` and in `DERIVED_ENTITY_TYPES`. On the merge: `test:module-write-access` 22/22, `test:disabled-module-reminders` 12/12, `test:reminders-routes` 33/33, `test:notifications` 49/49, `test:fasting-reminders` 16/16, plus permissions, preferences, push, document-expiry, openapi, i18n, frontend-audit, changelog, schema-mirror and the migration guards - all green. The six new `health.fasting.*` keys are present in all 24 locales.

### Nice to have

- `server/permissions.js:545` - `replaceSubjectPermissions()` has no production caller left. `server/routes/permissions.js:79` and `:123` were the last two on `main`, and this PR moves both to `writeSubjectPermissions()` inside its own transaction at `server/routes/permissions.js:30-36`. Only `test/test-permissions.js` and `test/test-extension-permissions.js` still reach it. Worth either keeping one caller on the wrapper or noting that it is now a test-only export.
- `server/services/notifications.js:547-548` - a fasting row suppressed by a failed sync appears in no counter: `counters.due` is taken after the filter, so it is neither `due` nor `skipped`, and the `log.error` is the only trace. Counting it as `skipped` would make a repeating failure visible in the run figures.

### On our side, no action needed

- The only conflict with `main` is still the one `test` line in `package.json`. It is purely additive on both sides - you add `test:fasting-reminders`, `main` added `test:calendar-reminder-after-start` and `test:module-write-access` - and I will resolve it when merging. Everything else, including `server/db.js` and all 24 locale files, auto-merges.
- Migration: 219 is still the next free number on `main`, which ends at 218. #1256 and #1257 both claim 219 as well and also rebuild `reminders`, with `health_prevention_due` where you have the two fasting types. Whichever lands second gets renumbered by me, with both sets of types in the CHECK list, and I will move the literals in `server/db-schema-test.js`, the four fasting test files and `docs/SPEC.md` along with it. Your predecessor-derived test is what will flag it if a type goes missing, so that part is covered.

### One thing for #1180

#1180 still carries `fdc9c619` as its parent commit, so its copy of `public/reminders.js` is byte-identical to the version with the toast regression, and `server/services/notifications.js:412` there still calls `syncAllFastingReminders()` without the try/catch. Its own conflict with `main` is just the same `package.json` line, nothing substantive changed underneath it. Restacking it on `38e8d48b` would be enough - otherwise merging it would bring both of these back.

### Verdict

Approve. Both points from the last review are fixed and hold up when the fix is switched off, and the rest of the list stayed done. Thanks for the careful work on this one, including the mutation check you ran on the migration fixture yourself.

ulsklyc:
Heads-up before you see activity on the branch, because one thing I told you in the review has
stopped being true.

#1256 just merged as `bcc35edb7`. It took migration 219, and it touches the same delivery pipeline
yours does. When I approved you, the only conflict with `main` was the single `test` line in
`package.json`, and I said I would resolve that at merge time. That is no longer the shape of it:

| | at review time | now |
|---|---|---|
| conflicting files | 1 | 8 |
| conflict hunks | 1 | 14 |
| your migration | 219 was free | 219 is taken, yours becomes 220 |

The files are `package.json`, `public/reminders.js`, `server/db.js`, `server/routes/preferences.js`,
`server/routes/reminders.js`, `server/services/notifications.js`, `test/test-notifications.js` and
`test/test-push.js`.

**This is still my job, not yours.** I am resolving it and renumbering 219 to 220 on my side,
exactly as promised, and I will push the merge to this branch if I can. Nothing is expected from
you right now, and please do not rebase while I am in there - if you have work in flight, say so
and I will stop and wait for you instead.

Two things I am being deliberate about, since they are the parts that go wrong quietly:

- **Both reminder sources have to survive.** You add `fasting_goal` and `fasting_next_start`;
  #1256 adds `health_prevention_due`. Both of you rebuild the `reminders` table, so the surviving
  CHECK list has to carry both sets, and so do `ORIGIN_MODULE`, `VALID_ENTITY_TYPES` and
  `DERIVED_ENTITY_TYPES`. A merge that keeps only one set still passes a lot of suites.
- **Neither sync may stop the other.** Your try/catch isolation in
  `server/services/notifications.js` was one of the fixes I approved. #1256 adds its own sync in
  the same place, and `main` has gained a filter chain since (#1285, #1289). All of it stays, each
  sync with its own try/catch, before the due-row read.

I am running the full `npm test` chain on the merge result rather than a selection, because this
is the case where per-PR green says nothing: both of your branches were green against `main`
separately and are still incompatible with each other.

Your approval stands. This is merge mechanics, not a new review round.

PR #1180, opened 2026-09-13
feat(dashboard): add fasting widget

PechJiri:
## Summary

Add an optional, self-only fasting dashboard widget using the Health journal's
shared timer and controls. Refs #1173.

## Current integration state

#1177, #1178 and #1179 are merged. The current branch head
`d81f9eff43546bd8200577a20641d732638e4586` is the maintainer-created merge
of current `main` (`d01e6c02441670840fe9439b6a9c8fcb4d696548`) into the
reviewed widget tip. It is mergeable.

The widget diff contains neither `public/reminders.js` nor
`server/services/notifications.js`; the reminder toast and notification
error-handling fixes from #1179 therefore remain intact.

## Changes

- Widget picker/catalog integration with Health and fasting capability gating.
- Circular timer adapting to all supported widget sizes.
- Elapsed/remaining switch, goal status, start/end actions and journal navigation.
- Shared safety acknowledgement and lifecycle behavior, no separate data store.
- Permission and browser tests, translations and contributor documentation.

## Verification

- `npm test` passed (exit 0) on the restacked widget tree.
- `npm run test:fasting-browser`: 23/23 passed.
- On the final integration state: `npm run test:denied-module-reminders`,
  `npm run test:dashboard-fasting` (7/7), and `npm run test:suite-chain`
  (10/10) passed.
- `git diff --check` is clean; `fasting` remains opt-in through
  `DEFAULT_HIDDEN_WIDGETS`, and dashboard CSS imports the shared controls.

The browser harness disables service workers, so this is not an installed-PWA
end-to-end claim. Final merge remains maintainer-controlled.

## Assistance

Drafted with Codex.


ulsklyc:
v2.67.0 shipped today (`fabab86b`), and it moved the whole `[Unreleased]` section down into the release block. Your CHANGELOG bullet still merges without a conflict, but it lands **inside** the published `## [2.67.0]` section. I measured that on the merge tree rather than on your branch, where it sits correctly.

When you rebase this onto the restacked #1178, please move the bullet back under `## [Unreleased]` and check the result with:

```bash
```

Nothing signals this on its own: git merges the hunk cleanly, GitHub reports the PR as mergeable, and `test:changelog` checks structure rather than placement. It is not something you could have seen from your side.

ulsklyc (changes requested):
Thanks for the widget. I reviewed only its own commit, `5d1f2052` against `e1ac3c88`; line numbers below refer to `5d1f2052`.

**Next step: rebase onto #1179.** This is the agreed order, not a finding. #1179 is now a single commit on top of `main` (`5a76ed13`), so please rebase this branch onto #1179's head, leaving out the old journal, insights and reminders commits. #1179 still has to take a new migration number, so expect to follow it once more; the widget itself needs no migration. Three things move with it:

- `main` added the `waste` widget to the same line of `WIDGET_IDS` (`public/utils/dashboard-widgets.js:42`). Keep both sides.
- `test/test-dashboard.js:2201` then has to say 20. `main` already changed that literal from 18 to 19, the same edit as yours, so git merges it without a conflict and the suite only turns red when it runs.
- CHANGELOG: under `## [Unreleased]` / `### Added`, directly after the reminders bullet from #1179, above `### Changed`.

### Blocking

- `public/utils/dashboard-widgets.js:148` - `fasting` is not in `DEFAULT_HIDDEN_WIDGETS`, so the widget is switched on in every existing dashboard layout with the update. The edit in `test/test-dashboard.js:2252` shows it: `fasting` now appears in the visible list of an existing, user-ordered layout. On your branch the effect is limited because `health_use_fasting` has no `default` there and falls back to `none`; on `main` the journal landed with `default: 'allow'` (`server/permissions.js:127` on `main`), and Health defaults to `write`. After the rebase every member with Health access therefore finds a fasting timer with a "Start fast" button on their start page, without having asked for it. The PR and #1173 describe the widget as optional, and the cycle widget, the closest precedent, is hidden by default. Please add `fasting` to `DEFAULT_HIDDEN_WIDGETS` and restore the expectation at `:2252`.
- `public/styles/dashboard.css:5230` (after the rebase) - the widget relies on `fasting-controls.css` for the dial, its segments and the clock switch, and so do the safety and finish dialogs it opens (`public/styles/fasting-controls.css:1` says so). On your branch that works because the old journal commit `5de23b66` linked the stylesheet globally in `public/index.html:71`. The journal that landed on `main` (`546eb3ee`) does not: it loads the file through `@import url('/styles/fasting-controls.css');` at the top of `public/styles/health.css`, and the router keeps only one page stylesheet active at a time. After the rebase the dashboard never loads it, and nothing in `npm test` would notice. Put the same `@import` at the top of `public/styles/dashboard.css`.

### Should fix

- `server/routes/dashboard.js:679-689` - the new `fasting` key of `/dashboard` has no test in `npm test`. `test/test-dashboard-fasting.js` only checks the registration, and `test:fasting-browser` runs outside CI. This is health data inside a household-wide endpoint, so it deserves a route test: another member's active fast never appears, capability `none` gives `null`, and Health `none` (as well as an API token without the `health` scope) gives the `null` from `DENIED_PAYLOAD`.

### Nice to have

- `server/routes/dashboard.js:680-685` - the route re-implements the capability check, the clock-mode lookup and the two fast queries that `server/services/fasting.js` already owns (`getFastingClockMode` and the state queries). Calling the service keeps the dashboard and the Health page from drifting apart.
- `docs/test-suites.md:11` and `:19` - `test-fasting-reminders-browser.js` is still part of `test:fasting-browser`, but the reminders are no longer mentioned in either line.
- `CHANGELOG.md:28` - the bullet describes internals ("reuses the journal controls"). Something like "**A fasting timer can be added to the dashboard** (Refs #1173). It shows only your own fast and lets you start or finish it there." says what users get.

### Verdict

Request changes. Both blocking points are small edits that fit into the rebase.

PechJiri:
Implemented every item from review 5250388254, including the nice-to-haves.

- Rebased the widget to exactly one commit above the current #1179 head (`fdc9c61916e91969b9ea22a27520d93b987468ed`); the older journal, insights and reminders commits are not duplicated.
- Preserved the `waste` widget from main, added `fasting` as widget 20, and kept existing layouts unchanged by making it opt-in through `DEFAULT_HIDDEN_WIDGETS`.
- Dashboard now imports the shared fasting controls directly.
- The dashboard route delegates fasting state to the fasting service and has real route coverage for self-only data, capability denial, Health module denial, and API tokens without Health scope.
- Added the requested test-suite documentation and user-facing Unreleased changelog entry in the requested position.
- Kept fasting and medication dashboard slices failure-isolated; a fasting load failure now uses the standard retryable widget error. The widget also uses the Health module accent.

Verification on final commit `007743c199d135edd9d9f5d872ba40cd84f5f696`:

- Superpowers review: pass after addressing all findings
- `npm run test-parallel`: 345/345 green
- `node --test --test-concurrency=1 test/test-fasting-dashboard-browser.js`: 2/2 green
- `node --experimental-sqlite test/test-dashboard.js`: 113/113 green
- `npm run test:dashboard-fasting`: 7/7 green
- `npm run test:frontend-audit`: green
- `npm run test:suite-chain`: 10/10 green
- `git diff --check`: clean

The branch has been force-updated with lease and is ready for the next review once GitHub checks finish.

ulsklyc (approved):
Thanks, this round closes everything from my last review, the nice-to-haves included. I re-reviewed the widget commit `007743c1` on top of #1179's `fdc9c619`; line numbers below refer to `007743c1`.

### Resolved, and checked against a broken copy

I switched each fix off in a copy of the branch and ran the suites again. Every one went red:

- **Opt-in.** `fasting` is in `DEFAULT_HIDDEN_WIDGETS` (`public/utils/dashboard-widgets.js:148`). Without it, `test:dashboard-fasting` fails, and so does the existing-layout check in `test:dashboard`, which now finds the widget where it should not be.
- **Styles.** `@import url('/styles/fasting-controls.css');` opens `public/styles/dashboard.css:1`; without it `test:dashboard-fasting` fails.
- **Route coverage.** `test/test-dashboard-fasting.js:139-190` tests the payload over HTTP. Returning another member's active fast, bypassing the capability check, dropping `fasting: null` from the denied Health payload, or letting a fasting failure wipe the medication slice each turns its own test red.
- **Service delegation.** `server/routes/dashboard.js:691` now uses `getFastingState()`, and the medication and fasting slices fail independently of each other.
- `WIDGET_IDS` keeps `waste` next to `fasting`, `test/test-dashboard.js:2285` expects 20, and the docs and the CHANGELOG entry (`CHANGELOG.md:69`) are where I asked for them.

`npm test` passes on this head and on local merges with `main` at `59c02232`, `624ab72d` and `f3c0bbee`; against the current `aedf80d8` I ran the database, migration, fasting and dashboard suites, all green.

### Nice to have

- `test/test-dashboard-fasting.js:44` - the file brings its own `buildMigratedDatabase()`. The helper from #1179 (`test/helpers/migrated-database.js`) does the same and also handles `foreignKeysOff` migrations. Harmless on an empty database, just a second copy to keep in step.

### Merge order

Nothing is left on the widget itself. It lands right after #1179, which still has one change to make. The widget commit does not touch the files that change is about (`public/reminders.js`, `server/services/notifications.js`), so there is no need to restack it again: I will bring it across when merging, together with the `test` line in `package.json`, which conflicts because of our own #1292 and #1294.

### Verdict

Approve. The approval is for the widget commit; #1179 has its own review.

ulsklyc:
Holding this one until it is restacked, and the reason is not the conflict.

#1180 still has `fdc9c619` as its parent, so it carries a copy of `public/reminders.js` that is
byte-identical to the version with the toast regression, and its
`server/services/notifications.js:412` still calls `syncAllFastingReminders()` without the
try/catch. Both of those are exactly what you fixed in #1179 at `38e8d48b`. Merging this as it
stands would bring both back.

Its own conflict with `main` is only the one `test` line in `package.json`, purely additive, and
nothing substantive changed underneath it. Restacking on `38e8d48b` should be all it needs.

The two points from my last review are still open on the widget itself:

- `fasting` is missing from `DEFAULT_HIDDEN_WIDGETS`, while `main` has `health_use_fasting`
  defaulting to allow, so the widget appears for households that never asked for it.
- `dashboard.css` has no `@import` for `fasting-controls.css`.

No rush from my side - #1179 is approved and goes in first, so a restack on top of it is the
natural next step.

ulsklyc:
Thanks for the restack - that was exactly it, and I checked rather than assumed:

- Your branch now sits on `9cb8874e6`, the merged #1179, and `public/reminders.js` is byte-identical
  to `main`. Both defects I was worried about are gone.
- `fasting` is in `DEFAULT_HIDDEN_WIDGETS` (`public/utils/dashboard-widgets.js:148`).
- `dashboard.css` line 1 has the `@import` for `fasting-controls.css`.

That closes both points from my review.

**I pushed one commit to your branch: `d81f9eff4`, a merge of current `main`.** The conflict was mine,
not yours - four of my own PRs went into `main` while this one was waiting (#1287, #1290, #1269,
#1289), and each of them adds a line to the `test` chain in `package.json`. That single line was the
only conflict. I resolved it by union, keeping every step from both sides, and verified nothing was
lost or duplicated: 353 steps from `main` plus your `npm run test:dashboard-fasting` makes 354.

Nothing of yours was touched. Green afterwards: `test:dashboard-fasting`, `test:dashboard`,
`test:suite-chain`, `test:changelog`, `test:i18n`, `test:frontend-audit`.

I will do the proper re-review on this state and update my verdict.

ulsklyc (changes requested):
Full read of the whole diff this time, not just a check of the two earlier points. Reviewed at head `d81f9eff4`, which sits on today's `main` (`d01e6c024` is the merge base).

First, so you do not have to guess what that commit is: **`d81f9eff4` is mine, not yours.** I merged `origin/main` into your branch myself. The only conflict was the `test` chain line in `package.json`, resolved as a union (353 + 1 = 354 entries). Nothing else was touched. Your own commit is `0461f779e`.

**This reverses my own approval, and the fault for that is mine.** I approved this on 19 September at `007743c1`. The blocking point below was already in the branch then, byte for byte: `test/test-fasting-dashboard-browser.js` has the same SHA1 at `007743c1` and at `d81f9eff4`, 94 lines both times. I did not read that file closely enough the first time, and a shallower second look would not have caught it either - this round was a full read of the whole diff rather than a check of the two earlier points.

**The two findings from the first round stay resolved.** `fasting` is in `DEFAULT_HIDDEN_WIDGETS` (`public/utils/dashboard-widgets.js:148`) and `@import url('/styles/fasting-controls.css')` is the first line of `public/styles/dashboard.css`, matching how `health.css` does it. `public/reminders.js` and `server/services/notifications.js` are not in the diff, so the #1179 fixes are intact.

What I measured against your head, all exit 0: `dashboard`, `dashboard-permissions`, `dashboard-fasting`, `permissions`, `suite-chain`, `frontend-audit`, `i18n`, `i18n-plural`, `changelog`, `typography`, `lucide-icons`, `layer-boundary`, `sw-precache`, `health-fasting`, `preferences-dashboard`, `extension-widgets`, `module-registry-parity`, `mobile-scroll-layout`.

I also mutation-tested the new Node suite twice, because "does it measure anything" is the question I care about most:

- Leaking another member's family-visible active fast into `active` (`server/routes/dashboard.js:694`) turns `dashboard fasting payload is self-only ...` red.
- Making the 403 path fall back to a raw `health_fasts` read turns `dashboard fasting payload is null when the capability is denied` red.

So `test/test-dashboard-fasting.js` is a real behaviour suite over a mounted router and real migrations, and it holds all four permission axes: module denial, API-token scope, the `health_use_fasting` capability, and the self-only subject. That is the part of this PR I am most comfortable with, and it is the part I was most worried about. Thank you for building it that way.

## Blocking

- `test/test-fasting-dashboard-browser.js:24-28` - the read-only assertion cannot go red for the reason its test name gives. `setPermissions` rebuilds the whole store (`public/permissions.js:63-71`), so a payload without `capabilities` leaves `capabilities: {}`. `canUseFasting()` (`public/permissions.js:120-123`) then returns `false`, and `public/pages/dashboard.js:400` drops `fasting` out of the grid entirely - `renderFastingWidget` never runs. `!document.querySelector('[data-fasting-widget-action]')` is therefore satisfied whether or not the `moduleAccess('health') === 'write'` gate at `public/pages/dashboard.js:2078` exists at all. What is being measured is a lost capability, not read-only access. CONTRIBUTING, "What a human guarantees": a guard that was never seen red is not evidence. The production gate itself is correct - this is only the test. Two lines fix it: pass `capabilities: { health_use_fasting: 'allow' }` alongside `modules: { health: 'read' }`, and assert that `.fasting-widget` is still present while the action button is gone.

## Should fix

- `server/routes/dashboard.js:693` - `zone_mode` is dropped from the payload, so the dial is not actually the page's dial. `updateFastingClock` reads `settings.zone_mode` (`public/components/fasting-controls.js:329`) and hands it to `fastingDialMarkup`, which defaults to `'timer'` (`public/components/fasting-dial.js:6`); `fastingDialModel` only draws the metabolic zone rings for `'educational'` (`public/utils/health-fasting.js:138-144`). Somebody who picked the educational dial on the Health page gets a different dial on the dashboard. That also undercuts the assertion at `test/test-fasting-dashboard-browser.js:16` ("Widget shares the segmented page dial") and the SPEC line you added. One extra field in `settings` fixes it; everything else is already wired.

- `server/routes/dashboard.js:694-695` - `fastingState.active` and `fastingState.history[0]` are passed straight through, and those are `SELECT *` rows from `health_fasts`, so they carry `note` (up to 2000 characters of free text) and `rating` (`server/db.js:8738-8739`). `/dashboard` is in `API_CACHE_WHITELIST` (`public/sw.js:35`), so a fasting note ends up in the device's Cache Storage for a widget that never reads it. It is the viewer's own data, so this is not a permissions hole, but it is exactly the minimisation the `DENIED_PAYLOAD` comment block above asks for ("die Daten lagen auf der Leitung, im Netzwerk-Tab und im Service-Worker-Cache"). The widget needs `id`, `revision`, `start_at`, `goal_minutes` and the last entry's `end_at`. While you are there: `getFastingState` fetches ten history rows and discards nine, plus `server_now`, `display_tzid`, `acknowledged` and a second `resolvePermissions` on every dashboard request - for a widget that is hidden by default.

- `server/routes/dashboard.js:75` and `:699` - `fasting: null` breaks the shape rule this file states for itself. The comment above `DENIED_PAYLOAD` says the value is always the empty *form*, never a missing field, so the response keeps its shape for `/api/v1` consumers and no client trips over something it did not expect. `null` is not the empty form, and your own client trips over it immediately at `public/pages/dashboard.js:2067`. It also collapses two different situations into one signal: "denied" should make the tile disappear, "the query failed" should show the retry tile, and right now they are indistinguishable. `{ settings: { clock_mode: 'auto' }, active: null, lastCompleted: null }` for the denial would keep both the shape and the distinction.

- `public/pages/dashboard.js:5050` - the new `window.addEventListener('pageshow', ...)` is a dashboard-wide behaviour change coming out of a widget PR, and it duplicates something the component already offers. `startFastingClock(root, active, last, initial, options)` wires `visibilitychange` and `pageshow` itself when given `options.refresh`, scoped to `root.isConnected` and its own `AbortController` (`public/components/fasting-controls.js:363-367`) - that is how `public/pages/health-fasting.js:174` does it. `public/pages/dashboard.js:2097` omits the fifth argument and rebuilds the mechanism at page level instead. The listener also does not check `event.persisted`, so it fires on an ordinary page load too, right after `load`, where it races the `/dashboard` request `render()` just issued. Passing `{ refresh: refreshDashboardData }` would let line 5050 go away; if you would rather keep it, it needs the `event.persisted` guard.

- `public/pages/dashboard.js:2069-2081` - the widget has no live region for the ticking timer. The page carries `<p class="sr-only" role="status" aria-live="polite" data-fasting-announcement>` (`public/pages/health-fasting.js:126`) and the shared updater fills it every minute (`public/components/fasting-controls.js:331-335`), so adding the element is one line and it starts working. The page's read-only hint (`health.fasting.readOnly`, `public/pages/health-fasting.js:129`) is missing too - dropping the action button is right, but then nothing in the widget says why.

## Nice to have

- `public/pages/dashboard.js:1984` - `/health/cycle/*-Endpunkten` became `/health/cycle/(sub)-Endpunkten` in an unrelated comment. I reverted it and ran every suite that reads this file as text plus the audit suites; all green. If there was a reason, it is worth a word, otherwise this is churn on a comment it makes less precise.

- `public/pages/dashboard.js:2069` and `:2078` - `widgetHeader` already links to `/health/fasting`, and the second button links to `/health/fasting#history`. Two links to the same page in one tile.

- `test/test-dashboard-fasting.js:33-34` - pinning `if (!fasting) throw new Error(` as source text measures the spelling rather than the behaviour, and it is the very line that would have to change for the payload-shape point above. The other six tests in that file measure real behaviour, which is why this one stands out.

- CONTRIBUTING, "If you contribute", asks for one line in the description if an agent helped write it. Not a mark against the contribution, it just tells the reviewer where to look.

## Verdict

request-changes, which means withdrawing my approval of 19 September. I want to be plain that this is a correction of my own review, not a moving target: the blocker was in the branch when I approved, I missed it, and the only thing that changed is that I looked properly. Nothing you pushed since then caused it - your one commit in between is `0461f779e`, everything else is `main`.

The permission work here is correct on all four axes and I proved it by mutation; nothing I found is a rights or data-loss defect. The one blocker is a test that cannot fail for the reason it claims, and the rest are small and concrete. Sorry for the extra round.

ulsklyc:
Heads up: the commit `ce379e515` on your branch is mine, not yours. I merged `origin/main` into it
so the PR stops showing as conflicting.

Only one file was actually in conflict: the `test` chain in `package.json`. Your branch had 354
steps, `main` now has 355 (two suites arrived with #1295 an hour ago), and the only step unique to
your side was `npm run test:dashboard-fasting`. I resolved it as a union - 356 steps, nothing
dropped, no duplicates - and checked it with `test:suite-chain` and `test:changelog`, both green.
Everything else merged cleanly; I touched no other file.

The conflict was caused by my own merges, which is why I cleared it rather than leaving it to you.
It changes nothing about the review: the blocking point is still the read-only assertion in
`test/test-fasting-dashboard-browser.js`, and the ball is with you there.

PechJiri:
Prepared for final review in 05e6ef25.

- The read-only browser check now keeps `health_use_fasting: allow` and proves the widget remains visible while its action is absent.
- Also included the requested follow-through: educational dial mode, minimized cached payload, stable denied-payload shape, component-scoped persisted-page refresh, live status/read-only hint, and the small cleanup items.
- Verification: `test:dashboard-fasting` 8/8, fasting browser 2/2, `test:frontend-audit` 401/401, `test:i18n`, `test:permissions`, `test:suite-chain`, `test:changelog`, and `git diff --check` all pass.

I also ran an independent Codex/Superpowers review against this commit; it found no actionable issues. The PR description now discloses Codex assistance as requested.

ulsklyc:
Merged. Thank you for the thorough round - all ten findings were addressed, and I verified each one at the code rather than taking the list on trust.

Three things I resolved at merge time instead of sending the branch back for a fourth round:

- **`package.json`**, the only conflict: the union of both test chains, 357 steps. `main` gained `test:sync-calendar-move-assignment` with #1306 after you last merged; your branch contributes `test:dashboard-fasting`. Nothing lost, nothing duplicated.
- **The CHANGELOG entry** would have landed inside the already published `## [2.68.0]` section, because v2.68.0 went out after your merge. Moved back under `## [Unreleased]` into its own `### Added`. Not your doing - our release commit moved the target.
- **One test name**: `dashboard fasting payload is null when the capability is denied` was still claiming the old behaviour while asserting the empty shape it now returns. Renamed to say what it measures.

Two things I did not touch, neither blocking, both worth knowing:

1. The `event.persisted` guard sits in the shared `public/components/fasting-controls.js`, so `public/pages/health-fasting.js:174` loses its refresh on an ordinary `pageshow` as well. That is correct - a plain `pageshow` also fires on first load, where the refresh was redundant - but it is a behaviour change to the Health page arriving through a dashboard PR, and it is in neither the CHANGELOG nor the PR text.
2. Deleting the source-text pin also removed `assert.doesNotMatch(source, /SELECT \* FROM health_fasts/)`. The payload itself is covered by the new `deepEqual`, but nothing covers any more that a failed slice reaches the client retry tile.

On measurement, so the record is straight: I ran fifteen suites on the resolved merge tree, every exit code read from a file - `dashboard-fasting`, `dashboard`, `changelog`, `suite-chain`, `i18n`, `i18n-translated`, `permissions`, `preferences-dashboard`, `sw-precache`, `module-registry-parity`, `extension-widgets`, `frontend-audit`, `typography`, `layer-boundary`, `lucide-icons`, all exit 0.

The five Puppeteer files under `test:fasting-browser` are **not** among them. They hang here because Puppeteer finds no Chrome in this environment - on `main` exactly as on your branch, so nothing to do with your work. You hooked the new file into `test:fasting-browser` correctly, which is where `docs/test-suites.md` wants it: that chain is the pre-release handrail and deliberately outside CI.

One last thing worth saying plainly: no CI run was ever created for `05e6ef25`. Not a waiting one, none at all. That is a known failure mode here - the comment at the top of `.github/workflows/ci.yml` documents it from two earlier cases - and it means "no checks" on your PR was never you waiting on us to click approve. Sorry for the part of the wait that caused.

PR #1181, opened 2026-09-13
feat(health): cycle tracker v2 - flow analytics, daily insights, privacy

mclgoerg:
The Cycle tab grows from a period calendar into a full tracker, built around three ideas: make logged data **visible** (flow was tracked but almost invisible), make predictions **honest** (measured beats estimated, and the UI says which one it is showing), and keep the household angle **privacy-first** (new intimacy logging is hard-private, the new partner reminder shares a date and nothing else).

### Daily insight bubble
One glanceable line at the top of the tab: cycle day + phase, plus the single most relevant thing today - period expected (with a start action, or an "end it?" action if a logged period is still open), *"today is often your strongest pain day (cramps)"* from your graded intensities, symptoms likely today, an approaching PMS window, or the fertile window. Deliberately silent when there's nothing to say, and pattern-language throughout ("based on your pattern", never a prognosis).

Cycle tab with insight bubble

### Flow strength, finally visible
- 4-step dot scale on the calendar (size + tint), with a legend
- heaviest-flow chip per period in History
- a per-cycle flow-intensity chart in Trends
- a calm, non-diagnostic hint when recent periods run repeatedly heavy or longer than a week

On desktop, the month calendar keeps mobile's compact proportions (a container query splits the section into calendar + annotation rail instead of inflating day cells to 160px squares across the full page width).

Calendar with flow scale, PMS shading and confirmed ovulation

### Richer day log
Cervical mucus (5-step), LH and pregnancy tests, multi-select feelings (replacing the single mood, normalized like symptoms were in migration 178), and intimacy - plus quick links to log a painkiller or weight in the existing modules. The modal now opens at the flow row instead of auto-scrolling past it.

**Intimacy is hard-private by design:** the server strips it from every non-owner read regardless of the day's visibility flag, and the bulk "set all to family" action never exposes it.

Day log with the new fields

### Honest predictions
- A BBT-confirmed ovulation now moves the **month calendar** too, not just the ring - the two can no longer contradict each other on one screen (confirmed cells render solid, predictions stay outlined)
- Implausible cycle gaps (overlaps, future-dated typos) no longer poison the averages, while consistently long cycles (oligomenorrhea) keep their real history instead of falling back to a 28-day default
- The BBT chart is date-proportional and breaks across logging gaps; the severity chart shows its real 1–3 scale
- Symptom likelihood projects into the next cycle instead of only backwards, and the pattern picker now sits next to the calendar it annotates

### New per-person modes
- **Contraception**: hormonal methods pause fertile-window prediction, with the reason shown instead of an empty tile
- **Perimenopause**: next period as a min–max range instead of false-precision, regularity judgement off
- **PMS window**: derived from your own luteal patterns, shading on the calendar (own view only), toggleable

Contraception suppression note
Perimenopause range prediction

### Household features
- **Partner reminder (opt-in by the owner)**: the chosen member gets "«Name»'s period is predicted to start in N days" - the date only, never any log content, and no read access is granted. Switching partners re-targets the reminder; the in-app list names the owner just like the push does.
- Family read-only view respects everything: person-neutral copy, no intimacy, no PMS shading, no insight bubble.

### Everything else
Feelings-by-phase and pain-summary trends, one details expander per symptom row (instead of two stacked disclosures), CSV import of period history (comma/semicolon, `YYYY-MM-DD`/`DD.MM.YYYY`, all-or-nothing with per-row errors), a "next period" tile on the Health overview, and discoverability hints for the existing ICS feed and dashboard widget.

Trends: flow intensity and pain summary
Mobile

## Data model
Migrations 210-213 (209 stays with the fasting journal, per your allocation): four nullable day-log scalars, a normalized `cycle_day_log_feelings` table (178-style backfill from `mood`, column frozen and now actively nulled on write), five `cycle_settings` columns, and a widened `cycle_reminder_anchors.kind` for the partner case. No changes to existing rows beyond the feelings backfill.

## Testing
Full `npm test` green (0 failures). Focused suites grew: `test:health-cycle` 87→145, `test:health-api` 104→129, `test:cycle-reminders` 8→18, `test:reminders-routes` 28→30, `test:health-nav` 19→32, plus new coverage in `test:cycle-symptoms-migration`. An adversarial review pass ran before opening this PR; every confirmed finding (stale partner-reminder recipient, legacy-mood resurrection, overlapping-period bubble action, long-cycle regression, PMS opt-out in family views, misattributed in-app partner reminder, future-date anchor mismatch) is fixed and covered by tests. Browser-verified end to end on a deployed test instance at 390/1440px, light and dark, including a four-account family-visibility matrix.

## Deliberately not done
- Feelings as PMS-window inputs (needs a generic-predicate refactor of `symptomCyclePattern()` - noted in the util docs)
- A mood bridge on the Overview vitals tile (would require nested interactive elements)
- Batched symptom-pattern computation (per-render cost is documented and bounded; follow-up material)


ulsklyc (changes requested):
Thanks, this is a big step for the Cycle tab. Making flow visible and saying which predictions are measured and which are estimated are both good calls. Review below; CI has not run yet, I will approve the run once the conflicts are resolved.

## Decisions on the open product questions

1. **Pregnancy test, LH test and cervical mucus become owner-only, like intimacy.** Today only `intimacy` is stripped for non-owners (`server/routes/health/cycle.js:310-315`). The other three still reach every family member on a day marked `family`, and the bulk "set all to family" action exposes them for every past day at once, although the family view never shows them. A positive pregnancy test is at least as sensitive as the intimacy entry. Please strip them the same way, add a test per field, and name the owner-only axis in SPEC.
2. **The partner reminder only offers people who will actually get it.** `syncPartnerReminder()` (`server/services/cycle-reminders.js`) silently drops the reminder when the chosen person lacks Health access or has the Cycle tab off, and the select in `public/pages/health.js` lists every other member, children included. The Cycle-tab condition hits exactly the most likely recipient: a partner with no cycle of their own. Please drop that condition (Health access is enough), leave children out of the select, and offer only members who pass the same check the sync uses. A recipient-side opt-out is not needed in this PR; dismissing stays the way out.

## Migration numbers

Main is at **208** now, and the fasting journal (#1177) is queued for **209**. Please take **210-213**. Whichever PR lands first takes the next free number. Besides the four `version:` lines this touches:
- `test/test-cycle-symptoms-migration.js:143` (`m.version === 196`) plus its `v196` titles,
- the comments naming the numbers in `cycle.js`, `cycle-reminders.js`, `health-cycle.js`, `test-health-api.js` and `test-health-nav.js`,
- the migration rows and headings in `docs/SPEC.md`.

The branch conflicts with main in at least `CHANGELOG.md`, `server/db.js` and `server/routes/reminders.js`, where main added a `waste_pickup` branch to the pending query.

## Blocking

- **Em-dashes:** 188 added lines in the locale files contain an em-dash (U+2014), plus 3 in the CHANGELOG bullet and 21 in SPEC. The rule is a plain hyphen.

## Should fix

Each checked at the code.

- **The calendar feed still publishes fertility under hormonal contraception.** `server/services/cycle-ics.js:120` only checks `track_fertility`. The tab says the fertile-window prediction is paused, while the subscribed feed keeps sending fertile window and ovulation events. Please apply the same contraception condition as `predictCycle()` and add a test to `test:cycle-ics`.
- **The partner reminder outlives the owner.** The reminder row belongs to the recipient (`created_by`, `cycle-reminders.js:121-122`), while the anchor cascades away with the owner (`cycle_reminder_anchors.user_id ... ON DELETE CASCADE`). `reminders.entity_id` has no foreign key, and `syncAllCycleReminders()` only looks at owners' settings and anchors. So after an admin deletes the owner, the partner keeps a reminder without an anchor, shown as "Next period - null", which they can only dismiss. Please remove cycle reminders whose anchor is gone during the sync, with a test.
- **References to things that do not exist.** 92 added lines cite `D-6` to `D-15` and "DECISIONS.md", but `docs/DECISIONS.md` numbers its entries 1-6. Two of these (`D-13`, `D-15`) are in `server/openapi/paths/health.js`, which is the public API document. Please replace them with the actual reason or a SPEC reference. Please also remove the workflow notes: 28 "Review-Runde" lines, "Task-Vorgabe" in `cycle.js`, and the stale `TODO(W6)` in `server/services/notifications.js:172` (the key exists in all locales).
- **`mood` and the public API.** On main, `mood` is free text (`v.str(b.mood, ...)`). The new route lowercases it and rejects anything outside the fixed list, so `POST` with `mood: "tired"` goes from 201 to 400, while the OpenAPI text still says `mood` is accepted. Migration 196 copies `TRIM(mood)` without `LOWER`, so a stored "Good" is backfilled, never shown and deleted on the next save. Please use `LOWER(TRIM(mood))`, decide what happens to values outside the list, and mention the stricter validation in the CHANGELOG, since `/api/v1` is a promised surface.

## Nice to have

- The migration 195 comment says `ALTER TABLE` cannot add a CHECK, but 197 does exactly that with `ADD COLUMN ... CHECK(...)`.
- `public/reminders.js:258-262`: without `cycle_owner_name`, a partner reminder falls back to the owner's own "Next period" text. A neutral text is safer.
- The PR body promises "«Name»'s period is predicted to start in N days"; the code sends `{{name}}'s period - 2026-06-29`.

## What is fine

No other read path leaks day-log fields: MCP, search, dashboard, ICS and the period export only read periods and settings. The partner reminder carries only the date and the owner's name. The import route needs `health:write`, and settings only ever write the caller's own row. No existing migration was changed, and no `toISOString()` date keys were added.

ulsklyc (changes requested):
Thanks for the thorough round. Most of it holds at the code; two points need another push, and the order with #1177 matters now.

## Checked at 5e631a0e

All four owner-only fields are stripped, with a test per field and the bulk variants. One eligibility function feeds the list, the sync and `PUT`, and children are excluded. The ICS feed checks contraception, with the pill/copper pair. Orphaned partner reminders are cleaned up in the sync, with a test that deletes the owner. Migrations 210-213 are appended after 208, and no existing entry changed. The backfill uses `LOWER(TRIM(mood))`. The OpenAPI text is clean, and no em-dash (U+2014) is left.

## Blocking

- **The merge put the CHANGELOG bullet into a released section.** In `92302125` the cycle tracker bullet moved from `## [Unreleased]` to `## [2.66.1] - 2026-09-14` / `### Security` (`CHANGELOG.md:244`, right after the advisory), and it stays there when merged with current main. The in-app changelog would show it as part of the 2.66.1 security patch, and the notes of the next release would miss it. Please move it back under `[Unreleased]` / `### Added`.

## Should fix

- **A second settings save drops the partner.** `PUT /health/cycle/settings` answers with the plain `cycle_settings` row (`server/routes/health/cycle.js:576`), without `eligible_partners`, and `public/pages/health.js:6578` replaces `cycle.settings` with it. Reopening the settings dialog without reloading the tab builds the partner select from `s.eligible_partners || []` (`:6375`), so only "Off" is left. The next save, say for the PMS toggle, sends `notify_partner_user_id: null`, and the sync removes the partner's reminder without a word. Returning `eligible_partners` from `PUT` as well (the same helper as `GET`) fixes it; a test that saves twice would pin it.
- **En-dashes (U+2013):** 18 added lines; the rule covers both dashes. User-visible ones: `periodToday`, `periodOverdue`, `periodOverdue_one` and `periodStillOpen` in `public/locales/de.json`, the date ranges in `public/pages/health.js:3844` and `:4741`, and the placeholder at `:5612`-`:5613`. The rest is SPEC and test titles.
- **Legacy `mood` values outside the list are not left untouched.** The migration comment and your reply say they stay in the frozen column, but the day-log upsert sets `mood = NULL` on every save (`cycle.js:378`, pinned by `test/test-health-api.js:1084`). A value stored through the API on main is invisible after the update and gone with the next edit of that day. Either keep the column when the request carries no `mood`, or say so in the CHANGELOG bullet next to the stricter validation.
- **Leftover references:** `cycle.js:298` "(Migration 178 bzw. 196)" should say 211, `server/services/cycle-reminders.js:20-21` "Migration 198" should say 213, `public/pages/health.js:4563` still has "D-9", and a test title in `test/test-cycle-reminders.js:321` has "(R-5)".

## Nice to have

- Until the next full sync, a deleted owner's partner reminder still comes through `GET /reminders/pending` with `cycle_anchor_kind` NULL and falls back to the owner's own text. Filtering pending `cycle_period` rows by an existing anchor would close that window.
- `cycle.js:510`: `parseInt("12abc", 10)` accepts 12; `Number.isInteger(Number(x))` as in `intInRange` is stricter.

## CI and the order with #1177

CI on `5e631a0e` stopped at the fourth suite: `npm test` is one `&&` chain and ends at the migration gap test, and Node 22 was cancelled. So the rest of the chain has not run in CI for this head. The plan stays as announced: #1177 takes 209 and lands first. Then please merge main again (keep 209 and append 210-213 after it; the two heads conflict in `server/db.js` and `test/test-health-structure.js`), check the CHANGELOG placement once more and push, and CI will run the whole chain.

ulsklyc (commented):
Thanks, round two holds at the code. Checked at 59ea34d8.

## Fixed

- The CHANGELOG bullet is back under `[Unreleased]` / `### Added`.
- `PUT /cycle/settings` returns `eligible_partners`. With the old response restored, your save-twice test fails.
- No em or en dash is left in any added line of the branch.
- A day-log save without `feelings` and `mood` leaves a legacy `mood` alone. With the old upsert restored, the "bleibt erhalten" test fails.
- The four stale references are gone.
- `GET /reminders/pending` hides cycle rows whose anchor is gone. Without the filter, your new test fails. The push path already syncs before it reads due reminders, so it had no such gap.
- `notify_partner_user_id` now parses strictly.

health-api, health-cycle, health-structure, cycle-reminders, reminders-routes and changelog all pass on their own.

## Should fix

- **The restored CHANGELOG wording understates the privacy rule.** `CHANGELOG.md:15` now says "and intimacy - the latter is never shown to anyone but yourself". But cervical mucus, both test results and intimacy are all stripped for everyone but the owner (`server/routes/health/cycle.js:314-319`, SPEC "Owner-only axis"). The wording after the merge was right on this point. Please name all four again.
- **SPEC and OpenAPI still describe the old day-log save.** `docs/SPEC.md:3077-3078` says every save replaces the feelings rows and sets `mood` to NULL. The `POST /cycle/logs` description (`server/openapi/paths/health.js:110`) says `mood` is never written again. Neither matches the new rule: a save without both keys keeps feelings and `mood`, while a save without `symptoms` still clears symptoms. Please describe that rule in both places.
- **The app's own dialog still drops a legacy mood.** The day-log form always sends `feelings` (`public/pages/health.js:6315`), so editing such a day in the app clears `mood` (`cycle.js:382`). A value outside the list never showed as a chip, so the user can't see what goes. That behavior is fine with me, but then the CHANGELOG sentence about the stricter validation should say an older free-text mood is dropped the next time that day is saved in the app.

## Nice to have

- `"12abc"` has no test. With `parseInt` restored, `test:health-api` stays green (138/138).
- In the "bleibt erhalten" test, `assert.deepEqual(updated.body.data.feelings, [])` (`test/test-health-api.js:1156`) can't fail, because the day never had feelings. Seed one feeling row first. Right now, a change that clears the rows on such a save stays green.
- The `PUT /cycle/settings` OpenAPI description could mention that the response now carries `eligible_partners` as well.

## After #1177 lands

The plan is unchanged. Merging main after #1177 conflicts in two files:

- **`server/db.js`:** keep 209 and append 210-213 unchanged after it.
- **`test/test-health-structure.js`:** #1177 replaced the fixed route count with `EXPECTED.length` and dropped the number from the title. Keep that form together with your route entries.

Everything else merges on its own. The blank line between the last `[2.66.1]` bullet and `## [2.66.0]` went missing in `92302125`. If it's still missing at merge time, I'll put it back.

mclgoerg:
Round three is in, pushed as `819531d2` + `36c7b910`.

**Should fix**
- The CHANGELOG sentence names all four owner-only fields again (cervical mucus, both test results, intimacy), matching SPEC's "Owner-only axis".
- SPEC and the OpenAPI `POST /cycle/logs` text now describe the actual rule from round two: a save that provides either `feelings` or the legacy `mood` fully replaces the feelings rows and nulls `mood` - the same replace-when-provided semantics `symptoms` already has - while a save that provides neither leaves both untouched.
- Added the clause you asked for: since the app's own day-log form always sends `feelings`, editing any day there still clears an old free-text `mood` on the next save; the preservation only helps a caller that omits both keys entirely.

**Nice to have**
- Added the `"...abc"` -> 400 test for `notify_partner_user_id`, built around a real eligible id so it actually exercises the strict-parse path rather than passing for the wrong reason.
- The "stays untouched" test now seeds a real feeling first and asserts it survives the no-op save, instead of asserting an already-empty list stayed empty.
- PUT `/cycle/settings`'s OpenAPI text now says its response matches GET, including `eligible_partners`.
- The missing blank line between `[2.66.1]` and `[2.66.0]` is back.

Nothing in `server/db.js` or `test/test-health-structure.js` was touched this round, and #1177 still hasn't landed on main as far as I can tell - I'll hold the migration renumbering and the two-file merge exactly as you described until it does.

`test:changelog`, `test:health-api` (139), `test:openapi-structure`, `test:openapi-coverage` and `test:docs-landing` all green; the disclosed 209 gap is still the only thing failing in the full chain.

ulsklyc (commented):
Thanks, round three holds at the code. Checked at 36c7b910.

## Fixed

- The CHANGELOG names all four owner-only fields again (`CHANGELOG.md:15-16`).
- SPEC and the `POST /cycle/logs` description now state the feelings rule the upsert follows (`server/routes/health/cycle.js:355-390`): with `feelings` or `mood` present as a key, a save replaces the rows and nulls `mood`. With neither key, it leaves both alone.
- The CHANGELOG now says the app's form clears an older free-text mood. I checked what that costs. Migration 211 backfills only the seven known values, and `GET /cycle/logs` always returns `feelings` as an array, so neither the dialog nor `normalizeFeelingEntries()` ever falls back to such a value. It never showed in the app, and main's `<select>` dropped it on save the same way. Nothing visible is lost, and the sentence is accurate.
- The `${userB}abc` test fails with `parseInt` restored (138/139).
- The "stays untouched" test fails once a save without either key clears the feelings rows (138/139).
- The `PUT /cycle/settings` description mentions `eligible_partners`, and the blank line before `## [2.66.0]` is back.

health-api (139), openapi-structure, openapi-coverage, changelog and docs-landing pass on their own.

## Should fix

- **`symptoms` doesn't work the way the new text says it does.** `docs/SPEC.md:3079-3080` calls the feelings rule "the same full-replace-when-provided semantics `symptoms` already has". `server/openapi/paths/health.js:110` says "same full-replace semantics as `symptoms`". But `replaceSymptoms()` runs on every save (`cycle.js:389`), and an omitted `symptoms` normalizes to `[]`. I seeded one symptom, then saved the same day again with only `flow`: the response came back with `symptoms: []`. That's the contrast from last round: feelings are kept when both keys are omitted, symptoms are cleared. Please describe that difference instead of claiming they match. While you're there, the line break after "full-replace-when-" renders as "full-replace-when- provided".
- **A day logged only with feelings no longer gets a calendar dot.** `hasLog` (`public/utils/health-cycle.js:1643`) still looks only at `flow`, `symptoms`, `mood` and `note`. On main, a mood-only day got the dot. Now a save never writes `mood`, so `buildCycleCalendar()` returns `hasLog: false` for a day with only `feelings: ['good']`. It does the same for a day with only cervical mucus or only an LH test, and returns `true` for `mood: 'good'`. An older mood-only day loses its dot the first time it's saved in the app. Please add `feelings?.length` and the four new fields (`basal_temp` was already missing on main), with a test next to the empty-symptoms one at `test/test-health-cycle.js:1100`.
- **SPEC still says the backfill copies every non-empty `mood`** (`docs/SPEC.md:3074-3075`). Migration 211 copies only values that are in the seven-key list after `LOWER(TRIM(...))` (`server/db.js:8379-8382`). Anything else stays only in the frozen column. The migration is right, so please change the sentence to match it.

## Nice to have

- `CHANGELOG.md:22` is still a 35-character stub ("shows up everywhere it matters: a") in a paragraph filled to about 100 columns.
- The `GET /cycle/logs` description (`server/openapi/paths/health.js:103`) still says `mood` is "no longer written". A save with `feelings` or `mood` now sets it to NULL.

## After #1177 lands

The plan is unchanged: this merges after #1177. The migration order and the two-file conflict (`server/db.js`, `test/test-health-structure.js`) stay with you, as described last round.

mclgoerg:
Round four is in, pushed as `dc72a164`.

**Should fix**
- `hasLog` now also checks `feelings`, `cervix_mucus`, `lh_test`, `pregnancy_test` and `basal_temp`, not just flow/symptoms/mood/note - a feelings-only or mucus-only day gets its calendar dot back, and so does an old mood-only day the next time it's resaved. Two new tests pin it (feelings-only, mucus-only).
- SPEC and the OpenAPI text no longer claim feelings and symptoms share replace semantics - they don't, and now say so: symptoms are fully replaced on every save (an omitted array clears), feelings/mood are only replaced when the request names one of those keys. Fixed the stray line-wrap space while I was in there.
- SPEC's backfill sentence now matches migration 211 exactly - only values that pass `LOWER(TRIM(mood))` against the seven-key list get copied, anything else stays only in the frozen column.

**Nice to have**
- The orphaned short line in the CHANGELOG paragraph is reflowed to the surrounding column width, no wording changed.
- `GET /cycle/logs`'s description matches `POST`'s now - mood is written (nulled) only when a save's body names `feelings` or `mood`.

`health-cycle` (147), `health-api` (139), `changelog`, `openapi-structure`, `openapi-coverage`, `docs-landing` and `frontend-audit` all green. No changes to `server/db.js`, `test/test-health-structure.js`, or any migration number this round either - still waiting on #1177.

ulsklyc (commented):
Thanks, round four holds at the code. Checked at dc72a164.

## Fixed

- `hasLog` (`public/utils/health-cycle.js:1649-1659`) now counts feelings and the four new fields. Restoring the old expression fails both new tests (145/147), and dropping only `feelings?.length` or only `cervix_mucus` fails the matching test. I held the list against every column of `cycle_day_logs`. Only `intimacy` is left out, and that day already gets its own heart for the owner, while everyone else never receives the field. Empty strings and `basal_temp: 0` can't produce a false dot, because the route stores neither.
- SPEC (`docs/SPEC.md:3073-3087`) and both OpenAPI descriptions (`server/openapi/paths/health.js:103`, `:110`) now describe what `cycle.js:355-390` does: `symptoms` is replaced on every save, while feelings and `mood` are replaced only when the body names one of those keys.
- The backfill sentence matches migration 211 (`server/db.js:8379-8382`).
- The GET description no longer says `mood` is never written.

health-cycle (147), health-api (139), openapi-structure, openapi-coverage, changelog and frontend-audit pass on their own.

## Before merge: new conflict with main

Since last round, main moved the household member predicate into `server/services/household-members.js` (#1227, #1228). `git merge-tree` against main now lists `server/services/member-email.js` as conflicting, but the part that breaks is elsewhere:

- `server/services/cycle-reminders.js:42` imports `isHouseholdMember` from `./member-email.js`. That file merges cleanly, and main no longer exports the function there. If you keep both sides of the conflict, the server stops at load time with `SyntaxError: The requested module './member-email.js' does not provide an export named 'isHouseholdMember'`.
- `listHouseholdMembers()` (`member-email.js:108`) still uses `HOUSEHOLD_MEMBER_SQL`, which main removed. Main also added `test:household-member-guard`, which fails on any `FROM users ... .all()` list without `householdMemberSql()`.

I tried one resolution on the merged tree: `${householdMemberSql('u')}` in `listHouseholdMembers()` (main's `member-email.js` already imports it), plus `import { isHouseholdMember } from './household-members.js'` in `cycle-reminders.js`. With that, the module loads, the guard passes (47/47) and health-api passes (139). This is separate from #1177 and can be done in the same merge.

## Nice to have

- Two changes to `hasLog` still leave every test green. Dropping `lh_test`, `pregnancy_test` and `basal_temp` together is one. Changing `log.feelings?.length` to `log.feelings` is the other, and it matters more: `GET /cycle/logs` returns `feelings: []` on every row, so every day saved empty would get a dot. Adding `feelings: []` to the empty-symptoms fixture (`test/test-health-cycle.js:1100`), plus one row per remaining field, would cover both. The mucus fixture at `:1137` uses `'egg_white'`, but the value is `eggwhite`.
- The reflow split a code span across `CHANGELOG.md:16-17` ("On `POST" / "/api/v1/health/cycle/logs`,"). GitHub renders it fine, but the in-app changelog strips backticks line by line (`server/routes/changelog.js:49`, `:95`), so after release the "What's new" view would show two stray backticks. Keeping `POST /api/v1/health/cycle/logs` on one line avoids that.

## After #1177 lands

Unchanged: this merges after #1177, and the migration order plus the `server/db.js` / `test/test-health-structure.js` conflict stay with you, as described before.

mclgoerg:
Round five is in, pushed as `040c0f9f`.

**Nice to have**
- Six new tests close both gaps: an empty-feelings-array day now has its own test asserting `hasLog: false` (the truthy-array bug you described would fail exactly this one), and `mood`, `lh_test`, `pregnancy_test` and `basal_temp` each get their own isolated day. Counter-proved by removing one field at a time from the condition - each removal broke exactly its own test and nothing else.
- The mucus fixture is `eggwhite` now, matching the real enum.
- The changelog paragraph is rewrapped so `` `POST /api/v1/health/cycle/logs` `` sits on one physical line - no wording changed.

`health-cycle` (152), `changelog` and `docs-landing` all green, dash count still zero.

Noted and not touched yet: the household-member-predicate conflict from #1227/#1228, together with your validated resolution (`householdMemberSql('u')` in `listHouseholdMembers()`, importing `isHouseholdMember` from the new `household-members.js` in `cycle-reminders.js`). I'll apply that in the same pass as the #1177 migration merge once #1177 lands - still unmerged as of this check.

ulsklyc:
#1177 is merged (546eb3ee), so migration 209 is taken on main. As planned, please renumber yours to 210-213 and merge main. Besides the `server/db.js` / `test/test-health-structure.js` conflict, the `member-email.js` / `cycle-reminders.js` resolution from my last review applies. I'll look at 040c0f9f together with that update.

mclgoerg:
Merged main and renumbered, pushed as `b010f23f`.

Migrations stayed at 210-213, appended after 209 as planned. Both conflicts resolved as you described:
- `server/db.js`: kept 209 (fasting) and appended 210-213 after it - one wrinkle worth mentioning: git's own merge had silently dropped the `` `, },\n];`` closing between migration 209 and 210 (the two sides' trailing boilerplate was textually identical, so it collapsed the conflict region and left 209's template literal unterminated). Caught it immediately via `node --check`, added the missing closing back.
- `server/services/member-email.js` / `cycle-reminders.js`: removed the now-duplicate `isHouseholdMember` from `member-email.js`, `listHouseholdMembers()` uses `householdMemberSql('u')`, and `cycle-reminders.js` imports `isHouseholdMember` from `household-members.js` instead.
- `test/test-health-structure.js`: kept your dynamic `EXPECTED.length` form; our `POST /cycle/import` entry merged into the list cleanly on its own.

`test:migrations-append-only` and `test:household-member-guard` are green now (47/47 on the latter, matching what you validated). Full `npm test` chain: exit 0, 296 suites, 0 failures - the disclosed gap is gone.

ulsklyc (approved):
Round 6 on `b010f23f`. Approving. I checked the merge itself rather than its description, and re-read the three round-4 items against the code.

**The merge.** I isolated your hand edits by diffing your merge against the mechanical one (`git merge-tree --write-tree 963c672d 040c0f9f`, then diffing that tree against `b010f23f`): 4 files, 9 insertions. `server/services/cycle-reminders.js:42-43` resolves exactly as discussed, with `isHouseholdMember` from `household-members.js` and `listHouseholdMembers` from `member-email.js`, and `member-email.js:92-103` is on `householdMemberSql('u')`. An ESM link probe imports the server cleanly and migrates a fresh database to v213. The `test` chain in `package.json` is byte-identical to main's: main's six new suites are all present, and you added no new suite name.

**Migrations.** main ends at 209; 210 to 213 are appended, gapless and unique, with nothing reordered. Because the conflict boundary sat inside migration 209's `up` template, I byte-compared 205 to 209 across the merge rather than eyeballing them: identical, so the fasting journal SQL was not truncated.

**Round 4.** All three are fixed, and I verified the prose against the code instead of trusting it. `symptoms` really is unconditional (`server/routes/health/cycle.js:389`, and `normalizeSymptomEntries(undefined)` returns `[]`), `feelings` and `mood` really are `hasOwnProperty`-gated (`:355-356`, `:382`), and the SPEC backfill paragraph matches migration 211's SQL. `hasLog` covers feelings through `.length`, which avoids the truthy-empty-array trap.

One thing from `040c0f9f` worth naming: the CHANGELOG reflow is a real fix, not cosmetic. The in-app parser strips backticks per line, so the code span you split across two lines rendered broken in the app while `test:changelog` stayed green.

22 suites pass on the merged tree, including the main-side ones the merge touched (`health-fasting`, `settings-admin-gate`, `permissions`, `people-pickers`).

### Nice to have, neither blocking the merge

1. `public/utils/health-cycle.js:1649` - `hasLog` enumerates nine day-log fields but not `intimacy`. This is not a missing-marker bug: an intimacy-only day still gets its own heart marker (`public/pages/health.js:5009`, own view only), so the round-4 regression class is closed. But this condition is now the single place that answers "what counts as a logged day", `intimacy` is the only day-log scalar absent from it, and your six new tests pin every other field in isolation while intimacy-only stays unpinned.
2. `server/routes/reminders.js:148` and `server/services/notifications.js:507` - the same `SELECT u.display_name FROM cycle_reminder_anchors a JOIN users u ...` lives twice under the same name `cycleOwnerName`, with different signatures. The comment at `reminders.js:146` says "keine zweite Wahrheit" while being exactly a second copy.

Thanks for the six rounds, and for taking the renumbering cleanly.

PR #1197, opened 2026-09-14
docs(unraid): list waste collection in the store overview

ulsklyc:
**Draft on purpose: merge only after v2.67.0 is tagged.** Unraid Community Applications reads `ca_profile.xml` from `main`, so a merge publishes this text in the store at once, while the Waste collection module (#1063) is still under `[Unreleased]` and reaches households only with the next release.

## Summary

The repository overview in `ca_profile.xml` lists the modules by name. The optional Waste collection module is missing from that list. This adds it next to housekeeping.

It was taken out of today's documentation sweep (01923340) for exactly the timing reason above.

No behaviour change, no CHANGELOG entry (the module's own entry covers it).

## Before merging

- `git tag --contains 8b298415` names the release that carries #1063.
- Then mark ready for review and merge.


ulsklyc:
Closing: this landed on main with the v2.67.0 documentation sync (`fabab86b`). `ca_profile.xml` already lists waste collection in the app overview, and the diff here is character-identical to what went in, so there is nothing left to merge.

Issue #1164, opened 2026-09-12
Period navigation: one position and one visibility rule for the "Today"/"Current" reset (follow-up to #1148)


mclgoerg:
Follow-up to the review of #1148 (https://github.com/ulsklyc/yuvomi/pull/1148#pullrequestreview — "CHANGES_REQUESTED"), where the "Today"/"Current" reposition was split out of the PR so the position question can be decided once, in one place. The question asked there:

> is there a module where the *current* arrangement actually reads badly, or is this purely about the inconsistency between the three? The first would move me; the second is real but cheaper to fix in the other direction.

## The honest answer: mostly pure inconsistency

I re-checked all three modules in a running instance (seeded demo data, light theme) at 390 / 768 / 1024 / 1440 px, plus the `docs/screenshots/` set. No module is confusing or broken. Two of the three read fine at every width:

- **Calendar** — "Today" left of the prev/next arrows (`calendar.js`, `#cal-today`). Reads fine.
- **Budget** — "Current" right of the arrows, inside the nav group but behind the stepper (`budget.js`, `#budget-today`), exactly as its comment records: *„Aktuell" ist ein Reset, kein Navigationsschritt: hinter dem Stepper statt zwischen Pfeil und Wert.* Reads fine at all four widths; the pill sits between the stepper and "+ Entry" and never reads as a third step.

So by the framing above: this is the second case, and the cheaper fix is moving Calendar (one button, one module) to match the recorded Budget rule — not re-landing the reverted diff, which moved two modules to match Calendar.

## The one caveat: Meals is the weakest of the three, mildly

Meals is the only module whose reset is not adjacent to its stepper at all. "Today" (`meals.js`, `#week-today`) lives in the `page-toolbar__actions` slot:

- At 1440 px the week range and its arrows sit far left; "Today" sits far right, grouped with the recipe-rail toggle, "Randomize plan" and "+ Meal" — content actions of a very different kind ("Randomize plan" rewrites up to 28 slots; "Today" just navigates). The reset is ~700 px from the range it resets.
- Below ~768 px the toolbar wraps (`page-toolbar--wrap`) and "Today" lands on a second row next to "Randomize plan", detached from the stepper row entirely.

[...]

PR #1200, opened 2026-09-14
fix(ui): one position and one visibility rule for the period-nav reset (#1164)

mclgoerg:
Implements #1164 as decided - one PR for all three modules, since it is one rule. (Deliberately "implements", not a closing keyword, in case the issue should stay open until release.)

## Position - the rule Budget already records

> „Aktuell" ist ein Reset, kein Navigationsschritt: hinter dem Stepper statt zwischen Pfeil und Wert.

- **Calendar:** `#cal-today` moved from before the prev/next arrows to after them.
- **Meals:** `#week-today` left the `page-toolbar__actions` slot and moved into `week-nav`, behind the arrows. It no longer sits next to "Randomize plan", and it stays on the stepper row when the toolbar wraps below ~768 px.
- **Budget:** position untouched.

The moved markup carries a comment in the same spirit as Budget's, referencing #1164.

## Visibility - Calendar's rule, applied to all three

`syncTodayButton()` in calendar.js stays the model: the reset is `hidden` while the current period is displayed and keeps its slot, so the header height never jumps - and a screen reader never hears a control that has no effect. Meals hides on the current week; Budget hides on the current month (and, on the reports tab, on a today anchor; a tab without month navigation keeps the reset hidden regardless, the existing TAB_CAPS block keeps the last word there). The mechanism is exactly Calendar's: the `hidden` attribute, enforced by the shared `.btn[hidden] { display: none !important; }` rule in layout.css - the same rule that already makes Calendar's `hidden` effective. No new strings, no new CSS; Budget's label stays "Current" (it jumps to the current month, as re-affirmed in the #1148 review).

## The two requirements from the decision

**1. Accessible name and shortcuts follow the move; nothing selects by position.**
- Each button's accessible name is its visible text (`calendar.today` / `meals.today` / `budget.currentMonth`) and travels with the element; DOM order now matches visual order, so sequential focus order does too.
- No keyboard shortcut targets any of the three buttons: `n` targets `.page-fab`, `f` targets `#cal-search` (router.js `SHORTCUTS`); every today/current handler binds by id.
- Grepped for positional selection (`:first-child`/`:last-child`/`:nth-*`, sibling combinators in CSS; `firstElementChild`/`children[...]`/`nextElementSibling`/`previousElementSibling` in JS) scoped to `cal-toolbar`/`week-nav`/`budget-nav`/`page-toolbar__center` and the three ids, across public/styles, the three pages and router.js: 0 real matches (the single grep hit is a prose comment in kitchen-tabs.css).

**2. A behavioral test in each of the three suites, not source-text matching.**
Each page renders its period nav through a small builder used by the real render path (`periodNavHtml()` / `weekNavHtml()` / `monthNavHtml()`, exported via the page's `__test`). Each suite renders it and asserts (a) the element order - previous, value, next, reset - from the rendered markup, and (b) the hidden state by driving the real sync function (`syncTodayButton()` / `syncCurrentButton()`) against the current and a non-current period. test-budget-ui.js now runs under the browser loader so budget.js is really imported (same minimal DOM stub as test-waste-ui.js/test-shopping-ux.js; the `/rrule-ui.js` loader stub gained the `intervalUnitLabel` export budget.js needs).

Counter-proofs (change reverted, suite run, change restored, suite green again):

| Reverted | Result | Restored |
|---|---|---|
| Calendar "Today" moved back before the arrows | test:calendar 153 passed / 1 failed - exactly the new test | 154 / 0 |
| Meals "Today" moved back into the actions slot | test:meals 48 / 1 - exactly the new test | 49 / 0 |
| Meals hidden logic removed (`btn.hidden = false`) | test:meals 48 / 1 | 49 / 0 |
| Budget refinement removed (back to `!caps.month` only) | test:budget-ui 72 passed / 1 failed - exactly the new test | 73 / 0 |

## Verification

test:calendar (154 + 27), test:calendar-structure (5), test:calendar-defaults (7), test:meals (49), test:meals-routes (57), test:budget-ui (73), test:budget-structure (5), test:frontend-audit (386), test:mobile-scroll-layout (9), test:i18n (54), test:changelog (34), test:suite-chain (4), test:browser-loader-stubs (3) - all green against the committed HEAD, plus one full `npm test` (exit 0). CHANGELOG entry under `[Unreleased]` / `### Changed`; docs/test-suites.md updated for the three suites.

Timing as set in the issue: interface work, rides the next interface release - no urgency from this side.


ulsklyc (changes requested):
Thanks for building this so carefully. The order is right, every path that changes the period reaches the sync in all three pages, "today" comes from `todayKey()` throughout, and the two counter-proofs missing from your table hold too: Calendar with `btn.hidden = false` and Budget with the reset moved before "‹" both turn their suites red. One thing does not hold in the browser, though, and it is the premise of the decision itself.

## Blocking

- **Hiding the reset moves "›" under the pointer, so a second click jumps back.** `hidden` resolves to `.btn[hidden] { display: none !important }` (`public/styles/layout.css:3417`), which removes the box, and the label is the flexible item (`.budget-nav__label { flex: 1 }` in `budget.css`, `.cal-toolbar__label { flex: 1 1 auto }` in `calendar.css`). On the current period "›" sits at the end of the row; one click later the reset appears behind it, pushes it left by the reset's width and takes exactly its place. Clicking "›" twice without moving the pointer goes to the next period and straight back. Measured with a layout probe on a copy of this head and of main, 33 layouts each (320/375/390 touch, 640/768/1024/1280 pointer, de, plus fr and uk on phones): the second click lands on the reset in Budget 11/11, Calendar 7/11 and Meals 5/11, and on main 0/33. Example, Budget at 320 px in German: "›" covers x 252-300 on September, after one click the reset covers x 228-300.

  The decision asked for the button to keep its slot, and that is what is missing: nothing reserves the space. Whatever the mechanism, "›" must not move when the reset appears or disappears. Keeping the box and taking it out of the pointer, focus and accessibility paths (for example `visibility: hidden` together with `inert`, via a class instead of `hidden`) is the direct way. Calendar below 768 px currently hands exactly those 66 px to the label on purpose (`calendar.js:1880-1885`), so that spot needs a look.

- **Meals on phones: the week label runs under "›" and "Heute".** With `#week-today` inside `.week-nav` the row carries 66-102 px more, and `.week-nav__label` is `white-space: nowrap; min-width: 0` without clipping (`public/styles/meals.css:66-80`). Overflow of the label on a non-current week: 76 px at 320 in German, 112 px in French, 97 px in Ukrainian, and 21/57/42 px at 375. On main the worst case is 9 px at 320. Staying on the stepper row only helps if the row fits, so phones need their own answer, for example a shorter date format at that width. Screenshots at 320 and 375 in de and fr on a non-current week would make it easy to judge.

## Should fix

- **Keyboard focus drops to `<body>`.** Enter on the reset loads, the sync hides the focused button, and nothing moves focus (`meals.js:761-767`, `budget.js:584-596`). Calendar already did this on main (`goToday()`, `calendar.js:1930`), but Meals and Budget kept focus on their still-visible button, so for those two it is new. If the reset had focus, move it to a stepper control before hiding; taking Calendar along would be welcome.
- **Budget Reports uses a different rule than Calendar.** On Reports, `syncCurrentButton()` hides only when `state.reportAnchor === todayKey()` (`budget.js:462-464`). Calendar hides when the shown range contains today (`calendar.js:1896`). Entering Reports from a past month anchors on the 1st (`anchorForMonth()`), so stepping forward to the current month shows "Current" while that month is on screen, and clicking it changes nothing visible, which is the case the decision wanted to avoid. The same goes for the week and year ranges. Please compare containment per range; the new test pins the anchor equality and needs to follow.
- **The reset flashes during the first load.** `weekNavHtml()` and `monthNavHtml()` render it without `hidden`, and the sync only runs after the first data load. Both pages always open on the current period, so the button shows while `/meals` and `/budget` load and then disappears, moving "›" once more. Rendering it hidden, or syncing right after the markup insert, avoids that.
- **"The header height never jumps" in the CHANGELOG does not hold.** Budget switches between 170 and 174 px at 640 and 768 px (`.btn` carries a larger `min-height` than `.btn--icon`), where main stays at 174. Calendar's jumps at 640 and at 375/390 exist on main already, but the new sentence promises the opposite. Please say what holds after the fix.

## Nice to have

- **The tests stop before the render path.** They render the builder string and run the sync against a stub root, so removing the call from the render path stays green: I deleted `syncTodayButton();` in `renderWeekGrid()` (`meals.js:383`) and `syncCurrentButton();` (`budget.js:983`), and `test:meals` and `test:budget-ui` stayed green. The decision's wording is met; a render through the page's own entry point would bind the wiring as well. None of the tests can see the layout findings above, so the browser checks carry those.
- `test/test-meals.js:820` computes the expected week with the same `getMondayOf(todayKey())` as the page, so `toLocalDateKey()` in `meals.js:244` would still pass unless the test pins a household zone whose day differs from the process zone.
- `public/styles/meals.css:236-246`: the comment still puts "Heute" below the stepper row.

No hard-constraint issues: no `innerHTML`, no new strings, no dash characters, `package.json` only adds the loader flag to `test:budget-ui`, and the CHANGELOG entry merges into one `### Changed` section with main. Timing as agreed: the next interface release after the merge.

ulsklyc (changes requested):
Thanks for the careful follow-up. I re-ran the layout probe against this head and against main: 60 layouts each (Calendar, Meals, Budget and Budget Reports at 320/375/640/768/1280 in de, fr and uk). Most of the last round is fixed and measured:

- The slot holds. A second click at the old "›" position never lands on the reset (0/60), "›" is back in place after using the reset, and Tab skips the reset only while it is inert.
- Focus goes to "‹" after Enter on the reset in all three modules (0/60 on body; main Calendar was 15/15 on body).
- Reports compare containment: entering Reports from August and stepping to September hides the reset, and week, month and year all behave, including with a delayed stats response.
- No visible reset frames during load (0/60). With the pre-render sync removed, the same probe sees 4 and 2 frames, so the check is not blind.
- Header height no longer changes when paging, except Budget uk/320, which is identical on main. The CHANGELOG sentence holds now.

## Blocking

- **Meals at 320 px in French and Ukrainian: the week label still runs under "›".** The reserved slot is as wide as the translated word (65.7 px "Heute", 87.2 px uk, 101.5 px "Aujourd'hui"). In French that leaves the label 78.5 px for 112.9 px of text: 34 px overflow, 30 px of it under "›"; Ukrainian 20 px, 16 px under "›". German now fits, and main's worst case was 9 px. `.week-nav__label` is `nowrap` without clipping (`meals.css:66-80`). Whatever the approach (a narrower reset at that width, or a label that clips), "›" should keep its x and no text should sit under it at 320 in fr and uk.

## Nice to have

- **The wiring tests read source text, they do not render.** Removing the calls turns them red, but `// syncTodayButton();` and `if (false) syncTodayButton();` (Meals and Budget) stay green.
- **Nothing in the suites holds the layout mechanism.** With `.btn.is-current { display: none }`, calendar, meals, budget-ui, frontend-audit and mobile-scroll-layout all stay green, which brings back last round's blocker unseen. The same goes for switching the narrow week format off. A small CSS guard would pin it.
- **The zone test checks `todayKey()` itself, not `syncTodayButton()`.** Replacing `todayKey()` in `meals.js:277` with `new Date().toISOString().slice(0, 10)` keeps `test:meals` green.
- **`formatWeekLabel()` reads `matchMedia` only at render.** Rotating across 640 px keeps the old format until the next week change.
- **Stale text:**
  - `calendar.js:1891` names `moveFocusOffCurrentButton`, which does not exist.
  - `layout.css:2258-2269` says the Calendar label now sits at its 7ch floor below 768 px, but it was never truncated in any measured layout.
  - `docs/test-suites.md:65`, `:96` and `:117` still describe `hidden`.
- **Dashes:** the comments at `meals.js:79` and `:99` use an en dash; please use "-". The rendered separator at `:103` matches main and can stay.
- **CHANGELOG:** the sentence about the first cut using `hidden` describes review history rather than the change and could go.

The branch now conflicts with main in `CHANGELOG.md` only: both sides add the first bullet under `[Unreleased]`, `### Changed`, so keeping both resolves it.

mclgoerg:
Round 3 findings addressed. Pushed `a096445e` (merge of `31b65efe` with `upstream/main`), CI green on Node 22.x/24.x: https://github.com/ulsklyc/yuvomi/actions/runs/34963874821

## Blocking - Meals overflow at 320px in fr/uk

Root cause confirmed against a real rendered page (server started via `test/server-ready.js` against a fresh DB, driven with Puppeteer at 320px viewport, locale switched through `localStorage['yuvomi-locale']`): the reset button's own reserved width scales with its translated word ("Heute" 68.1px / "Aujourd'hui" 105.7px / "Сьогодні" 91.25px in this environment's font), and `.week-nav__label` (`meals.css:66-80`) had `white-space: nowrap` with no clipping - text painted straight over the "›" arrow whenever the label didn't fit.

Fix: `.week-nav__label` now also carries `overflow: hidden; text-overflow: ellipsis;`. This is a pure paint-level change - the box's own width was already whatever the flex layout gave it (that never moved), so "›" keeps its x-position exactly as before. Measured before/after (real browser, 320px, en-dash-normalized label text is `14.09. – 20.09.` in this test fixture's week since the underlying date formatter doesn't vary by locale here):

| locale | label available width | before: chars painted at/past the arrow's left edge | after |
|---|---|---|---|
| de | 97px | 4 | 0 (clipped) |
| fr | 59px | 8 | 0 (clipped) |
| uk | 74px | 6 | 0 (clipped) |

Screenshots at 320px confirm no glyph renders under or past "›" in any of the three locales after the fix, and `nextLeft`/`nextRight` (the arrow's own rect) are byte-identical before and after in all three locales, so the round-1 slot-shift bug is not reintroduced.

(The exact px figures differ from the review's own probe because font rendering differs headless-vs-real-browser-profile; the reproduction and the fix's effect - overlap present before, gone after, arrow static throughout - is what was verified here.)

## Nice-to-have (all six)

1. **Wiring tests now render, not read source.** `renderWeekGrid()`/`updateTabs()` gained test-only wrappers (`renderWeekGridForTest()`, `updateTabsForTest()`) that swap in a test container and call the unmodified function. The rewritten tests in `test-meals.js`/`test-budget-ui.js` assert the real `.is-current` result on a real (minimal-DOM) `#week-today`/`#budget-today` node. Counter-proved: commenting out `syncTodayButton();` in `renderWeekGrid()` -> 52/1 (was 53/0); `if (false) syncCurrentButton();` in `updateTabs()` -> assertion failure (was 76/0). Both restored to green after.

2. **CSS mechanism guard added.** New test in `test-frontend-audit.js` reads the actual `.btn.is-current` rule from `layout.css` via `eachRule()` and asserts `visibility: hidden` is present and `display: none` is not. Counter-proved: swapping to `display: none` -> red (was 387/0, became 386/1); restored to green. The narrow-week-format switch also got a real behavioral guard (item 2 below, via a spy on `window.matchMedia`) rather than a source-text guard, since a source-text check for a ternary that references the same variable either way wouldn't have been meaningfully different from the original problem.

3. **Household-timezone test now goes through `syncTodayButton()`.** Rewritten to set `state.currentWeek` to the LA-correct Monday and the UTC-process Monday and assert `.is-current` on both, instead of calling `todayKey()` directly. Counter-proved with the reviewer's exact mutation (`todayKey()` -> `new Date().toISOString().slice(0, 10)` at `meals.js:277`): 53/1 (was 54/0); restored to green.

4. **`formatWeekLabel()` re-evaluates on breakpoint crossing.** Added a held `MediaQueryList` (`(max-width: 639px)`, same pattern as `router.js`'s `darkSchemeQuery` - a bare `matchMedia(...).addEventListener(...)` would let the engine collect the list and the listener would go silent) with a `change` listener that re-runs `renderWeekGrid()` when the container is still connected. New behavioral test calls the exported handler directly and confirms it redraws when connected, and does nothing when not (post-navigation-away). Counter-proved: gutting the handler body -> red; restored to green.

5. **Stale text fixed:**
   - `calendar.js:1891` no longer names a nonexistent `moveFocusOffCurrentButton` - corrected to describe the inline logic in `syncTodayButton()` itself.
   - `layout.css:2258-2269`'s claim that the label sits at its 7ch floor is corrected: it now says the label permanently lost 66px of headroom, and that the 7ch floor is a safety net that doesn't actually engage in the measured layouts (round 3 review).
   - `docs/test-suites.md:65,96,117` updated from `hidden` to `.is-current` + `inert`.

6. **En dashes fixed.** `meals.js:79,99` (comments) now use `-`. Line 103's rendered separator (`${from} – ${to}`) is untouched per the review.

7. **CHANGELOG:** removed the "(a first cut used `hidden`...)" aside describing review history rather than the shipped change.

## Merge conflict

Fetched `upstream/main` (now at `3a53b3a2`) and merged; the only conflict was the expected `CHANGELOG.md` append-only collision under `## [Unreleased]` / `### Changed` - kept both bullets (#1164 reset-position/visibility entry and #1212 README entry), neither moved into a released section.

## Verification

All of `test:calendar`, `test:calendar-structure`, `test:calendar-defaults`, `test:meals` (55/0), `test:meals-routes`, `test:budget-ui` (76/0), `test:budget-structure`, `test:frontend-audit` (387/0), `test:mobile-scroll-layout`, `test:i18n`, `test:changelog`, `test:suite-chain` green locally against the merge commit, plus a full `npm test` (exit 0). Pushed SHA `a096445e4756a2a713a88ed6ccd2463c893a1862`, GitHub CI green on Node 22.x and 24.x: https://github.com/ulsklyc/yuvomi/actions/runs/34963874821

ulsklyc (commented):
Thanks for the quick turnaround and the detailed write-up. I re-ran the same probe against `a096445e` and against main: 60 layouts each (Calendar, Meals, Budget and Budget Reports at 320/375/640/768/1280 in de, fr and uk).

- The round-3 blocker is fixed. With the label clipped, no glyph paints under "›" in any layout (0/60; main 3/60, Meals at 320 in every locale, 5 px). "›" sits at the same x as on the previous head, does not move after the reset (0/60), and a second click at its old position never lands on the reset (0/60).
- The rest of round 3 still holds: focus goes to "‹" after Enter on the reset (0/60 on body; main Calendar 15/15), no visible reset frames during load (0/60), Tab skips the reset only while it is inert (60/60), Reports hide it in the current month and year, and header height only changes for Budget and Reports at uk/320 (-8 px), same as main.
- The test changes hold up. Each of these mutations now turns its suite red: `// syncTodayButton();` and `if (false) syncTodayButton();` in `renderWeekGrid()` (meals 53/2), the same two for `syncCurrentButton()` in `updateTabs()` (budget-ui 75/1), `.btn.is-current { display: none }` (frontend-audit 386/1), `todayKey()` replaced by `new Date().toISOString().slice(0, 10)` at `meals.js:302` (meals 54/1), and an empty `onNarrowWeekLabelQueryChange()` (meals 54/1).
- The listener is registered once: after three visits to /meals a rotation re-renders the grid once, and rotating on /calendar afterwards does nothing and logs no errors.
- meals, budget-ui, calendar, frontend-audit, mobile-scroll-layout, i18n and changelog are green on this head, and the branch still merges cleanly into current main.

## Should fix

- **Meals at 320 px: the week label loses its end date in French and Ukrainian.** The reset's slot is as wide as the translated word (101.5 px fr, 87.2 px uk), so the label gets 78.5 px and 92.8 px for 112.9 px of text. French now shows `14.09.`, the separator and an ellipsis; Ukrainian the same plus a lone `2`. It looks the same on the current week, where the slot next to the cut label is empty. Main shows the whole range, both dates with the year, at this width (9 px overflow). German fits. Round 3 listed clipping as one option, so that part is on me: measured, it costs the end date. A shorter date form, a narrower reset at that width, or letting the reset wrap would keep both ends of the week readable at 320 in fr and uk.
- **The breakpoint handler re-renders the whole week grid, not just the label.** `onNarrowWeekLabelQueryChange()` (`meals.js:125-127`) calls `renderWeekGrid()`, which replaces every card, re-runs the stagger and the scroll to today's column (`meals.js:575-577`). Rotating a 375x812 phone crosses 640 px: with keyboard focus on a meal card, focus drops to `<body>` (main keeps it on the card). With the meal dialog open it is fine: the dialog stays open and Escape returns focus to a card. Updating only `#week-label` in the handler would avoid the focus loss.

## Nice to have

- **Two parts of this round are not held by a test.** Removing `overflow: hidden; text-overflow: ellipsis` from `.week-nav__label` (`meals.css:93-94`) keeps meals, frontend-audit and mobile-scroll-layout green, so the blocker could come back unseen; it would fit next to the new `.btn.is-current` guard. Deleting the `addEventListener('change', ...)` line (`meals.js:129`) keeps `test:meals` green as well, because the test calls the exported handler directly.
- **Stale text:** `docs/test-suites.md:117` still says the Budget reset hides on today's report anchor, but `reportShowsToday()` (`budget.js:475-490`) checks containment since round 3.

mclgoerg:
Pushed `1b8b0d8c`, addressing everything.

## Should fix

**1. Meals overflow at 320px in fr/uk - the end date silently disappeared.** Confirmed the mechanism: the reset button's reserved slot scales with the translated word ("Aujourd'hui"/"Сьогодні" vs "Heute"), and in this environment's font that left the label as little as ~59px (fr) / ~74px (uk) at 320px against a ~133px full range - the round-3 ellipsis clip (correctly) hid the overflow instead of painting it under "›", but that meant the second date just vanished.

Went with the reviewer's third option (letting the reset wrap) rather than a narrower date format or a narrower button: below 640px, `.week-nav__today` now gets `flex-basis: 100%` and its own row under the arrows/label (`meals.css`). The slot stays `.is-current`/`visibility`-driven exactly as before, so it's always reserved whether the button is shown or not - only its position moved from "beside the arrows" to "its own line". This removes the width contention entirely instead of trimming it: the label now gets the full row minus the two 48px arrows, in every locale, regardless of month/week boundaries.

Verified with a real headless Chromium against the seeded app (server via a Puppeteer harness in the same shape as `test/document-guards-harness.js`, locale via `localStorage['yuvomi-locale']`), before/after:

| | before (this branch's prior head) | after |
|---|---|---|
| de/fr/uk @ 320px, label `scrollWidth` vs `clientWidth` | fr: 133 vs 59 (clipped), uk: 133 vs 74 (clipped), de: 133 vs 97 (clipped) | 133 vs 133 in all three (no clip) |
| de/fr/uk @ 375px | fr: 133 vs 114 (still clipped) | 133 vs 133 |
| `#week-next` x-position, current vs. non-current week, same width | varied per locale (depended on the reset's own text width) | identical per width, and identical across de/fr/uk (arrow position no longer depends on the reset's content at all) |
| `#week-next` x-position @ 320/375/639 vs @ 640/768/1280 (year-bearing format, reset back on the arrows' row) | - | unchanged from the pre-existing behavior (only the sub-640px range is affected) |

Also checked Tab order at 320px didn't regress: from `#week-prev`, Tab lands on `#week-next` on the current week (reset `inert`, correctly skipped) and reaches `#week-today` next once a non-current week makes it interactive again - same as before this fix, since DOM order is untouched.

**2. `onNarrowWeekLabelQueryChange()` rebuilt the whole grid for a label-only change.** It called `renderWeekGrid()` on every breakpoint crossing, which rebuilds every meal card, reruns the stagger and re-scrolls to today's column - a phone rotation with a card focused dropped focus to `<body>`. Extracted `updateWeekLabel()` (just `#week-label` + `syncTodayButton()`) and pointed the handler at that instead.

New test drives the real handler with a spy on the grid's own mutation methods (`removeAttribute`/`setAttribute`/`replaceChildren`) and a stand-in focused element: after `onNarrowWeekLabelQueryChange()`, the grid spy sees zero calls and `document.activeElement` is untouched, while the label/reset still update (`.is-current` flips as expected). Counter-proved: reverting the handler back to `renderWeekGrid()` turns exactly this test red (`test:meals` 56/1); restored, 57/0.

## Nice to have (both)

**3. Test guards for round-3's untested changes.**
- `.week-nav__label`'s `overflow: hidden; text-overflow: ellipsis;` now has a `test-frontend-audit.js` guard reading the real rule via `eachRule()`, same pattern as the existing `.btn.is-current` guard next to it. Counter-proved: removing both properties turns it red (`test:frontend-audit` 387/1); restored, 388/0.
- `meals.js`'s `addEventListener('change', ...)` wiring now has a `test-meals.js` guard that imports a fresh module instance (cache-busted specifier, same pattern as `test-nav-badges.js`/`test-overlay-history.js`) against a `window.matchMedia` whose `addEventListener` is itself a spy - only a real top-level `addEventListener('change', onNarrowWeekLabelQueryChange)` call produces a hit. Counter-proved: commenting out the line turns it red (`test:meals` 56/1, distinct test than #2 above); restored, 57/0.

**4. Stale doc fixed.** `docs/test-suites.md:117` described Budget's reset as hiding on "today's report anchor" (anchor-equality); updated to describe the range-containment check `reportShowsToday()` has actually used since round 3 (year/month bounds, or `reportRangeFrom`/`reportRangeTo` for the week view).

## Verification

`test:calendar`, `test:calendar-structure`, `test:calendar-defaults`, `test:meals` (57/0), `test:meals-routes`, `test:budget-ui`, `test:budget-structure`, `test:frontend-audit` (388/0), `test:mobile-scroll-layout`, `test:i18n`, `test:changelog`, `test:suite-chain`: all green against the committed head. Full `npm test`: exit 0.

Pushed SHA `1b8b0d8c5e13944e9d380f95930927dd5d52efc9`. GitHub CI green on Node 22.x and 24.x: https://github.com/ulsklyc/yuvomi/actions/runs/34971551288

ulsklyc (changes requested):
Thanks for the push and the write-up. I re-measured `1b8b0d8c` against main in a real browser (Meals at 320/360/375/414/600/639/640/768 in de, fr and uk, current and next week; Calendar and Budget at 320 and 375), ran mutations against the new tests, and ran the suites on the merge with current main.

- Both round-4 should-fix items are fixed. At 320 px the Meals week label now fits in every locale (113 of 113 px, no ellipsis, 24/24 layouts), so both dates are readable again in fr and uk.
- The reset's own row does not bring back a jump: header height and the week grid's top edge are identical on the current and the next week in all 24 layouts. "›" moves by 2.9 px between weeks because the label text gets narrower, the same 3 px as on main.
- Rotating across 640 px keeps keyboard focus on the focused meal card (375 to 700 to 375 to 700), and the label switches format each time. After leaving /meals through the app and crossing the breakpoint three times there are no errors.
- Reverting the handler to `renderWeekGrid()` turns `test:meals` red (56/1), so does commenting out the `addEventListener` line (56/1), and removing the ellipsis declarations turns `test:frontend-audit` red (387/1).
- `docs/test-suites.md:117` now matches `reportShowsToday()`.
- meals, budget-ui, calendar, frontend-audit, mobile-scroll-layout, i18n and changelog are green on this head.

## Blocking

- **The branch no longer merges into main without breaking Budget.** GitHub only reports the CHANGELOG conflict, but main gained `export const __test = { responsiblePickerHtml, responsibleFilterName, toggleResponsibleFilter, responsibleFilterLabel }` in `public/pages/budget.js` today (#1228, line 2147), and this branch adds its own `export const __test` at the end of the same file (line 3546). The hunks don't overlap, so git merges them silently, and the merged `budget.js` fails to parse: `SyntaxError: Identifier '__test' has already been declared`. On the merged tree the Budget page cannot load, `test:budget-ui` fails at import (0/1), main's `test:people-pickers` fails the same way (0/1, 15/0 on main), and `test:frontend-audit` goes red on its ESM parse guard (392/1). Merging current main and folding both key sets into a single `__test` object fixes it; the CHANGELOG conflict just needs both bullets under `### Changed`. Please run `test:budget-ui`, `test:people-pickers` and `test:frontend-audit` on the merged result.

## Nice to have

- **Half of the new focus test cannot fail.** In `test/test-meals.js` (lines 1186 and 1191) `document` is a plain object, so nothing the handler does can change `activeElement`: a handler that also calls `_container.querySelector('#week-prev')?.focus?.()` stays green (57/0). The grid spy is what actually holds the fix. Either drop the focus assertion or give the fake elements a `focus()` that writes `document.activeElement`.
- **The wiring test accepts any function.** `test/test-meals.js:1241` checks `typeof call.handler === 'function'`, so `addEventListener('change', renderWeekGrid)` stays green (57/0), which would bring round 4's full-grid rebuild back through the wiring. Comparing `call.handler.name` to `'onNarrowWeekLabelQueryChange'` would close that.
- Whichever rule ends up keeping the fr/uk label unclipped below 640 px should be held by a test. Today removing the own-row block (`public/styles/meals.css:295-303`) keeps meals, frontend-audit and mobile-scroll-layout green, and the end date is clipped again. A guard next to the `.btn.is-current` one would do.
- Two comments describe the reset's row: `public/styles/meals.css:257-259` and `public/pages/meals.js:80`. Please keep them in line with wherever the reset ends up after the change below.

## Should fix: keep the reset on the stepper row on phones

I checked the own row with the maintainer, and the direction is to not spend a row on it. Below 640 px the Meals header is 173 px instead of 121 px on main (+52 px: the 48 px reset row plus the 4 px row gap), in every locale and on every week, so the week grid starts 52 px lower. On the current week, which is where the page opens, that row is empty. When the reset is visible it spans the full width (288 px at 320) under a stepper that stays left-aligned, and Calendar and Budget keep theirs on the stepper row.

- Below 640 px, render the Meals reset as an icon-only button on the stepper row, with the translated "Today" text as its `aria-label` (and `title`), same `.is-current` / `inert` / `visibility` rule as now, so the slot stays reserved and nothing jumps. From 640 px up it stays the text button.
- By my arithmetic that leaves the label roughly 130 px at 320 px (288 px row minus two 48 px arrows, a 44-48 px icon button and the gaps), against the 113 px the short format needs, in every locale, because the button no longer scales with the translated word. Please confirm that in the browser at 320 and 375 px in de, fr and uk, as you did this round.
- If you pick a Lucide icon, check that the name exists in `public/vendor/lucide`: an unknown name renders an empty button without any error.
- `NARROW_WEEK_LABEL_QUERY` (`public/pages/meals.js:93`) can then stay as it is; drop the year only where the full range really doesn't fit.
- The CHANGELOG line (`CHANGELOG.md:118`, "behind the period stepper in all three headers") stays true with this.

ulsklyc:
v2.67.0 shipped today (`fabab86b`), and it moved the whole `[Unreleased]` section down into the release block. That affects CHANGELOG bullets written before today: on the branch they sit correctly, but in the merge they can land **inside** the published `## [2.67.0]` section.

I could not measure it directly for this PR, because its merge tree still carries conflict markers from the base change, so please treat this as something to check rather than as a confirmed finding. On the other open PRs I did measure it, and there the bullets do land in the released block.

When you resolve the conflict with `main`, please make sure the bullet ends up under `## [Unreleased]`:

```bash
```

Nothing signals this on its own: git merges the hunk cleanly and `test:changelog` checks structure rather than placement.

mclgoerg:
Pushed `361d6ec4`.

## The stale merge / v2.67.0 CHANGELOG hazard

Re-merged `upstream/main` (now includes v2.67.0 and the cycle-tracker/fasting work). No conflict markers this time - the CHANGELOG hunk landed cleanly again, but exactly as flagged: the period-nav bullet ended up inside the now-published `## [2.67.0]` section instead of `## [Unreleased]`. Moved it back by hand and verified with your own check:

```
$ awk '/^## \[Unreleased\]/{p=1} /^## \[/{if(p&&!/Unreleased/)exit} p' CHANGELOG.md
## [Unreleased]

### Changed

- **The jump-to-now reset sits behind the period stepper in all three period-navigation headers,
  and hides while the current period is on screen** (#1164). ...
```

`budget.js`'s `__test` export is still a single object at the end of the file (no new collision from this merge - `grep -c "^export const __test"` is 1). `test:budget-ui` (76/0), `test:people-pickers` (15/0) and `test:frontend-audit` (396/0 at that point) all green on the merged tree.

## Should-fix: icon-only reset below 640px

Implemented as scoped. `#week-today` stays on the stepper row at every width now; `public/styles/meals.css:295-303`'s own-row block (`flex-wrap`, `flex-basis: 100%`, `order: 1`) is gone, replaced by a narrow-width rule that only constrains the button to arrow width (`min-width/min-height: var(--target-base)`, same token the arrows implicitly get via `--target-base`'s touch bump). `syncTodayButton()` (meals.js) now also drives the button's content: below 640px it clears the text and inserts a single `<i data-lucide="calendar-check" aria-hidden="true">`, calling `lucide.createIcons({ el: btn })`; at/above 640px it's the translated text again. `t('meals.today')` always goes onto `aria-label`/`title` regardless of width. `dataset.iconOnly` avoids redundant DOM churn on every render.

Went with `insertAdjacentHTML` (not `innerHTML =`) to stay inside the existing invariant - `test:frontend-audit`'s innerHTML/outerHTML guard caught my first draft immediately.

**Icon name:** `calendar-check`. Checked against the bundle the way `test-lucide-icons.js` does it (`PascalName:` inside `public/lucide.min.js`) before using it - `CalendarCheck` is present. `test:lucide-icons` also passes (3/3).

**Real-browser measurement** (Puppeteer against a live seeded instance, same harness as `test/document-guards-harness.js`), current week (reset hidden, slot still reserved) and next week (reset visible) at 320px and 375px in de/fr/uk - 12 layouts total:

| locale | width | label | today width | clipped |
|---|---|---|---|---|
| de | 320 | "14.09. – 20.09." (132px) | 48px | no |
| de | 375 | "14.09. – 20.09." (133px) | 48px | no |
| fr | 320 | "14.09. – 20.09." (132px) | 48px | no |
| fr | 375 | "14.09. – 20.09." (133px) | 48px | no |
| uk | 320 | "14.09. – 20.09." (132px) | 48px | no |
| uk | 375 | "14.09. – 20.09." (133px) | 48px | no |

Next-week pass (button actually visible, icon-only, aria-label correctly localized - "Heute"/"Aujourd'hui"/"Сьогодні") repeated the same result: no clipping, 48px reserved in every locale. Matches your arithmetic - the label gets ~130-133px against the ~113px the short format needs, and the reserved slot no longer scales with the translated word.

`NARROW_WEEK_LABEL_QUERY` and the drop-the-year logic are untouched. The CHANGELOG line at (now) `CHANGELOG.md:12` ("behind the period stepper in all three headers") stays accurate - it describes the position fix from #1164, not the row/icon detail, and that's still true.

Tab order / inert: unchanged mechanism, only content changes with width - confirmed via `test:meals`'s existing `.is-current`/`inert` tests, which stayed green throughout (they exercise the real `syncTodayButton()`, not a text-content assumption).

Comments updated: `meals.css:267-286` documents the reversal from Round 4's own row to the icon-only approach, `meals.js:79-83` and the `syncTodayButton()` doc comment (`meals.js:296-311`) describe the new width-driven content swap.

Added a CSS-mechanism guard in `test-frontend-audit.js`, next to the existing `.btn.is-current` one: checks the `@media (max-width: 639px)` rule for `.week-nav__today` requires `min-width: var(--target-base)` and rejects `flex-basis: 100%`. Counter-proved: swapped the rule back to Round 4's `flex-basis: 100%` own-row block - 396/1 red. Restored - 397/0 green.

## Nice-to-haves

- **Focus stand-in couldn't fail.** Gave `fakeResetButton()` a real `focus()` that writes `globalThis.document.activeElement = this`. Counter-proved: temporarily added `_container?.querySelector('#week-prev')?.focus?.()` inside `updateWeekLabel()` (a handler that also touches focus) - `test:meals` went 56/1 red on exactly the intended assertion ("onNarrowWeekLabelQueryChange() darf den Fokus nicht verschieben..."). Reverted - 57/0 green.
- **Wiring test accepted any function.** Changed `typeof call.handler === 'function'` to `call.handler.name === 'onNarrowWeekLabelQueryChange'`. Counter-proved: registered `renderWeekGrid` directly on the media-query listener - 56/1 red with the expected message. Reverted - 57/0 green.

## Verification

All of `test:calendar`, `test:calendar-structure`, `test:calendar-defaults`, `test:meals` (57/0), `test:meals-routes`, `test:budget-ui` (76/0), `test:budget-structure`, `test:people-pickers` (15/0), `test:frontend-audit` (397/0), `test:mobile-scroll-layout`, `test:i18n`, `test:changelog`, `test:suite-chain` green on the pushed head. Full `npm test` (297 suites in the chain) ran to completion against the exact committed tree, `fail 0` on every suite, no `✖` anywhere.

GitHub Actions on `361d6ec4`:
- [Tests (Node.js 22.x)](https://github.com/ulsklyc/yuvomi/actions/runs/35087642347/job/104766147575) - success
- [Tests (Node.js 24.x)](https://github.com/ulsklyc/yuvomi/actions/runs/35087642347/job/104766147381) - success

`gh pr view 1200 --json headRefOid,mergeable` now reports `headRefOid: 361d6ec4...` (matches) and `mergeable: MERGEABLE` (no longer `CONFLICTING`).

ulsklyc (changes requested):
Round 5 read against `361d6ec4`, and against the merge with current `main` (`ba79cbe4`), materialised with `git merge-tree --write-tree` plus `git archive` into a scratch tree.

Three of the four open points are closed, measured rather than taken on trust: `budget.js` carries exactly one `export const __test` (`node --check` clean at head and in the merge), the reset stays on the stepper row below 640px and is icon-only as decided, and the #1164 bullet sits under `## [Unreleased]` both at head and in the resolved merge (line 62, with `## [2.67.0]` starting at line 84). `calendar-check` is in the bundle and `lucide.createIcons({ el })` is the scoped call this repo patches in `public/lucide-scope.js`, so that part is right.

## Blocking

- `test/test-frontend-audit.js:17889` - the new CSS guard's comment states "die JS-Haelfte - Icon statt Text - haelt test-meals.js fest". Nothing in `test/test-meals.js` reaches that half: every `matchMedia` stub there returns `{ matches: false }` (`test/test-meals.js:966`, and the two `fakeResetButton()` factories give `insertAdjacentHTML()` an empty body), so the narrow branch is never entered. Counter-proof on the resolved merge tree: deleting `public/pages/meals.js:337-353` in full, that is the `aria-label`/`title` assignment, the icon insertion and the text restore, leaves `test:meals`, `test:frontend-audit` and `test:mobile-scroll-layout` at exit 0 with zero failures. CONTRIBUTING.md, "What a human guarantees": "Behaviour that is claimed to be fixed carries a test that fails against the state before the fix. A guard that was never seen red is not evidence." Please drive the real `syncTodayButton()` once with a `matchMedia` stub returning `{ matches: true }` and assert both directions: icon-only with no visible word and the translated word on `aria-label`, then back to the visible word above the breakpoint. Correct the comment once the sibling test exists.

- The branch is `CONFLICTING` again because `main` moved to `4843896d` and `ba79cbe4`. The cause is mechanical, not substantive: `git merge-tree` reports a conflict in `CHANGELOG.md` alone, where both sides append directly under `## [Unreleased]`. `package.json` and `test/test-frontend-audit.js` are the only other files you share with those two commits and both auto-merge cleanly. Resolve by keeping both sides, and put `### Changed` ahead of `### Security` so the section order matches the rest of the file. For reference, on that resolution `test:changelog`, `test:meals`, `test:frontend-audit`, `test:budget-ui`, `test:calendar`, `test:lucide-icons`, `test:i18n`, `test:suite-chain`, `test:parallel-runner` and `test:readme-consistency` were all green here.

## Should fix

- `CHANGELOG.md:10-24` - the bullet describes position and visibility only. Below 640px the Meals reset now loses its visible word entirely and becomes an icon, which every phone user sees; the word only survives on `aria-label`/`title`. That is a user-facing change and belongs in the entry, one clause is enough.

## Nice to have

- `public/pages/meals.js:87` - a bare empty line inside the block comment breaks the ` * ` column of an otherwise continuous paragraph.
- `public/styles/meals.css:285-288` repeats the three declarations of `.btn--icon` (`public/styles/layout.css:5093-5097`) verbatim. If the narrow reset is meant to be an icon button, say so at the rule so the touch-target value keeps a single source.
- `test/test-frontend-audit.js:17899` - the guard rejects `flex-basis: 100%` inside the `.week-nav__today` rule only. The round-4 shape it guards against also needed `.week-nav { flex-wrap: wrap; row-gap: ... }`, which this test cannot see; pinning the absence of that too would close the shape rather than one of its two declarations.
- `docs/test-suites.md` still describes the Meals period head as position plus visibility. The content now depends on viewport width as well, which is worth one half-sentence where the suite is described.

mclgoerg:
Round 6 fixed at `aa003329` (merge of `74ee98ff` with `upstream/main` @ `ba79cbe4`). CI green on both Node 22.x and 24.x: https://github.com/ulsklyc/yuvomi/actions/runs/35148863280

## Blocking 1 - icon-only behavior was never tested

Added a real behavioral test in `test/test-meals.js` ("syncTodayButton() schaltet unter 640px wirklich auf ein textloses Icon um und zurueck"): it stubs `window.matchMedia` to return `{ matches: true }` for `NARROW_WEEK_LABEL_QUERY`, drives the actual `syncTodayButton()`, and asserts both directions - narrow: `textContent === ''`, inserted markup contains `data-lucide="calendar-check"`, `aria-label`/`title` carry the translated word, `dataset.iconOnly === 'true'`; then `matches:false` again: `textContent` restored to the word, `dataset.iconOnly === 'false'`. `fakeResetButton()`'s `insertAdjacentHTML()` now actually records the inserted markup (it used to discard it, which is part of why the gap existed).

Counter-proof, re-run exactly as you did it: deleted `public/pages/meals.js:337-353` in full (the `aria-label`/`title` assignment, icon insertion, text restore).

- Before deletion (baseline): `test:meals` 58 passed, 0 failed.
- After deletion: `test:meals` 57 passed, **1 failed** (only the new test - "unter 640px muss der Reset ein data-lucide=\"calendar-check\"-Icon einfuegen"); `test:frontend-audit` 397 passed, 0 failed; `test:mobile-scroll-layout` 9 passed, 0 failed - confirming your finding that neither of those two ever saw this regression.
- After restoring the code: `test:meals` back to 58 passed, 0 failed; `git diff --stat` on `meals.js` empty.

Fixed the misleading comment on the CSS guard (`test/test-frontend-audit.js`, the Round 5 Should-fix test) - it no longer credits `test-meals.js` in general, it names the new test by description and says explicitly that the CSS guard covers the CSS shape only.

## Blocking 2 - merge conflict

`git fetch upstream && git merge upstream/main`. Exactly one conflict, in `CHANGELOG.md`, both sides appending under `## [Unreleased]`; `package.json` and `test/test-frontend-audit.js` auto-merged cleanly as predicted. Resolved keeping both sides' bullets, `### Changed` ahead of `### Security` (matching `### Added`/`### Changed`/`### Fixed`/`### Security` used elsewhere in the file, e.g. `[2.67.0]`, `[2.66.0]`, `[2.65.0]`).

`[Unreleased]` placement re-verified on the resolved merge:
```
```
prints `### Added` (#1208, #1205) then `### Changed` (our #1164 bullet, present) then `### Security` (#1241 x2) - nothing from `[2.67.0]` included, our bullet confirmed inside `[Unreleased]`.

Suites re-verified on our own resolution (not assumed from your numbers): `test:changelog` 34/0, `test:meals` 58/0, `test:frontend-audit` 400/0, `test:budget-ui` 76/0, `test:calendar` 27/0, `test:lucide-icons` 3/0, `test:i18n` 54/0, `test:suite-chain` 8/0, `test:parallel-runner` 24/0, `test:readme-consistency` 17/0 - all green.

## Should fix - CHANGELOG entry incomplete

Added one clause to the existing #1164 bullet: "Below 640px, Meals' reset additionally loses its visible word and becomes an icon-only button, the same width the two arrows already use; the word survives only on `aria-label`/`title`, not visually."

## Nice to have (all four)

1. `public/pages/meals.js:87` - removed the bare blank line inside the block comment; the paragraph break now uses ` *` like the rest of the comment.
2. `public/styles/meals.css:285-288` - kept the duplication (adding `.btn--icon` would also pull in `.week-nav .btn--icon { color: var(--color-accent) }`, which isn't part of this fix and would change the icon-only button's color as an unintended side effect) but added a comment at the rule stating explicitly that it deliberately re-declares `.btn--icon`'s touch-target shape, so the value keeps one conceptual source even though it's physically duplicated.
3. `test/test-frontend-audit.js` - extended the Round 5 CSS guard to also assert `.week-nav` carries no `flex-wrap: wrap` (neither inside the narrow media query nor in the base rule), closing the parent half of round 4's old shape, not just the child's `flex-basis`. Counter-proved: temporarily added `flex-wrap: wrap` to the base `.week-nav` rule, `test:frontend-audit` dropped to 1 failing (exactly this assertion); reverted, back to 397/0 (before merge) / 400/0 (after merge, once the new blocking-1 test's suite count is included), `git diff --stat` on `meals.css` empty afterward.
4. `docs/test-suites.md` - added a half-sentence to the `test:meals` line noting the icon-vs-text switch below 640px is now covered too, alongside position and visibility.

## Verification

After the merge, all of: `test:calendar`, `test:calendar-structure`, `test:calendar-defaults`, `test:meals`, `test:meals-routes`, `test:budget-ui`, `test:budget-structure`, `test:people-pickers`, `test:frontend-audit`, `test:mobile-scroll-layout`, `test:i18n`, `test:changelog`, `test:suite-chain`, `test:lucide-icons`, `test:parallel-runner`, `test:readme-consistency` are green, plus a full `npm test` at exit 0 against the exact committed HEAD (`aa003329`) before pushing.

Pushed SHA: `aa003329f775e47eb272a1678c415423dbbfe3fb`. `gh pr view` confirms `headRefOid` matches and `mergeable`/`mergeStateStatus` are `MERGEABLE`/`CLEAN` (no longer `CONFLICTING`). CI (both Node versions) confirmed green on this exact SHA, not assumed from the local run.

ulsklyc (approved):
Round 6 read against `aa003329` and against the resolved merge with current `main` (`b592a43f`), not against the branch alone. `aa003329` has real merge parents (`74ee98ff` + `ba79cbe4`), so the base is genuinely current as of that merge; your own work since round 5 is the single commit `74ee98ff` (5 files, 104 lines), and that is what I read.

Everything below was measured, not taken on trust.

**Round 5's blocker is closed, and the guard is real.** The exact mutation round 5 described - deleting `public/pages/meals.js:337-352` in full - now turns `test:meals` red (57/1, was 58/0). Four narrower mutations go red too: renaming the icon to `calendar-x`, dropping the `aria-label` assignment, dropping the `title` assignment, and neutering the text-restore body while keeping `dataset.iconOnly = 'false'`. Both directions of the new test carry weight, which is what was missing. `calendar-check` resolves to `CalendarCheck` in the bundle and `test:lucide-icons` goes red on a bogus name, so the icon is covered as well.

**Round 5's nice-to-have 3 is closed and mutation-proven.** All three assertions fire: `flex-wrap: wrap` in the `.week-nav` base rule, `flex-wrap: wrap` inside the `@media (max-width: 639px)` block, and `flex-basis: 100%` on the narrow `.week-nav__today` each turn `test:frontend-audit` red on their own (401/0 clean, 400/1 under each mutation). The corrected comment at `test/test-frontend-audit.js:17889` now names the right sibling test.

The other three round-5 items - the blank line in the block comment, the `.btn--icon` duplication note in `public/styles/meals.css`, and the `docs/test-suites.md` sentence about viewport dependence - are all in. The CHANGELOG now names the icon-only change ("loses its visible word and becomes an icon-only button ... the word survives only on `aria-label`/`title`"), which closes round 5's should-fix.

**On the green checks:** they are bound to `baseSHA=ba79cbe4`, thirteen commits behind `main`, so they do not speak for the merge. I ran it myself on a real merge with `b592a43f`: `test:meals`, `test:budget-ui`, `test:calendar`, `test:people-pickers`, `test:frontend-audit`, `test:i18n`, `test:i18n-translated`, `test:lucide-icons`, `test:mobile-scroll-layout`, `test:changelog`, `test:suite-chain` and `test:readme-consistency` are all green. `public/pages/calendar.js` moved on both sides but auto-merges cleanly, and neither `calendar.js` nor `budget.js` ends up with a second `export const __test` - the round-4 trap does not come back.

Hard constraints are clean on your own diff: no `innerHTML`/`outerHTML` write (the single hit is a fake-DOM property in `test/test-meals.js`), no `eval`, no `require`, no `toISOString().slice(0, 10)`, no new i18n keys (all nine keys you call already exist in all 24 locales), every new CSS declaration from `tokens.css` (`--space-2`, `--target-base`), no server or `MIGRATIONS` change, and no em- or en-dash in new prose.

## Blocking

- Nothing.

## Should fix

- The branch is `CONFLICTING` only because `main` moved to `b592a43f`; `git merge-tree` reports `CHANGELOG.md` as the single conflicted file, and `docs/test-suites.md`, `package.json`, `public/pages/calendar.js` and `test/test-frontend-audit.js` all auto-merge. Resolve by keeping both sides with `### Changed` ahead of `### Fixed`, which gives `Added / Changed / Fixed / Security` and matches the rest of the file. On that resolution the #1164 bullet sits at `CHANGELOG.md:56` under `## [Unreleased]` (line 8), well ahead of `## [2.67.0]` at line 181, and `test:changelog` is green (34/0). This is mechanical and does not need another review round.

## Nice to have

- `test/test-frontend-audit.js:18039-18041` - the narrow `.week-nav` lookup matches on `/max-width:\s*639px/`, and the assertion sits behind `if (weekNavRule)`. That is fine today and it does fire once such a rule exists (verified). It stays blind only if the wrap comes back under a differently spelled threshold, e.g. `@media (max-width: 640px)`. Matching any `max-width` at or below the breakpoint would close that last gap. Not worth a round on its own.

## Verdict

approve - six rounds in, every finding from round 5 is closed and independently mutation-verified, and the only thing left is the CHANGELOG conflict resolution at merge time.

mclgoerg:
Merged current `main` (`b592a43f`). The only conflict was `CHANGELOG.md` as predicted — resolved keeping both sides, `### Changed` ahead of `### Fixed` (giving `Added / Changed / Fixed / Security`, matching the rest of the file). Re-verified `[Unreleased]` placement with your own check:

```
```

— our #1164 bullet is present under `Changed`, nothing from a released section included.

Full `npm test` green on the merged tree. CI confirmed on `10297c82`: both Node 22.x and 24.x pass. Ready to merge.

PR #1255, opened 2026-09-17
feat(documents): expiry dates and reminders for passports, permits, and other documents

mclgoerg:
## Summary

Family documents can now carry an optional expiry date and a reminder lead time - a passport, a
residence permit, an insurance card, anything with a "this stops being valid on..." date.

- `expires_at` and `expiry_reminder_days` (0-365 days, mirroring
  `inventory_item_dates.reminder_offset_days`; an explicit `0` lead is preserved, not coerced to a
  default) on `family_documents`, validated on both `POST /documents` and `PUT /documents/:id`.
- One `reminders` row per document (`entity_type = 'document_expiry'`, derived - not settable
  through the generic `POST /reminders`, since `syncDocumentExpiryReminder()` fully owns it),
  recomputed on every write - no periodic sync needed, since a document only changes when someone
  edits it. Torn down on delete, on a folder-tree delete, and on archiving (an archived scan
  shouldn't keep nagging).
- An expiry status chip and an "expiring soon" filter chip on the Documents page, plus an
  `?expiring=<days>` list filter (bound to the household's own day, not the UTC day) and an expiry
  sort option.
- `dateStatus()`/`DATE_STATUS_ALERT_DAYS` extracted from `inventory-warranty.js` into a neutral
  `public/utils/date-status.js` (re-exported from its old home under its old name, so existing
  Inventory imports are untouched) - Documents needed the same "valid / expiring / expired" chip
  Inventory's warranty status already had, and a second copy of that logic wasn't the answer.
- The `reminders.entity_type` CHECK is widened for `document_expiry` in this migration (218) only -
  a follow-up PR that also needs a new entity type widens it again itself when it actually starts
  writing that type.

## Test plan

- [x] New suite `test:document-expiry` (20 tests): validation incl. an explicit `0` lead
  preserved, reminder created/moved/removed on write, no reminder for an already-past date,
  teardown on delete/folder-tree-delete/archive, ownership follows `created_by` not the editor, the
  `?expiring=` filter (incl. a household-timezone-boundary case), a module-scoped credential and a
  second household member each get an empty `/pending`, not another member's document.
- [x] `npm test` - full suite, exit 0, rebased onto current `main`.


ulsklyc (changes requested):
Thanks for this, and for the SPEC and test work that came with it. The feature is cleanly built; three things have to change before it can go in, and the first one affects all three of your open PRs.

Reviewed against `23d1caaf`, with `main` at `a7380a9c9`.

## Blocking

**1. `server/db.js:8591` - migration `version: 217` is already taken on `main`.**

`main` carries `217` since #1258 ("Reminders of a deleted task or event are removed with it"). `migrate()` filters with `MIGRATIONS.filter((m) => !applied.has(m.version))`, so on every installation that already applied the other 217, yours counts as done and never runs. `family_documents.expires_at` is then missing while `documentSelect()` selects `d.expires_at`, and the whole documents list answers "no such column". `test:migrations-append-only` goes red as well.

The next free number is **218**, and #1256 and #1257 sit on top of this branch and take 217/218/219 too, so all three need renumbering in stack order. Renumbering is mine to coordinate rather than yours to guess at, so please do not race it: tell me the order you want them merged in and I will give you the numbers.

**2. `server/routes/reminders.js:19` - `document_expiry` is settable, and that leaks document names.**

It is listed in `VALID_ENTITY_TYPES` but not in `DERIVED_ENTITY_TYPES` (line 86), so it ends up in `SETTABLE_ENTITY_TYPES`. `POST /api/v1/reminders` then accepts `{entity_type: 'document_expiry', entity_id: <any id>}` and only checks `mayTouchOrigin(req, 'document_expiry', 'write')`, which is the `documents` module as a whole. There is no visibility check on the individual document. `GET /reminders/pending` resolves the title with `(SELECT name FROM family_documents WHERE id = r.entity_id)` and no `documentVisibleSql()`, so a member can walk the ids and read back the names of other people's private documents, and `processDueNotifications()` will even push them.

The same line also contradicts the design: the reminder is fully derived. `syncDocumentExpiryReminder()` deletes **all** `document_expiry` rows for an id on every write, including hand made ones, while `POST /reminders` only ever deletes the caller's own. Moving `document_expiry` into `DERIVED_ENTITY_TYPES` fixes both.

**3. `server/routes/documents.js:1092` - `date('now')` is the UTC day.**

```sql
AND d.expires_at IS NOT NULL AND date(d.expires_at) <= date('now', '+' || @expiringWithinDays || ' days')
```

`expires_at` is a locally entered calendar day; `date('now')` is UTC. East of UTC a document that expires in exactly N days drops out of the filter during the local evening, west of it one extra day comes in. This is the only `date('now')` in `server/routes/` and `server/services/` - `server/services/task-scope.js:39` spells out why ("the day comes in as a parameter on purpose, not as `date('now')` from SQLite"). It wants a bound day key from `todayKey(db)` / `shiftDateKey()` (`server/utils/timezone.js`).

The new test hides it: `test/test-document-expiry.js` builds its date keys in `dateKey()` from `getUTCFullYear/getUTCMonth/getUTCDate`, so both sides are UTC and the test stays green in every zone.

## Should fix

- **`server/routes/documents.js:1053-1060`** - the comment says an invalid value is rejected rather than silently skipping the filter, the code does exactly the opposite: `?expiring=abc` gives `NaN` gives `expiringWithinDays = null` gives an empty clause gives 200 with the full list. Either answer 400 or correct the comment; as it stands the comment immunises the code against its own counter-check.
- **`server/openapi/schemas.js:265` and `server/openapi/paths/documents.js:236`** - `expires_at` and `expiry_reminder_days` now come out of `GET/POST/PUT /api/v1/documents` but are not in `FamilyDocument.properties`, and `?expiring=` is missing from `params` where `status` and `category` are described. `/api/v1` is promised surface, the spec is the contract.
- **`server/routes/documents.js:1091` vs `public/pages/documents.js:385`** - the `?expiring=` filter has no caller. The chip filters client side over `state.allDocuments` (`matchesExpiringSoon`). So the server filter exists only for API consumers, which is exactly where the UTC bug above sits and where no UI ever makes it visible. Either hang the UI on it or drop the branch and its SPEC paragraph.
- **`server/db.js:8600`** - `'health_prevention_due'` joins the CHECK although, in your own words, nothing writes that type yet. A CHECK can never be narrowed again under the append-only rule, so a documents PR would permanently fix a reminder type for a feature that does not exist, without an issue and without an entry in `docs/SCOPE.md` or `docs/DECISIONS.md`. It is not in `ORIGIN_MODULE` or `REMINDER_ORIGINS` either, so it could not be delivered. It belongs in the follow up PR.
- **`docs/SPEC.md:2096-2097`** - the table says "(migration v209)", the paragraph below says "(v217)". v209 is the fasting migration. One of the two is wrong, and after renumbering both are.
- **`public/locales/ar.json`, `cs.json`, `pl.json`, `ru.json`, `uk.json`** - `expiredDays`/`expiringInDays` only get the base form and `_one`, while the same files carry `_few`/`_many` for comparable counting keys (`shoppingOpen_few`). `test:i18n-plural` only enforces `_one`, so the run stays green while Arabic reads wrong at count=2 and Czech from 5 on.
- **`test/test-document-expiry.js:1420-1428`** - the harness mounts `documentsRouter` behind a stub that sets `req.authUserId`/`req.session` directly: no module gate, no CSRF, no `access_permissions`. So the suite measures no rights or visibility axis at all, and now that the reminder carries a document name, that is the axis that needs one. `process.env.TZ` is not pinned either.
- **`CHANGELOG.md:54`** - no issue reference, unlike every neighbouring entry. CONTRIBUTING asks for an issue first for anything beyond a trivial fix.

## Nice to have

- `public/utils/date-status.js:11` still exports `WARRANTY_ALERT_DAYS`. For documents the name lies. `test/test-frontend-audit.js:14370` already reads it at the new path, so a rename with a re-export from `inventory-warranty.js` would be one more place.
- `public/pages/documents.js:392` - `expiringCount()` calls `dateStatus(doc.expires_at)` twice per document.
- The PR text and `docs/SPEC.md:2149` say the write path is `PATCH /documents/:id`; the code changes `router.put('/:id')`. The only PATCH is `/:id/archive`.

mclgoerg:
Thanks for the thorough review. All three blocking points fixed, plus every should-fix and nice-to-have, at `d5e15ab9`.

**Migration order**: this PR (#1255) → #1256 → #1257, unchanged from how they're stacked. Renumbered in that order: **218** here, **219** in #1256, **220** in #1257 (all three pushed).

**Blocking**
1. Renumbered 209→218. Also moved `health_prevention_due` out of this migration's CHECK entirely - #1256 now widens the CHECK itself when it first writes that type, per your point in "should fix" below.
2. `document_expiry` moved into `DERIVED_ENTITY_TYPES`. Added the missing rationale paragraph to the file's own comment block (it already explains why `subscription`/`inventory_item` etc. are *not* derived; this explains why `document_expiry` is, unlike its siblings).
3. Fixed via `todayKey(db.get())`, same pattern as `task-scope.js`. Added a suite following `test-budget-month-zone.js`'s technique (`mock.timers` frozen at a UTC/household-day boundary) so a regression to `date('now')` would actually go red instead of passing silently in UTC CI.

One more blocking issue I found while fixing #3, not in your review: the `reminders` table rebuild itself crashed on any real installation. Migration 217 (also new, from #1258) adds two `AFTER DELETE` triggers on `tasks`/`calendar_events` whose bodies reference `reminders` by name. Rebuilding `reminders` via create-new/drop-old/rename-to-original-name trips a SQLite quirk where `ALTER TABLE ... RENAME TO` reparses the whole schema mid-operation and hits the momentarily-absent table - `SqliteError: no such table: main.reminders`, thrown from inside the migration transaction. Reproduced in isolation against `better-sqlite3` directly (nothing app-specific), fixed by dropping the two triggers before the rebuild and recreating them after - same technique `trg_search_tasks_ad` already uses in v114/v117/v166/v194, just in the other direction (there the rebuilt table lost its own triggers; here a foreign trigger lost its target).

**Should fix** - all applied: OpenAPI schema/params, i18n `_few` for ar/cs/pl/ru/uk (and the placeholder added everywhere else so the key set stays identical across locales), the stray v209/PATCH-PUT wording in SPEC.md, the invalid-`?expiring`-value comment corrected to match the actual silent-skip behavior, and real test coverage added for the derived-type rejection and the module-scope axis on `/pending` (a `calendar:read`-only credential can no longer reach a document name through the reminder toast).

I left the `?expiring=` filter itself in (fixed, not dropped) since it's genuine `/api/v1` surface for outside consumers even though the app's own chip filters client-side - documented that split explicitly in SPEC.md now so it doesn't read as dead code.

**Nice to have** - all applied: `WARRANTY_ALERT_DAYS` renamed to `DATE_STATUS_ALERT_DAYS` at its actual home in `date-status.js`, with the old name kept as an alias export from `inventory-warranty.js`; `expiringCount()` no longer computes `dateStatus()` twice per document; PR description corrected (`PUT`, not `PATCH`).

Full `npm test` green (7828 assertions).

ulsklyc (commented):
Re-reviewed against `d5e15ab908ddfa17f5d02fbe6d024fc658a7f651`, diffing only this PR's own two commits against the merge base `a7380a9c9`, with `main` at `b592a43fb`. All three blocking points are genuinely fixed, and I verified each one at the code rather than taking the summary for it.

The fourth one you found yourself is the interesting one. I reproduced the trigger crash independently against `better-sqlite3-multiple-ciphers` outside this tree, and it is exactly as you describe: `error in trigger trg_x: no such table: main.reminders`. Without the drop-and-recreate the migration would have aborted on every installation that carries `main`'s v217. The recreated bodies at `server/db.js:8708-8716` match v217 byte for byte. Thanks for chasing that down instead of trusting a green suite.

I also ran 31 suites on the merge result against current `main` (not on your base): `migrations-append-only`, `schema-mirror`, `rename-migration`, `document-expiry`, `reminder-orphans`, `reminders-routes`, `suite-chain`, `i18n`, `i18n-plural`, `i18n-translated`, `changelog`, `frontend-audit`, `openapi-coverage`, `openapi-structure`, `notifications`, `push`, `documents`, `documents-ux`, `typography`, `sw-precache`, `layer-boundary`, `inventory-warranty-status`, `inventory-tracked-dates`, `inventory-item-reminders`, `reminders`, `multi-reminders`, `reminders-entity-type-migration`, `mcp`, `search-permissions`, `document-owner`, `document-folder-delete`. All green.

One correction of my own: my round-1 request for `_many` in ar/cs/pl/ru/uk was wrong. There is no `_many` key anywhere in `public/locales/`; the house form is base + `_one` + `_few` in every locale, which is what you shipped. Ignore that half of the point.

## Blocking

Nothing.

## Should fix

- **PR description is now stale in two ways, and it becomes the squash commit message.** It still advertises `health_prevention_due` ("It also widens it for `health_prevention_due` at the same time") - you removed that, correctly, and `server/db.js:8663-8669` now explains why. It also still says the write path is `POST /documents` and `PATCH /documents/:id`; the code changes `router.put('/:id')` (`server/routes/documents.js:1226`), and you already fixed exactly this wording in `docs/SPEC.md`. The "Test plan" count (15 tests) is short of what the file now has as well.
- **`CHANGELOG.md:54` still carries no issue reference**, unlike every neighbouring entry, and I could not find an issue for this feature either. CONTRIBUTING asks for an issue first for anything beyond a trivial fix - a feature that adds two columns, a migration and a new reminder origin is the case it is written for.
- **`test/test-document-expiry.js` still has no per-member visibility axis and no pinned zone.** The module-scope test at line 510 is a real improvement and uses the genuine `mayTouchOrigin()` path, thanks. What is still missing is the axis the finding was about: every test runs as `OWNER` or `ADMIN`, so nothing measures a *second* member against a `visibility: 'private'` document. The reminder does stay safe today (`syncDocumentExpiryReminder()` sets `created_by` to the document's creator and `/pending` filters `r.created_by = ?`), but that is the invariant worth nailing down, because it is what keeps blocker 2 closed if `syncDocumentExpiryReminder()` is ever changed to fan out. Separately, `process.env.TZ` is still unpinned while `dateKey()`/`daysFromToday()` at line 114-126 compute in UTC, so every test except the Kiritimati one measures with both sides on the same clock.

## Nice to have

- **`test/test-document-expiry.js:443`** - `setHouseholdTimeZone('Pacific/Kiritimati')` is never reset. `mock.timers.reset()` in the `finally` puts the clock back, but the `sync_config` row stays, so every test declared after line 440 runs against a UTC+14 household. It is harmless right now (the later tests overwrite `remind_at` outright), but the next test appended to this file inherits a zone nobody set on purpose. A `test.after`, or setting it back to the previous value in the `finally`, keeps the file honest.
- **`server/db.js:8685-8686`** - `DROP TRIGGER trg_reminders_tasks_ad;` without `IF EXISTS`. It is safe as written, since v217 creates both unconditionally and always runs first. But a migration that throws leaves an installation stuck at 217 with a half-built schema, and `IF EXISTS` costs nothing for the one case where the assumption turns out not to hold.
- **`docs/SPEC.md:2147-2164`** uses long dashes in the range notation and in the prose, where the house style is a plain hyphen. The rest of that file does too, so it is a pre-existing habit rather than something you introduced; mentioning it only so the next pass over the section can pick them up.

mclgoerg:
Thanks for the re-review and for catching your own round-1 slip on \`_many\` - appreciated either way.

All should-fix and nice-to-have items applied at \`1c2b7ab2\` (rebased onto current \`main\`, which moved again in the meantime):

- PR description corrected - dropped the stale \`health_prevention_due\` mention (now accurate per \`server/db.js:8663-8669\`'s own comment), \`PUT\` not \`PATCH\`, test count updated to match the file.
- Added the missing visibility axis: a genuinely different second household member now gets an explicitly-asserted empty \`/pending\` against another member's \`private\` document with a due reminder, rather than the invariant only being described in a comment.
- \`process.env.TZ = 'UTC'\` pinned explicitly at the top of the file.
- The household-timezone test now clears its \`sync_config\` override in its own \`finally\` instead of leaking \`Pacific/Kiritimati\` into every test declared after it.
- \`DROP TRIGGER\` now takes \`IF EXISTS\`.

Left the \`CHANGELOG.md\` issue-reference point as-is - this was fork-first work and I couldn't find a matching upstream issue either. Didn't touch the pre-existing em/en dashes in \`docs/SPEC.md\` since you flagged those as a separate future pass, not this PR's job.

Full \`npm test\` green.

ulsklyc (approved):
Round 3 checked against `1c2b7ab2`. The conflict is gone - you rebased onto current `main`
(`8e8091287`), and the `package.json` resolution is clean: all 340 chain steps from `main` are
intact and `test:document-expiry` is the only addition.

Measured against an auto-merge of the previously reviewed head with current `main`, your own
delta since the last round is exactly three files - the `package.json` resolution, `DROP TRIGGER`
to `DROP TRIGGER IF EXISTS`, and the three test additions. No new functionality slipped in.

**On the migration crash you found: confirmed, and thank you for catching it.** I reproduced it
rather than take it on trust - real `better-sqlite3-multiple-ciphers`, the full v1-v217 chain,
then v218 with the drop/re-create removed:

```
error in trigger trg_reminders_tasks_ad: no such table: main.reminders
```

That would have aborted the migration on every existing installation. Neither of us saw it in
round 1 for a structural reason worth naming: the branch predated `main`'s v217, so the green CI
at the time could not have caught it.

The re-create is complete, which is the other half that matters - a drop without a full re-create
would be a silent data-loss path. Both triggers come back with byte-identical bodies, nothing
else in the schema references `reminders`, all four indexes survive, and deleting a task after
the migration still takes its reminders with it. `notification_deliveries` rows survive the
`DROP TABLE`, which is exactly what your `foreignKeysOff` note claims. `test-reminder-orphans.js`
already guards the re-create; I mutated a copy to remove only the two `CREATE TRIGGER` blocks and
it goes red (8 tests, 5 failures), so this is guarded rather than merely correct today.

Migration numbering is right: 218 entries, highest 218, no duplicate, and the `server/db.js` diff
is a single 70-line hunk with zero deletions, so `MIGRATIONS` grows only at the end.

I ran 30 suites locally, all green, including the three Inventory ones that import the
`dateStatus()` helper you moved.

One thing left, and it is not a blocker:

**`public/pages/documents.js:859`** - `renderExpiringChip()` renders unconditionally, including
at `count === 0`. The comment says "same pattern as `renderCategoryChips`", but that function
deliberately does the opposite a few lines up: only populated categories become chips, because
permanently visible filters that lead nowhere are noise. Right after the update no document has
an expiry date yet, so every household gets a permanent "expiring soon 0" chip that filters into
an empty list. Rendering only when `count > 0 || state.expiringSoon` follows the neighbouring
rule, and the second half keeps the active chip from jumping away while it is in use.

Minor, take it or leave it: `value="${doc?.expiry_reminder_days ?? ''}"` at line 1816 does not go
through `esc()` while the `expires_at` input two lines above does. Not exploitable - it is an
INTEGER column with an integer range validator and a single write path - just inconsistent
between two adjacent lines.

Heads-up on state, and it is our doing rather than yours: since I measured this, `main` has moved
to `5b44508cf` - #1200, #1178 and #1275 went in - so the PR is CONFLICTING again, in `CHANGELOG.md`
and `package.json`. The same two files as last time, and the same cause: us moving underneath you.
Nothing in the review above changes because of it.

One thing to watch when you resolve `CHANGELOG.md`: `main` now has a `### Changed` heading sitting
between the two sides, so the obvious "keep both" resolution drops your entry under `### Changed`
instead of `### Added` - and `test:changelog` stays green on that, so the suite will not tell you.

mclgoerg:
Thanks for the approval and for reproducing the trigger fix independently rather than taking it on trust.

Fixed both remaining points at \`4dc5c7c4\`:
- \`renderExpiringChip()\` now follows \`renderCategoryChips()\`'s own rule and only shows once it has hits (or is the active filter) - no more permanent "expiring soon 0" chip on a fresh install.
- The reminder-days input's value now goes through \`esc()\` like its neighbour.

Rebased onto current \`main\` again (moved twice more) - the \`CHANGELOG.md\` conflict needed the care you flagged; the entry now sits under \`### Added\`, above the new \`### Changed\` heading, not pulled under it.

Full \`npm test\` green.

ulsklyc:
Checked the delta from the approved `1c2b7ab2` to `4dc5c7c4`. The approval stands.

- **Rebase, nothing else.** `4dc5c7c4` sits linearly on current `main` (`5b44508cf`), no merge commits. `git range-diff 8e8091287..1c2b7ab2 5b44508cf..4dc5c7c4`: your two review-round commits are byte-identical, and the feature commit differs only in the conflict resolution in `CHANGELOG.md` and the `package.json` test chain. The one new commit touches only `public/pages/documents.js`.
- **Both points are fixed.** `renderExpiringChip()` now returns before rendering when `!count && !state.expiringSoon` (`documents.js:869`). So there is no "0" chip on a fresh install, and an active chip stays where it is. The reminder-days value goes through `esc()` (`documents.js:1822`), which is `String()`-safe for the integer and keeps an explicit `0`. Markup still goes through `replaceChildren()` + `insertAdjacentHTML`.
- **CHANGELOG:** compared against `main` in both directions. No `main` line is lost, and the only new text is your entry, under `### Added` above `### Changed` (`CHANGELOG.md:59`). Thanks for watching that one.
- **`package.json`:** all 341 steps of `main`'s chain are intact and in order. `test:document-expiry` is the only addition.
- **Migrations across the stack:** `main` ends at 217. This PR adds exactly 218 (one hunk, no deletions). #1256 adds 219 and #1257 adds 220. There are no duplicates, and v218 is byte-identical in all three branches. #1257 still carries an older revision of #1256's v219 and conflicts with `main`, so it needs its rebase once #1256 lands. That is expected and not something to do now.
- **Tests:** I tested on a separate worktree at `4dc5c7c4` merged with current `main` (a no-op, since you are already on it). 33 targeted suites all passed, and the full `npm test` chain ran green as well. CI is also green on Node 22 and 24.

Two notes, neither needs action from you:

- Neither round-3 change has a guard. Removing line 869 or the new `esc()` leaves every suite green. The `esc()` change is consistency only, so that is fine. The chip rule is a behaviour a later refactor could quietly undo. I'm not asking for a test in this PR.
- The PR title plus ` (#1255)` is 95 characters, and the limit for the squash subject is 72. I'll shorten it when I squash, for example to `feat(documents): add expiry dates and reminders (#1255)`.

PR #1256, opened 2026-09-18
feat(health): preventive care and vaccination tracking

mclgoerg:
Adds a Prevention tab to Health: a household-defined type registry (vaccination/check-up, icon, optional recurring interval) plus a per-person log of what was given and when.

### What changed

- A new **Prevention tab** shows a "due / overdue" list derived from each type's most recent record, and follows the same caregiver/visibility model as vitals/meds/labs/activities (`careAwareClause`/`writableClause`/`resolveOwner`): a caregiver logging for the person they care for gets that record scoped and defaulted through the owner's own visibility choice, not their own.
- A household-defined **type registry** (vaccination/check-up, an icon, and an optional recurring interval) - nothing is seeded, since a shipped catalog would go stale regionally. Deleting a type keeps its records via a name snapshot.
- **Caregivers also receive the due reminder themselves, subject to the owner's opt-in** (`health_prevention_notify_caregivers`, default off, mirroring `cycle_settings.notify_partner_user_id`). One `reminders` row per recipient (owner's carries `assigned_from = NULL`, a caregiver's carries `assigned_from = <subject>`), so each side's `pushed_at`/dismissed is independent; a caregiver's own self-set row is never overwritten; revoking the grant removes the inherited row immediately; a caregiver whose Health access is off gets nothing. Only the inherited row's push body names the subject, so a caregiver watching more than one person can tell them apart.
- Extracts `addMonthsClamped()`/`addYearsClamped()` into a new `server/utils/interval-date.js`, making `subscriptions.js#addBillingCycle` and `inventory-deadlines.js#warrantyEndDate` thin facades over the same month-addition-with-day-clamping math instead of a third copy.
- Migration 219 adds `health_prevention_types` and `health_prevention_records`.

### Test plan

- [x] `npm test` green locally
- [x] New suites: `test:health-prevention`, `test:health-prevention-interval`, `test:prevention-reminders`, `test:interval-date`


ulsklyc (changes requested):
Carefully built and well tested - thank you. The visibility work in particular is done the right way, through `careAwareClause()` almost everywhere. Three things have to fall before it can go in, and one of them is a decision for me rather than a defect in your work.

Reviewed against `0879c455`, with `main` at `a7380a9c9`. Since this PR stacks on #1255, I judged the second commit (`0879c455`, prevention); the document expiry parts belong to that review.

## Blocking

**1. `server/db.js:8591` - migration `version: 217` is already taken on `main`.**

Same as on #1255: `main` carries 217 since #1258, `migrate()` filters with `!applied.has(m.version)`, so on an existing installation this migration is **silently skipped**. No `family_documents.expires_at`, and no widened `reminders.entity_type` CHECK. The first `INSERT INTO reminders (... 'health_prevention_due')` then hits a CONSTRAINT error that the write path swallows into a log line: record saved, reminder never created, nothing visible to the user. `server/db.js:8631` (`version: 218`) collides with #1257 on top of that.

Next free is 218, and all three of your PRs need renumbering in stack order. Tell me the merge order you want and I will assign the numbers.

Checked and fine: the two triggers from `main`'s 217 hang on `tasks`/`calendar_events`, not on `reminders`, so they survive your rebuild; columns and the four indexes match the last rebuild on `main` (v203) exactly.

**2. `server/routes/health/prevention.js:380-390` - the one hand written visibility check has no test.**

Every other read path goes through `careAwareClause()`; this one spells the rule out itself. I walked it by hand and it filters correctly - but that is exactly what CLAUDE.md rules out as evidence: a guard never seen red proves nothing, and with vaccination data that is not a bar I want to lower. `test/test-health-prevention.js:6903` only exercises `/prevention/due` for the owner with a `family` record. Missing: Carol is neither owner nor caregiver and calls `/prevention/due?user_id=<Bob>` whose latest record is `private` (sees nothing), and the same with `family` (sees it).

## Should fix

- **`server/routes/health/visibility-defaults.js:37` + `public/settings/pages/personal-health.js:28-33`** - `prevention` is registered as a scope server side, but the row in the settings sheet is missing; `visibilityScopes()` still lists only `meds`/`labs`/`activities`/`fasting`. So `defaultVisibilityFor(db, ownerId, 'prevention')` stays `'private'` forever, because nobody can ever save anything else, and `defaultVisibility('prevention')` in `public/pages/health.js:3937` is dead code. The CHANGELOG promise at line 59 is then formally true but cannot be triggered. No test catches it: `test/test-health-visibility-defaults.js` does not enumerate the scopes.
- **`server/services/prevention-reminders.js:148-150`** - the fan out to caregivers is neither granted by the owner nor declinable by the recipient. The model you cite works differently: in `cycle-reminders.js#syncPartnerReminder` the owner publishes deliberately through `cycle_settings.notify_partner_user_id`. Here a care grant is enough, and from then on the type name of every prevention item, including ones marked `private`, plus the cared for person's name, lands on someone else's lock screen. Your PR text calls it a deliberate, accepted trade off; that is a decision for me, and I would rather make it before the merge than after. My inclination is to follow the cycle model, but let us settle it in the issue before you build anything.
- **`docs/test-suites.md:77`** - the line says a 62 route table while `test/test-health-structure.js:157` asserts `actual.length === 74` (65 on `main` plus your 9). 62 is right neither before nor after; the line was touched but with an invented number.
- **`docs/SPEC.md:131`** claims the prevention tab uses `dateStatus()` chips from `public/utils/date-status.js`. `public/pages/health.js` never imports it; `:3779` computes `const soon = !overdue && item.days_left <= 30;` with its own hard coded 30. The threshold now exists twice, and `test/test-frontend-audit.js`, which watches exactly those pairs, cannot see the second copy. Either use `dateStatus()` or drop the claim.
- **`public/settings/pages/modules-health.js:57`** - `t('settings.healthPreventionIntervalMonths', { count: type.default_interval_months })` renders a type entered in years as "every 120 months". The form right above (`:167`) offers months and years, and `intervalMonthsToInput()` is already imported next to it. Tetanus every 10 years is unreadable in the admin list.
- **`public/pages/health.js:4022`** - `if (nextDueOn) body.next_due_on = nextDueOn;` means an explicit due date can never be removed again through the UI. Clearing the field sends nothing, the stored value stays, and `computeDueForUser()` prefers `record.next_due_on` over the interval. An override accidentally set into the past then produces a permanent "overdue" row plus a reminder until the record is deleted. `interval_months` has the same behaviour on purpose right above, but there it only drives a display, not a reminder.
- **`server/routes/reminders.js:222` vs `server/services/notifications.js:238-240`** - the push names the cared for person (`preventionDueBody`), the in app toast through `/reminders/pending` does not fetch `prevention_subject_name` and shows only the type name. Your own reasoning in the module header (a caregiver with two charges cannot tell them apart) applies to the toast just as much.
- **`server/routes/health/prevention.js:373-375`** - the care check is written out as its own `SELECT 1 FROM health_care_grants ...` although `helpers.js#canWriteFor()` already holds that query, and the file header itself says `helpers.js` is the only place the clause may be built (#884). The third copy is the one that gets left behind at the next rework of the care model.

## Nice to have

- `server/routes/health/prevention.js:385-389` - one query per entry in the filter; `computeDueForUser()` has `visibility` loaded anyway.
- `public/pages/health.js:3813-3831` - the prevention rows carry `health-activity-row` classes from the activities tab, so a change to the activity look silently moves prevention.
- `public/pages/health.js:3879` and `public/settings/pages/modules-health.js:157` - inline layout (`style="display:flex;gap:var(--space-2)"`), while `public/styles/health.css` sits right there. The tokens are right, the place is not.
- `public/pages/health.js:3781` - `days_left === 0` reads as "due in 0 days"; a "today" string is missing.
- `health_prevention_types.key` (`server/db.js:8637`) is never read - the visibility scope is flat `'prevention'`, not per type. `public/settings/pages/modules-health.js:198-202` therefore slugifies it and appends a random suffix on 409. Either the column carries a job or it can go.
- `CHANGELOG.md:54` and `:63` carry no issue number, unlike their neighbours.

Explicitly checked and clean: no `innerHTML`/`outerHTML`, no CDN or npm frontend dependency, all user data through `esc()`, all 24 locales carry the new keys with real translations including `_one`, all referenced keys exist, all CSS values from `tokens.css`, every handler in `try/catch` with `{ data }`/`{ error, code }`, no `await` before a DB call, `todayKey()` rather than `toISOString().slice(0,10)` on both sides, the four new suites registered and hung into the `test` chain, `sw.js` updated, no tool attribution. `MIGRATIONS` is purely additive.

mclgoerg:
Thanks for the review. Both blocking points fixed, plus all should-fix and nice-to-have items, at `ef6673e7`. Migration renumbered to **219** (following #1255's 218, per the order confirmed there).

**Blocking**
1. Same rebuild-vs-foreign-trigger hazard I found while fixing #1255: this migration also rebuilds `reminders` (to widen the CHECK for `health_prevention_due`, since #1255's migration only widens for `document_expiry` now - see below), so it needed the same drop-triggers-before/recreate-after fix.
2. Added the missing test for the hand-written visibility check in `GET /prevention/due` (Bob/Carol/Dave, family vs. private on the latest record per type) and counter-proved it: disabled the check, confirmed the test goes red, restored it.

**The caregiver-consent decision**: went with the cycle model, as you leaned toward. `health_care_grants` now only governs read/write access to the data; a separate opt-in (`health_prevention_notify_caregivers`, off by default, toggle in Settings → Personal → Health) governs whether a caregiver gets pushed a name and item type at all. Takes effect immediately, same as a grant change already does. All six existing D6 fan-out tests updated to opt in explicitly (otherwise they'd have gone red against the new default), plus two new tests: no fan-out without opt-in despite an existing grant, and revoking the opt-in tears down an already-inherited row immediately.

**Should fix** - all applied:
- `visibilityScopes()` in `personal-health.js` gained the missing `prevention` row (the server-side scope already existed, nothing could ever reach it). Added a source-level guard test comparing `FLAT_SCOPES` keys against the settings sheet's rows so this can't silently regress again.
- The hand-written care-grant query in `prevention.js` replaced with `helpers.js#canWriteFor()`.
- Fixed the `docs/test-suites.md`/`test-health-structure.js` stale "62 routes" - turned out the real number was 74 even before this PR (an existing baseline drift, not something this PR introduced), corrected together with the test name and docstring.
- `docs/SPEC.md`'s claim corrected: `health.js` now imports the shared `DATE_STATUS_ALERT_DAYS` threshold (was a second hardcoded `30`), not the whole `dateStatus()` function - said precisely now.
- Type intervals entered in years now display in years in the admin list.
- Clearing an explicit due-date override now actually clears it (was silently a one-way ratchet).
- The in-app toast now names the cared-for person on an inherited row, matching the push notification.

**Nice to have** - all applied: no more per-entry re-query in the due-list filter, prevention rows got their own CSS classes instead of borrowing the activity tab's, two inline `style=""` replaced with classes, and a due-today item reads "Due today" instead of "Due in 0 days".

Left `health_prevention_types.key` as-is rather than removing it - happy to reconsider if you'd rather it go, but a stable per-household slug independent of the (renamable) display name seemed like a reasonable thing to keep even unused today.

Full `npm test` green.

ulsklyc (changes requested):
Round 2. Head reviewed: `6c9d7cbc` (you pushed again after your comment - `ef6673e7` and
`cd89db3f` are gone, so I read the current head). Reviewed only this PR's own four commits
(`pr1255..pr1256`), not the whole stack.

Both round-1 blockers are genuinely fixed, and I reproduced both rather than taking your word for
it. The caregiver consent decision is the right call and it holds up under mutation.

**What I verified myself**

- *Blocker 1.* Built the chain to 218, seeded a task + reminder + `notification_deliveries` row,
  then ran 219 three ways. As written: applies cleanly. With the two `DROP TRIGGER` lines and the
  re-creates removed: `error in trigger trg_reminders_tasks_ad: no such table: main.reminders`,
  exactly the crash. With the drops kept but the re-creates removed: it applies, both triggers are
  gone, and deleting a task leaves an orphan - the silent data-loss path, which
  `test:reminder-orphans` does catch. After the real migration both trigger bodies are identical to
  the v217 originals, all four indexes are back, the `notification_deliveries` row survived the
  `DROP TABLE`, and nothing in `sqlite_master` still names `reminders` unrestored.
- *Blocker 2.* Removing the filter line in `prevention.js` turns your new test red (15/16), and the
  test does exercise the youngest record per type in both directions. Replacing the hand-rolled
  care-grant query with `canWriteFor()` is correct: it reads the same table as
  `careAwareClause()`'s caregiver branch with no read/write distinction, so `/due` and `/records`
  now draw the same line.
- *Consent.* Default-off is real (`?.value === '1'` on a missing key), and the `sync_config` key the
  service reads is exactly what `cfgUserSet` writes, so the test seeding is not bypassing the
  route. Two mutations - forcing the gate open, and dropping the cleanup line - each turn two of
  your D6 tests red.
- *Also confirmed:* the `FLAT_SCOPES` guard goes red when the `prevention` row is removed; year
  intervals render as years; clearing `next_due_on` really does write NULL.

37/37 suites green on the head; 36/37 on a real merge with current main. The one failure is the
blocker below.

**Blocking**

- `test/test-health-structure.js:158` - `assert.equal(actual.length, 74);` replaced the previous
  `assert.equal(actual.length, EXPECTED.length)`. Current main carries 66 expected routes; plus
  your nine that is 75. Measured on the merge: `actual: 75, expected: 74`. `npm test` is a single
  `&&` chain, so nothing after it runs. Please put `EXPECTED.length` back - the `deepEqual` above
  already pins the exact table, and the derived form survives main moving.

**Should fix**

- `docs/test-suites.md:79` still says "62-Routen-Tabelle" where the test claims 74 (75 after the
  merge). You were right that the number was already stale on main (47 vs a real 66) - it is just
  wrong in a new way now rather than fixed.
- Three places still say "migration 218" after the renumber: `docs/SPEC.md` (twice) and
  `docs/test-suites.md:115`. It is 219.
- The PR body says "Migration 218 adds `health_prevention_types` ..." and "one commit on top of
  it" (it is four now). The title and body become the squash commit message.
- PR body line 3 has an em-dash ("merge that first — this branch ..."). CONTRIBUTING.md:338 covers
  commit messages, and this text becomes one.
- On your `key` question: please drop it. Nothing reads it - the only consumer is
  `prevention-due.js:61` passing it through as `type_key`, and `type_key` has no readers at all.
  The reason recorded in `modules-health.js:44` and `:232` ("Sichtbarkeits-Scope") is not right
  either; the visibility scope key is the literal `'prevention'` in `FLAT_SCOPES`. Because
  `MIGRATIONS` is append-only, keeping it makes a `NOT NULL UNIQUE` column permanent, along with
  `vKey`/`KEY_PATTERN`, the `key` branch in PATCH, and the client's `slugifyKey()` plus its
  409-retry-with-random-suffix dance. Removing it later costs a table rebuild.
- No test for clearing `next_due_on`. The path is right, but a regression back to sending
  `undefined` - the original one-way street - turns nothing red. A PATCH test that sends `null` and
  then expects the interval-derived due date is enough.
- No test for the caregiver branch of `GET /prevention/due`. Owner and uninvolved member are
  covered; the case that carries the feature's promise - `canWriteFor()` true, so Carol sees Bob's
  *private* due item - is not exercised, and Carol already exists in the suite.
- Merge note, not a finding against your code: the `CHANGELOG.md` conflict is sharp. Main inserted
  a `### Changed` heading between the two sides, so the obvious "keep both" resolution lands your
  Prevention entry *and* the Documents entry under `### Changed` instead of `### Added` - and I
  measured that `test:changelog` stays green on that. Both entries need to go above the
  `### Changed` line.

**Nice to have**

- Migration 219 uses `DROP TRIGGER trg_reminders_tasks_ad;` while v218 in #1255 uses
  `DROP TRIGGER IF EXISTS` for the same two triggers. Worth matching.
- `docs/SPEC.md` says "seven deep-link routes" and the list omits Fasting; `HEALTH_ROUTES` has
  eight. Older drift (main says "six"), but this PR edits that exact sentence. Same for the
  "alle sechs Routen"/"alle sechs Panels" comments in `public/utils/health-tabs.js`.
- `public/locales/hi.json`: `health.prevention.dueInDays` has base identical to `_one`. Its two
  closest neighbours put the oblique plural in the base - `documents.expiringInDays` and
  `inventory.trackedDateInDays` both use "दिनों". `overdueDays` already matches its neighbour.
- `prevention.js:179` overwrites a per-record `name` when its type is deleted. The UI never sends
  both, but `/api/v1` accepts both; `AND (name IS NULL OR name = '')` would keep it.
- The `FLAT_SCOPES` guard collects client keys with `/key: '(\w+)'/g` over the whole file, so any
  unrelated `key: '...'` in `personal-health.js` could satisfy a future scope by accident. Worth
  scoping the match to the `visibilityScopes()` block.

Heads-up on state: main has moved twice during this review (now `5b44508cf`). You will get
conflicts in `CHANGELOG.md`, `package.json` and all 24 locale files. That is us moving, not
anything you did. Migration 219 is still free, and `server/db.js` merges cleanly.

mclgoerg:
Round 3 fixes pushed:

**Should-fix:**
- Dropped `health_prevention_types.key` entirely - it was write-only (nothing read it except `prevention-due.js` passing it through as `type_key`, which itself had no readers). Removed the column, `KEY_PATTERN`/`vKey()`, the key branch in `POST`/`PATCH /prevention/types` (and their now-dead UNIQUE-violation 409 handlers), the client's `slugifyKey()` plus its 409-retry-with-random-suffix dance, and `type_key` from `computeDueForUser()`'s returned shape. Updated the tests that were sending `key` in request bodies.
- Added the two missing tests: PATCH clearing `next_due_on` via explicit `null` falls back to the interval-derived due date, and a real caregiver (an actual `canWriteFor()`-true relationship) sees a subject's `private` due item via `GET /prevention/due?user_id=`.
- Left the CHANGELOG issue-reference point as-is per our own prior call: this was fork-initiated work without a prior issue; happy to open one if you'd rather have it than not.

**Nice-to-have:**
- Type delete no longer overwrites a record's `name` snapshot if one is already set (`AND (name IS NULL OR name = '')`).
- Scoped the `FLAT_SCOPES` guard test's regex to just `visibilityScopes()` instead of the whole settings module.
- Fixed `hi.json`'s `dueInDays` base value to the oblique plural, matching its neighbors.
- Fixed the stale "seven deep-link routes"/"alle sechs Routen"/"alle sechs Panels" wording now that Fasting makes it eight.

PR body updated: migration number (218 -> 219), commit count (four, not one), and the em-dash fixed to a hyphen.

`npm test` is green.

ulsklyc (commented):
Round 3. Head reviewed: `037efb20`, rebased onto main `5b44508cf` and #1255 at `4dc5c7c4`. The range-diff against `6c9d7cbc` shows commits 2-4 unchanged and commit 1 differing only in rebase context, so I read `037efb20` itself plus the conflict resolutions.

Every point from round 2 is fixed in the code, and I checked the ones that carry weight by mutation, not by reading.

**What I verified**

- *Blocker.* `test/test-health-structure.js:164` is back on `EXPECTED.length`; `test:health-structure` is green with main's routes plus your nine.
- *`key`.* Gone everywhere: migration 219, POST/PATCH, `vKey`/`KEY_PATTERN`, the 409 branches, `type_key`, `slugifyKey()` and the retry. No leftover reader anywhere.
- *`next_due_on` -> null* (`test/test-health-prevention.js:287`): making the PATCH branch ignore a falsy value turns exactly this test red.
- *Caregiver branch of `/due`* (`:311`): narrowing `!canWriteFor(viewer, personId)` to `viewer !== personId` turns exactly this test red.
- *`FLAT_SCOPES` guard* (`test/test-health-visibility-defaults.js:167`): red when the `prevention` row goes, and still red with a decoy `key: 'prevention'` placed outside `visibilityScopes()`. The narrowing does its job.
- *CHANGELOG.* Both entries sit under `### Added`, above `### Changed`; no main line lost.
- Docs now say migration 219 and eight routes, `hi.json` is fixed, and the trigger drops use `IF EXISTS`.
- Full `npm test` on the head (which already contains current main): exit 0, all suites through the end of the chain. CI green on 22.x and 24.x.

**Should fix**

Both of these are things I missed in the earlier rounds, not problems your changes introduced. Sorry for bringing them up this late.

- `docs/SPEC.md:3627-3644` still describes the caregiver fan-out without the opt-in. `health_prevention_notify_caregivers` (default off) appears nowhere in the SPEC, and `:3642` still calls the fan-out "a deliberate, accepted trade-off". That is the framing we replaced in round 1. The SPEC is the reference for the data model, so it should say that the owner opts in, that the default is off, and that turning it off removes inherited rows at once. `docs/test-suites.md:85` could list the two opt-in tests too.
- `server/db.js:8830`: `health_prevention_types` has no `updated_at`, although `PATCH /prevention/types/:id` exists. CONTRIBUTING.md:393 asks for it on new entity tables, and the two closest household registries (`waste_types`, `schedule_shift_types`) both carry it. It costs one line now. Once 219 is on main, it takes a new migration, and `ALTER TABLE ADD COLUMN` cannot take the `strftime(...)` default. It is the same reasoning as for `key`.

**Nice to have**

- `server/routes/health/prevention.js:161`: the new `AND (name IS NULL OR name = '')` has no test. Removing it leaves the suite green. A record with its own `name` and a `type_id`, then deleting the type, then checking that `name` is unchanged would pin it.
- PR body: "four commits on top of it" is five now.
- CHANGELOG reference: leaving it as it is is fine, no issue needed for fork-initiated work.

Once the two should-fix items are in, this is ready from my side. It still merges after #1255.

One heads-up for #1257, since it is stacked on this one: it still sits on `6c9d7cbc`, and the migration 219 in that stack still creates `health_prevention_types` with `key TEXT UNIQUE NOT NULL`. When you rebase #1257, please take 219 from `037efb20` as it is now; if the old version survives a conflict resolution, creating a type fails again with the column the routes no longer write.

mclgoerg:
Rebased onto `main` now that #1255 is merged - the "stacked on" callout is gone from the PR body since the diff is standalone again (6714/125... just the prevention feature: 63 files, +4714/-125).

Round 4 (should-fix) fixes:
- `health_prevention_types` gets an `updated_at` column + `trg_health_prevention_types_updated_at` trigger, same shape as `waste_types`/`schedule_shift_types`. Migration 219 is still unreleased, so this is a plain column add rather than a follow-up migration.
- `docs/SPEC.md`'s caregiver fan-out section now leads with the opt-in: `health_prevention_notify_caregivers` (default off) is what gates an inherited row, not the care grant alone, and turning it off removes an inherited row immediately. Dropped the "deliberate, accepted trade-off" framing that predated the opt-in. `docs/test-suites.md` now names the two opt-in tests too.

Nice-to-have:
- Added a test pinning the type-delete name-overwrite guard (a record with its own `name` keeps it after its type is deleted).

`npm test` is green on the rebased head.

mclgoerg:
Rebased once more onto current `main` (two more PRs landed while this was in review, including a reminder-origin extraction that `health_prevention_due`/`assigned_from` needed folding into) - no functional changes from the last round, `npm test` is still green on the rebased head.

ulsklyc (approved):
Round 4, reviewed at `905a399a`, which is rebased onto `main` at `59c02232` and so already includes #1285.

Both should-fix items from round 3 are closed, and the nice-to-have too. I checked each one against the code. Where there is a test, I also broke the fix to see the test go red:

- **SPEC opt-in.** `docs/SPEC.md:3627-3634` now opens the fan-out section with the opt-in: `health_prevention_notify_caregivers`, default off, is the gate, and turning it off removes an inherited row at once. That matches `server/routes/preferences.js:1064-1072` and `server/services/prevention-reminders.js:156-165`. The "accepted trade-off" wording is gone, and `docs/test-suites.md:85` names both opt-in tests.
- **`updated_at`.** `server/db.js:8907-8910` adds the column and trigger in the same shape as `waste_types`. On a freshly migrated database, an `UPDATE` moves `updated_at` and leaves `created_at` alone.
- **Type-delete name guard.** Removing `AND (name IS NULL OR name = '')` from `server/routes/health/prevention.js:161` turns exactly your new test red (`test/test-health-prevention.js:235`, 17/18).
- **The #1285 fold-in.** `health_prevention_due` is in `ORIGIN_MODULE` (`server/services/reminder-origins.js:32`). A household that switches Health off gets no push and no `/pending` entry for it. `test:disabled-module-reminders` is green on this head, and dropping that one line turns its map guard red. The delivery query keeps main's `r.entity_id` next to your `r.assigned_from`, so neither side lost a column.
- Full `npm test` on this head: exit 0. CI is green on Node 22 and 24.

**Nice to have.** Neither is worth another round on its own:

- No test pins `updated_at`. Removing the column and trigger leaves `test:health-prevention`, `test:prevention-reminders` and `test:schema-mirror` green. The test below is green on your head, and red both without the trigger and without the column. It has to sit above the teardown test at `test/test-health-prevention.js:351`, because that one closes the server:

  ```js
  test('PATCH on a type moves updated_at, not created_at', async () => {
    const id = db.prepare(`INSERT INTO health_prevention_types (name, kind, created_at, updated_at)
      VALUES ('Masern', 'vaccination', '2000-01-01T00:00:00Z', '2000-01-01T00:00:00Z')`).run().lastInsertRowid;
    asA();
    const res = await call('PATCH', `/prevention/types/${id}`, { name: 'MMR' });
    assert.equal(res.status, 200);
    assert.notEqual(res.body.data.updated_at, '2000-01-01T00:00:00Z');
    assert.equal(res.body.data.created_at, '2000-01-01T00:00:00Z');
  });
  ```

- The SPEC's "removes it at once" for the opt-in depends on the sync call at `server/routes/preferences.js:1069`, and no test goes through that route. The opt-in tests write `sync_config` and call the sync themselves (`test/test-prevention-reminders.js:177-178`). If that call is dead-set, `test:prevention-reminders` and `test:preferences-routes` stay green. The 60-second pass still cleans up within a minute, so this one is optional.

`main` has moved since I measured: three PRs landed. The only conflict is the test line in `package.json`, and I'll resolve it when merging, so there is no need to rebase for it.

From my side this is ready to merge. It goes in before #1257.

mclgoerg:
Added the nice-to-have test pinning `updated_at`: PATCH moves it, `created_at` stays put. `npm test` is green. Thanks for the approval - this one's ready to merge whenever it's convenient (before #1257, per the stack).

mclgoerg:
Rebased once more onto current `main` (three more PRs landed) - no functional changes since the last comment. `npm test` is green on the rebased head.

ulsklyc (approved):
Re-reviewed at head `78241d39`, measured against my last approval at `905a399a`.

The rebase is clean. Diffing the two heads file by file, the only file that differs beyond what
main itself changed in `59c02232..aedf80d8` is `test/test-health-prevention.js` - the new test
commit. Nothing was silently edited in the PR's own files while merging the three new main commits.

**The nice-to-have test holds up.** `test/test-health-prevention.js:351-360` is real coverage, not
a comment-credited one: removing the `trg_health_prevention_types_updated_at` trigger from
migration 219 turns it red, and the suite builds its database from the real `MIGRATIONS` in
`server/db.js`, so it sees the actual migration rather than a mirror.

**The three new main commits are handled.**

- #1294's write-access drift guard scans every API path `public/` sends. Your routes all sit under
  the `health` prefix, which `server/scopes.js:33` already owns, so no new mapping was needed -
  `test:module-write-access` and `test:module-readonly-ui` are green on the merged head.
- #1285's `ORIGIN_MODULE` is fed correctly: `server/services/reminder-origins.js:32` carries
  `health_prevention_due`, and the value is in `VALID_ENTITY_TYPES`, `DERIVED_ENTITY_TYPES` and the
  migration's CHECK list. `test:disabled-module-reminders`, which holds both lists against that map,
  passes.
- #1288 and #1286 touch no migration, so there is no collision there.

**Migrations.** Main tops out at 218; this PR takes 219 and nothing else.
`test:migrations-append-only` and `test:schema-mirror` pass on the merged head. Worth knowing for
the queue: #1243, #1179 and #1180 also currently claim 219, so whichever of those lands after this
one has to renumber - that is their problem, not yours. Migration 219 dropping and recreating
`trg_reminders_tasks_ad` / `trg_reminders_events_ad` around the `reminders` rebuild is the right
call, and those two are indeed the only triggers in the schema whose body names `reminders`.

One thing I checked deliberately because it is a shape that usually leaks: the inherited caregiver
reminder has `created_by = caregiver` and `assigned_from = subject`, so it does not cascade when the
subject's user row is deleted. I drove it - the row does survive the delete with `assigned_from`
nulled, and then `syncAllPreventionReminders()` removes it, because
`server/services/prevention-reminders.js:184-188` sweeps `entity_id NOT IN (SELECT id FROM
health_prevention_records)`. That sweep is exactly the right shape and it is wired in at
`server/services/notifications.js:453`. Good.

Everything else I ran on the merged head is green: the structural guards (append-only, suite-chain,
schema-mirror, changelog, i18n, i18n-plural, i18n-translated, frontend-audit, typography,
settings-admin-gate, openapi-coverage, permissions) and every health and reminder suite including
the neighbours this feature does not touch.

## Nice to have

- `server/routes/preferences.js:1063-1071` - the opt-in route branch still has no test. The
  behaviour behind the flag is well covered, but `test/test-prevention-reminders.js` writes
  `sync_config` directly at `:71` and `:177`, so nothing exercises the route itself: not the
  `boolean` type check that returns 400, and not the immediate
  `syncPreventionRemindersForSubject()` that makes the switch "take effect now" rather than on the
  next periodic run. Same argument you accepted for the `updated_at` column - the code could be
  deleted and every suite would stay green.

## Verdict

approve

PR #1257, opened 2026-09-18
feat(inventory): recurring tracked dates, a service log, and a vehicle odometer

mclgoerg:
Adds recurring tracked dates, a service-history view, and a manual odometer for vehicles to Inventory.

### What changed

- A tracked date (TÜV, boiler service, chimney sweep, extinguisher check) can now carry a recurring interval in months: pressing "Done" on a due row writes a service-log entry (label/date snapshot, optional odometer/vendor/note) and rolls the date forward instead of just clearing it - the same row keeps its id, so its reminder and ICS UID stay stable and only move date. Without an interval, "Done" records the completion and removes the one-off deadline, same as before.
- A new read-only **History view** merges service-log rows with linked maintenance/accessory budget entries and linked documents into one dated timeline with a cost total - no new store, and both existing visibility rules (budget personal/shared mode, document sharing) run unchanged.
- A manual odometer reading (km/mi + reading date), gated by a new `inventory_categories.tracks_odometer` flag (seeded on for the built-in `vehicles` category) instead of a hardcoded category comparison - a household-renamed `vehicles` category keeps the flag through its stable key. There is currently no UI to flag a household's own category for tracking, and a category that had the flag has no way back once it's deleted. A tracked date can add a distance interval as a hint next to its date ("1,400 km to go"); a distance interval never produces a reminder, since the app has no way to know in advance when an odometer will cross it. The History view plots a small odometer trend chart above the timeline, reusing the shared chart geometry already used by Health/Budget.
- Migration 222 adds `interval_months`/`interval_distance` to `inventory_item_dates`, `odometer`/`odometer_unit`/`odometer_on` to `inventory_items`, `tracks_odometer` to `inventory_categories`, and the new `inventory_item_service_log` table.

### Test plan

- [x] `npm test` green locally
- [x] New suite: `test:inventory-service-log`


ulsklyc (changes requested):
Good work, and the parts that usually go wrong here are right: `addMonthsClamped` is pure UTC key arithmetic with no "today" in it, `dateStatus` goes through `parseLocalDateKey`/`todayKey`, and the new suite is registered, hung into the chain and described in `docs/test-suites.md`. Four things have to change before it can go in.

Reviewed against `e5601659`, with `main` at `a7380a9c9`. Since this stacks on #1256 and #1255, I judged your own four commits (`cb66dfd0..e5601659`); the rest belongs to those reviews.

## Blocking

**1. `server/db.js` - the migration numbers.**

The stack is built against a `main` whose highest migration was 216. Current `main` carries `217` since #1258, so the diff against `main` contains **two** entries with `version: 217` and `npm run test:migrations-append-only` goes red. Your 219 collides as well once #1255 and #1256 take 218/219.

Next free is 218. Renumbering across three stacked PRs is mine to coordinate: tell me the merge order you want and I will assign the numbers. Your 219 itself is cleanly appended and touches no existing entry.

**2. `server/routes/inventory/service-log.js:129` (`maybeAdvanceItemOdometer`) - an odometer is not monotonic here.**

Only the date is compared: `if (item.odometer_on == null || performedOn >= item.odometer_on)`. A newer entry with a *lower* value (5120 typed for 51200) silently overwrites `inventory_items.odometer`, and `odometerChartPoints()` draws the drop as a trend. The suite only covers the backdated case ("rueckdatierter Service-Log-Kilometerstand"), not the lower value at a newer date.

What makes it worse: log rows can be neither edited nor deleted from the UI, so the wrong value is not correctable by the user at all. `PUT /items/:id` also accepts any smaller value without a word (`items.js:263 ff.`).

**3. `server/routes/inventory/items.js:263` - the vehicle branch hangs on a user deletable key.**

`if (values.category !== 'vehicles')` nulls odometer, unit and reading date. But categories come from the user editable `inventory_categories` table (`items.js:89 validCategoryKeys()`), and `DELETE /api/v1/inventory/categories/:key` only protects `other`. Delete the "vehicles" category and its items move to `other`; the next item save then wipes the odometer without a word. A self created vehicle category ("motorbike", "camper") never gets the field at all.

Limiting it to vehicles is a fine decision, the mechanism is not: it belongs as a property on the category row, not as a literal in the validator.

**4. `public/pages/inventory.js:723`, `:795`, `:852`, `:981` - the unit reaches the UI as a raw DB code.**

`unit: item.odometer_unit || 'km'` (723), `const odometerUnit = item.odometer_unit || 'km'` (893, into `historyOdometerValue` and the chart tooltip), `${p.value} ${unit}` in the chart `<title>`. The locales do translate it (`ru: "км"`, `ar: "كم"`, `zh: "公里"`, `el: "χλμ"`). Only `odometerDetailValue` (981) uses `t('inventory.odometerUnitKm'/'...Mi')` correctly, so a Russian user reads "1400 км" on one line of the same detail view and "осталось 1400 km" on the next. UI text goes through `t('key')`.

## Should fix

- **`server/routes/inventory/service-log.js:195`** - `addMonthsClamped(trackedDate.date, ...)` rolls forward from the **due date**, not from the completion date. A long overdue interval therefore rolls into another past date; your own suite records it (`'2000-02-15'`, "the rolled date is in the past too"). The row stays overdue after "done", no reminder is created (`syncTrackedDateReminder` bails on a past `remind_at`), the done button stays, and a second press writes a second log row with the same `performed_on`. It wants a loop until the date is past `performed_on`.
- **`server/routes/inventory/service-log.js:165` (`deleteServiceLogEntry`)** - deletes a log row without recomputing `inventory_items.odometer`/`odometer_on`. Delete the newest reading and the item odometer stays on a value with no source left. This is the missing counterpart to `maybeAdvanceItemOdometer`.
- **`server/routes/inventory/service-log.js:242`** - `date: doc.created_at` is the timestamp of the *link* in `%Y-%m-%dT%H:%M:%SZ`, mixed into a timeline of plain day keys and handed to `formatDate()`. A file attached in the evening shows up on the neighbouring day depending on the household zone, and it is not the document's date anyway. At least cut it to the day key.
- **`public/pages/inventory.js:1004`** - `catch (err) { console.error(...); }` leaves `history = null`, so a failed `/history` call makes the whole history section disappear. "Loading failed" then looks exactly like "there is no history".
- **`server/routes/inventory/items.js:643`** - `PATCH /:id/service-log/:logId` uses `validateServiceLogInput`, so it requires `label` and `performed_on` and clears every omitted field. That is PUT semantics under a PATCH verb, and the OpenAPI description does not say so.
- **`CHANGELOG.md`** - "Every completion (and any manually logged repair) is kept in a new service-history view" promises a UI that does not exist: the frontend only calls `/complete` and `/history`. `POST`/`PATCH`/`DELETE /service-log` are API only. Either add the form or soften the sentence.
- **`docs/SPEC.md:2770 ff.`** - 11 new lines carry em dashes, plus an en dash in ``CHECK `1-600` ``. CONTRIBUTING.md:338: hyphens, not dashes. (CHANGELOG and locales are clean.)

## Nice to have

- `public/pages/inventory.js:981`, `:856` - odometer values are printed raw ("51200 km", chart axis `String(Math.round(val))`), while `getNumberFormat()` from `/i18n.js` is already used in budget.js, dashboard.js and housekeeping.js - and your own English placeholder writes "e.g. 15,000 km".
- `public/locales/de.json` `inventory.odometerLabel: "Kilometerstand"` carries the unit in the word although `odometer_unit` can be `mi`. "Zaehlerstand" would carry both.
- `public/styles/inventory.css:546` - `.inventory-odometer-hint` is dead, the class is set nowhere in `public/`.
- `public/styles/inventory.css:628` - raw `font-weight: 600` while the same diff uses `var(--font-weight-semibold, 600)` a few lines up.
- `trackedDatesDetailNode` - the done button only appears once `dateStatus()` is no longer `valid`, so at most 30 days before the date. An inspection done early cannot be recorded, and the manual log entry as a way out is missing too (see the CHANGELOG point).
- `server/routes/inventory/items.js:580` - `completeTrackedDate()` builds `{ trackedDate, logEntry }`, but the caller only reads `result.error` and answers with `loadItem()`. Two extra queries per completion for a value nobody uses.

mclgoerg:
Thanks for the review. All four blocking points fixed, plus should-fix and nice-to-have, at `5c65db86`. Migration renumbered to **220** (following #1255's 218 and #1256's 219).

**Blocking**
1. Renumbered.
2. The odometer is now monotonic against typos, not just backdating: an entry whose date would make it the item's current reading can no longer carry a lower value than the current one (400). A genuinely backdated entry stays unaffected since it never competes to become the current value - added a test for the exact typo scenario you described (5120 for 51200) plus the "genuinely backdated, still allowed" counterpart.
3. Moved the vehicles-only gate to a property on the category row (`inventory_categories.tracks_odometer`, seeded true only for `vehicles`) instead of a hardcoded string comparison. Deleting or replacing the category no longer silently breaks the feature - it now behaves exactly like deleting any other category (reassign to `other`, field clears on next save, same as switching any item away from a tracking category already did). One thing I did **not** do: build UI for a household to mark a *custom* category as odometer-tracking. The category editor is the shared cross-module `yuvomi-category-manager` component (also used by Budget), and it has no per-consumer extra-field mechanism - extending it felt like a separate, larger piece of work rather than something to fold into this fix. The data model supports it (a test creates a `camper` category with the flag set directly and confirms it works); happy to build the UI piece as a follow-up if you'd rather have it now.
4. Fixed at all four remaining spots via one shared `odometerUnitLabel()` helper instead of four separate call sites getting it right or wrong independently.

**Should fix** - all applied: the overdue-rollover loop (with a new test for a tracked date several intervals overdue), the missing delete-side odometer recompute (with a new test), the document-link day-key truncation, PUT instead of PATCH for the full-replace service-log update (no caller depended on the verb - the app only calls `/complete` and `/history` today), the CHANGELOG wording softened to match what's actually reachable from the UI, and SPEC.md's dashes/stale migration number.

**Nice to have** - all applied: `getNumberFormat()` for odometer display, the dead CSS class and untokenized `font-weight` removed, and `completeTrackedDate()` no longer does two unread re-fetches after writing.

Also found and cleaned a couple of leftover internal "Package C"/"Paket C" labels in CSS comments that had leaked into the shipped diff - harmless but shouldn't have been there.

Full `npm test` green.

ulsklyc (changes requested):
Thanks - the head I reviewed is `0191b566d`, not `5c65db86`; the whole stack was rebased at 14:24Z,
so I judged your five own commits (`f71f64ea..0191b566`) against `main` at `5b44508c`.

Three of the four blocking points are genuinely closed, and I verified them rather than reading them:
migration `220` appends cleanly and the merged chain reads 217/218/219/220 with no duplicate;
`odometerUnitLabel()` now covers all four display sites; and the tracked-date rollover, the
delete-side recompute, the document day-key, the `PUT` verb, the CHANGELOG wording and the SPEC
dashes are all in. `test:inventory-service-log` is green here (13/13), as are i18n, i18n-plural,
i18n-translated, frontend-audit, openapi-structure/coverage, changelog and migrations-append-only.

One blocker is still open, and it is the same one.

## Blocking

**`server/routes/inventory/service-log.js:61` - the typo guard switches itself off when the item
has a reading without a reading date.**

`odometerRegressionError()` returns `null` as soon as `item.odometer_on == null`, and
`maybeAdvanceItemOdometer()` (`:161`) overwrites unconditionally in that same case. The form allows
exactly that combination: `#inv-odometer` and `#inv-odometer-on` are independent fields, `saveItem()`
sends `odometer_on: null` when the date is left empty, and `validateItemFields()` does not require
it (it only defaults `odometer_unit` to `'km'`, `items.js:291`). Measured against this head:

```
item after create:   {"odometer":51200,"odometer_on":null}
POST /service-log:   201        (label "Inspektion", performed_on 2026-09-18, odometer 5120)
item after log:      {"odometer":5120,"odometer_on":"2026-09-18"}
```

That is round 1's scenario unchanged - 5120 for 51200, silently adopted, and the log row can be
neither edited nor deleted from the UI. With a reading date set, your guard works exactly as
described (400, "darf nicht unter den zuletzt erfassten Stand (51200) fallen"), and a genuinely
backdated lower entry is still accepted - I confirmed both.

Either fix closes it: default `odometer_on` to `todayKey()` (household zone,
`server/utils/timezone.js`) in `validateItemFields()` when a reading arrives without a date, the
same shape as the `'km'` default next to it - or treat `item.odometer_on == null` in
`odometerRegressionError()` as "always competing" instead of "nothing to compare".

## Should fix

- **`server/routes/inventory/service-log.js:203`** - `recomputeItemOdometer()` only looks at
  service-log rows, so it erases a reading typed into the item form. Measured: item created with
  `51200 / 2026-01-01` from the form, one log row at `52000 / 2026-05-01`, delete that row ->
  `{"odometer":null,"odometer_on":null}`. This is new in this round, i.e. the round-1 fix brought it
  in. Your own `odometerChartPoints()` treats "the item's own reading, if no log row represents it"
  as a first-class case; the delete path does not. API-only, hence not blocking, but it is silent
  data loss on the `/api/v1` surface. Only reset when the deleted row was the source of the current
  reading, and leave the item's own value alone otherwise.

- **`public/pages/inventory.js:1085`, `:1649`, `:1695`, `:1726`, `:1876`, `:1897`** - the client
  still holds the `'vehicles'` literal in six places while the server now reads the column. `:1876`
  and `:1897` are the harmful pair: they send `odometer: null` for any category other than
  `'vehicles'`, so saving the form wipes a value the server just accepted for a category you flagged.
  Your test proves the API path; the UI path does the opposite. `state.categories` (line 57) already
  carries `tracks_odometer`, so this is the same one-line predicate you wrote server-side.

  On the scoping question you raised: **not building a UI for a household to flag its own category
  is the right call** - `yuvomi-category-manager` is shared with Budget and has no per-consumer
  extra-field mechanism, and bending it here would be a bigger, separate change. What should not
  ship is the half state, where the data model says one thing and the form says another. Make the
  client read the flag; the editor UI can stay a follow-up. Deleting the category behaves exactly
  as you describe (reassign to `other`, cleared on next save) - the part worth writing down is that
  there is then no way back, since neither the UI nor `POST/PUT /inventory/categories` can set the
  flag again.

- **`docs/SPEC.md:2820` and `server/openapi/paths/inventory.js:51`/`:55`** - both still describe the
  old mechanism ("limited to the `vehicles` category", "silently cleared unless `category` is
  `vehicles`"), and `inventory_categories.tracks_odometer` appears in neither, although migration
  220 creates it. The code wins, so the stale line is the finding.

## Nice to have

- **`public/pages/inventory.js:1010`** - the one round-1 should-fix that did not make it: the `catch`
  only logs, `history` stays `null`, and the whole history section disappears. "Loading failed" then
  looks exactly like "there is no history".
- **`inventory.historyOdometerValue` / `trackedDateDistanceHint`** - `t()` interpolates `count` via
  `String(params.count)`, with no number format. So the detail row reads "51.200 km" while the
  timeline right below it reads "51200 km" and the hint reads "noch 15000 km". `getNumberFormat()`
  landed at two of four sites; format the value before the `t()` call and pass it as `{{value}}`.
- **`server/routes/inventory/items.js:672`** - `log.error('PATCH /:id/service-log/:logId error:')`
  in what is now the `PUT` handler.
- **`public/pages/inventory.js:966`/`:967`** - `gauge` now sits twice in a row, once for the
  odometer and once for the condition. And the "Done" button still only appears once `dateStatus()`
  is no longer `valid`, so at most 30 days ahead; with still no UI for a manual log entry, an
  inspection done early cannot be recorded at all. The CHANGELOG wording is correctly softened for
  that now, so this is a note, not a request.

## On merging (not a finding against you)

The branch is conflicting because `main` moved under it - that is on our side. When it is resolved:
`main` inserted a `### Changed` heading between the two conflicting CHANGELOG sides, so the obvious
"keep both in order" resolution drops the bullets below `### Changed`. The Inventory bullet belongs
under `### Added`, between the `#1178` entry and `### Changed`. `test:changelog` stays green either
way, so it needs eyes.

mclgoerg:
All four points from this round fixed:

**Blocking:** `odometerRegressionError()`'s self-disabling blind spot when an item has a reading but no reading date - `validateItemFields()` now defaults `odometer_on` to today's date (household zone) when a reading arrives without one, the same shape as the `odometer_unit` default right next to it. Confirmed your exact repro now rejects: creating an item with `odometer: 51200` (no date) then logging `5120` on the same day now returns 400 instead of silently adopting it.

**Should-fix:**
- `recomputeItemOdometer()` is now gated: it only re-derives `inventory_items.odometer`/`odometer_on` when the deleted log row's `(odometer, performed_on)` actually matches the item's current cached values - i.e. when that row was genuinely the source. An unrelated (e.g. backdated, never-current) row's deletion now leaves an independently-set value alone. Added a test for this exact case; the existing "delete the row that set the current reading" test still passes unchanged.
- The six hardcoded `'vehicles'` comparisons in `public/pages/inventory.js` now go through a `categoryTracksOdometer()` helper reading `state.categories`' `tracks_odometer` flag - including the two that were actively sending `odometer: null` for any category the server would now accept.
- `docs/SPEC.md` and `server/openapi/paths/inventory.js` now describe `inventory_categories.tracks_odometer` instead of the retired hardcoded-category mechanism, including the current gaps you flagged (no UI yet to flag a household's own category, no way back once a flagged category is deleted).

**Nice-to-have:**
- A failed history load now renders an error message instead of looking identical to "nothing recorded yet."
- `historyOdometerValue`/`trackedDateDistanceHint` now go through `getNumberFormat()` like the other two odometer display sites (all 24 locales updated, reusing each language's existing `tasks.historyLoadError` translation for the new `inventory.historyLoadError` key rather than a placeholder).
- Fixed the stale `PATCH` wording in the `PUT` handler's `log.error()`.
- The condition row no longer shares the `gauge` icon with the odometer row (now `sparkles`).
- Left the "Done" early/no-manual-entry note as-is, per your call that it's a note rather than a request.

Also rebased both this branch and #1256 onto current `main` (now includes #1255, and two more merges that landed while this was in review) - `main` had also drifted since your last look, so this includes conflict-resolution passes on `package.json`'s test chain, `CHANGELOG.md`, and a few shared reminder-origin files from an unrelated upstream PR (#1285's `withoutSwitchedOffModules`/`ORIGIN_MODULE` extraction), where `health_prevention_due`/`assigned_from` needed folding into the new shape.

`npm test` is green on the rebased head.

ulsklyc (commented):
Round 3, reviewed at `ff774ddf`. It sits on #1256 at `905a399a`, which sits on `main` at `59c02232`.

All four points from the last round are in. I measured them rather than only reading them:

- **Blocker.** `server/routes/inventory/items.js:307` now defaults `odometer_on` to today in the household zone. I ran my round-2 repro against this head. An item created with `odometer: 51200` and no date gets today as its `odometer_on`. Logging `5120` on that day returns 400, and the item stays at 51200. A backdated lower entry is still accepted, as intended. With the default dead-set, the same calls return 201 again and the item drops to 5120.
- **Delete path.** `server/routes/inventory/service-log.js:219-227`. Your new test at `test/test-inventory-service-log.js:429` goes red (13/14) when the gate is cut back to `if (item)`. The "source row deleted" test next to it still passes.
- **Client.** All six `'vehicles'` literals are gone. `categoryTracksOdometer()` at `public/pages/inventory.js:62` reads the flag, and `GET /inventory/categories` returns it through `SELECT *`.
- **Docs.** `docs/SPEC.md:2820-2837` and both OpenAPI summaries describe `tracks_odometer`, including the two gaps.
- **Nice-to-haves.** The history load error, `getNumberFormat()` for both strings in all 24 locales, the `PUT` log label and the `sparkles` icon are all in.
- Full `npm test` on this head: exit 0. CI is green on Node 22 and 24.

## Should fix

**The blocker fix has no test.** Setting `items.js:307` back to `values.odometer_on = vOdometerOn.value` leaves `test:inventory-service-log` (14/14) and `test:inventory-items-routes` (30/30) green. CONTRIBUTING.md:498 asks that a claimed fix come with a test that fails against the state before it, and this is the point that took two rounds. The test below is green on your head and red with the default dead-set. It reads the date the server stored, so it does not depend on the wall clock. Keep the `assert.ok`: without it, `performed_on: null` fails validation with a 400, and the test passes for the wrong reason.

```js
test('a reading without a date still arms the typo guard (Review #1257)', async () => {
  const created = await call('POST', '/items', { body: { name: 'Auto5', category: 'vehicles', odometer: 51200 } });
  const { id, odometer_on: readOn } = created.body.data;
  assert.ok(readOn, 'a reading without a date gets one');
  const typo = await call('POST', `/items/${id}/service-log`, { body: { label: 'Inspektion', performed_on: readOn, odometer: 5120 } });
  assert.equal(typo.status, 400);
  assert.equal((await call('GET', `/items/${id}`)).body.data.odometer, 51200);
});
```

**`inventory_item_service_log` has no `updated_at`** (`server/db.js:8987`, migration 220), although `PUT /:id/service-log/:logId` edits its rows. CONTRIBUTING.md:393 asks for the column on new entity tables, and the closest table on `main`, `housekeeping_maintenance_log`, has it. I missed this in both earlier rounds, sorry. It is the same point I raised on #1256: while 220 is unreleased, it costs one column and a trigger, and `health_prevention_types` in `905a399a` shows the shape. After a release it needs a new migration, and `ALTER TABLE ADD COLUMN` cannot take the `strftime(...)` default.

## Nice to have

- `server/routes/inventory/service-log.js:60`, called from `items.js:672`, checks a `PUT` on the log row that set the current reading against that row's own value. Measured: a log row with `520000` on 2026-09-10 (a typo upwards), then a `PUT` of the same row with `52000`, returns 400 with "darf nicht unter den zuletzt erfassten Stand (520000) fallen". The only way out is to delete the row and create it again. Checking a `PUT` against the reading as it stands without the row being edited would fix it. This is API-only for now, since the UI cannot edit log rows. In the UI, the way out after such a typo is the item form, and the `520000` stays in the history chart.
- `maybeAdvanceItemOdometer()` at `server/routes/inventory/service-log.js:159` does not check `tracks_odometer`. Measured: `POST /items/:id/service-log` with an `odometer` on an item in `other` sets `inventory_items.odometer` to 777, although `validateItemFields()` clears it on the next save. `/complete` goes through the same function. This is API-only too, since the completion sheet only asks for a reading when the category tracks one.
- The client predicate has no test either. Pointing `categoryTracksOdometer()` back at `categoryKey === 'vehicles'` leaves `test:inventory-category-labels`, `test:inventory-warranty-status`, `test:inventory-items-routes` and `test:frontend-audit` green. For page code I won't insist on one.

## On merging (not a finding against you)

This branch still carries #1256's own commits. Once #1256 is squash-merged, squash-merging this branch as it stands conflicts in `package.json` and `server/db.js`. I'll replay your six commits onto `main` instead. I tried that locally: they apply cleanly and give exactly this head's tree. Migration 220 stays as it is. `main` has also moved since I measured: three PRs landed. The only conflict from that is the test line in `package.json`, which I'll resolve when merging. You don't need to rebase for either of these.

Once the test and `updated_at` are in, this is ready from my side.

mclgoerg:
Should-fix items closed:

- **The blocker fix now has a test.** Added the exact test from the review, pinning that a reading without a date still gets one and still arms the typo guard.
- **`PUT` self-comparison.** Added `odometerBaselineExcluding()`: when the log row being edited is itself the source of the item's cached reading, the typo guard now checks against what the reading would be *without* that row (the next-most-recent remaining log row, or null), instead of the row's own pre-edit value. Correcting `520000` -> `52000` on the same row now succeeds; a genuinely-too-low value on that same row is still rejected. Added a test for both the fix and (implicitly, via the existing suite) that it doesn't loosen the guard for other rows.
- **`inventory_item_service_log.updated_at`.** Added the column + trigger, same shape as `health_prevention_types` and `housekeeping_maintenance_log`.

Nice-to-have also fixed: `maybeAdvanceItemOdometer()` now checks the item's category (`tracks_odometer`) before writing `inventory_items.odometer` - a log entry with an odometer value on a non-tracking category's item no longer sets it, closing the gap between the write path and the form (which already cleared it on save). Added a test for this too.

Left the client-predicate test as a "won't insist" per your note - happy to add it if you'd rather have it than not.

Rebased onto the refreshed #1256 (`59981068`, itself just a one-test addition on top of what you approved - no functional change). `npm test` is green on the rebased head.

mclgoerg:
Rebased once more onto the refreshed #1256 and current `main` (three more PRs landed there too) - no functional changes since the last comment. `npm test` is green on the rebased head.

ulsklyc (changes requested):
Re-reviewed at head `f7c6a454`, measured against my last round at `ff774ddf`. The stack sits on the
current #1256 head, so this is not an old copy of the predecessor - good.

**All three should-fix items from the last round are genuinely closed.** I reverted each one
individually against the head and ran the suite:

- putting `validateServiceLogInput(req.body, item)` back at `items.js:674` turns
  "PUT einer Log-Zeile prueft den Tippfehler-Schutz nicht gegen ihren eigenen alten Wert" red;
- removing `if (!itemCategoryTracksOdometer(itemId)) return;` at `service-log.js:173` turns the
  category-gate test red;
- changing `items.js:307` back to `vOdometerOn.value ?? null` turns "a reading without a date still
  arms the typo guard" red.

Each one was red against the state before its fix. `items.js:307` correctly uses
`todayKey(db.get(), new Date())` rather than a UTC slice.

The `'vehicles'` literal is gone from `public/pages/inventory.js` except for one comment at `:61`,
and both write sites I flagged last time (`:1896`, `:1917`) now go through
`categoryTracksOdometer()`. `/inventory/categories` returns `SELECT *`, so the flag is already on
the wire. That one is properly closed.

Everything green on the merged head: the structural guards, the five suites this PR adds, and the
eight neighbouring inventory suites. The new routes sit under the `inventory` prefix that
`server/scopes.js:29` already owns, so #1294's write-access drift guard needs no new entry, and the
PR introduces no new reminder origin - it writes the existing `inventory_tracked_date`, which is
already in #1285's `ORIGIN_MODULE`.

Migration numbering: main tops out at 218, #1256 takes 219, this takes 220. Clean as long as #1256
merges first.

## Should fix

- `server/routes/inventory/service-log.js:194-206` - `updateServiceLogEntry()` never writes the
  item's cached reading back. After the UPDATE it only calls
  `maybeAdvanceItemOdometer(itemId, values.performed_on, values.odometer)` at `:203`, which returns
  early on a null odometer and otherwise only writes when `performedOn >= item.odometer_on`. The
  DELETE path already guards exactly this at `:247-266` and calls `recomputeItemOdometer()` when the
  removed row was the source of the cached reading. PUT has no such guard. Driven against the head:

  Clearing the odometer on the only row that carried it leaves
  `inventory_items.odometer = 520000 / odometer_on = 2026-09-10` while the single log row now has
  `odometer = null`.

  Back-dating the leading row from `2026-09-10` to `2026-08-01`, with a second row at 50000 on
  `2026-09-01`, leaves the item at `520000 / 2026-09-10` - a date no row has any more, where the
  correct reading is 50000. The follow-up is the part that matters: a legitimate later entry of
  60000 on `2026-09-20` is then rejected with
  `400 "Kilometerstand darf nicht unter den zuletzt erfassten Stand (520000) fallen."`, and that
  includes the "Erledigt" sheet in the UI, because `validateCompletionInput(body, item)` reads the
  same `inventory_items` field. The control case - deleting the same row instead of editing it -
  recomputes correctly to `50000 / 2026-09-01`.

  Not blocking, because the UI never calls the service-log PUT or DELETE at all (the only item
  writes `public/pages/inventory.js` makes are `/dates/:id/complete`, `/entries` and item CRUD), so
  this is reachable only through the `/api/v1` surface, and a DELETE of the offending row recovers
  the state. But it is the same failure mode you just fixed on the validation side of this exact
  function, and the predicate already exists: `updateServiceLogEntry()` loads `existing` at `:195`,
  and `odometerBaselineExcluding()` at `:238-251` already computes "was this row the source". If it
  was, `recomputeItemOdometer(itemId)` belongs there instead of `maybeAdvanceItemOdometer(...)`,
  with the same guard the DELETE path uses against clobbering a reading set independently in the
  item form.

- `docs/SPEC.md:2823-2825` - "a household-renamed or deleted-and-recreated vehicle category, or a
  second one ("Motorcycle", "RV"), keeps or gains the flag through the category itself". The "gains"
  half is not true. The only writer of `tracks_odometer` is the migration
  (`server/db.js:9101`, `UPDATE ... WHERE key = 'vehicles'`); `POST /inventory/categories`
  (`categories.js:43-65`) does not set it and falls to the `0` default, and `PUT /:key` only writes
  `name`, `icon` and `label_key`. Renaming does keep the flag, because `key` never changes - that
  part is right. Three lines further down the same paragraph says the opposite and is correct
  ("there is currently no UI to flag a household's own category for tracking ... has no way back").
  Please drop the "or gains" claim so the paragraph stops contradicting itself; the one-way door
  itself is documented well and I have no objection to it.

## Nice to have

- `server/db.js:9126` and `:9129` - `inventory_item_service_log.updated_at` and its trigger are
  unobservable. `loadServiceLog()` (`service-log.js:139`) and `loadServiceLogEntry()` (`:148`)
  select an explicit column list that omits it, and no test names the column, so the column and the
  trigger could both be deleted and every suite would stay green. Same gap you closed on #1256 for
  `health_prevention_types` - either pin it with a test, or put it on the response the way the rest
  of the schema does.

- `public/pages/inventory.js:1103` - `const isVehicle = categoryTracksOdometer(item.category);`
  still reads `isVehicle` at `:1103`, `:1109` and `:1138`. The value is correct now; the name is the
  last trace of the literal the rest of the file moved away from.

## Verdict

request-changes

ulsklyc:
Status update, since the ground under this PR has moved since my last review.

**#1256 is merged**, as `bcc35edb7`. Your predecessor is in, so this branch no longer needs to carry
it. Since then `main` has also taken #1179, #1287, #1290, #1269 and #1289, and now sits at
`d01e6c024`.

## The restack

```
git rebase --onto origin/main 78241d39
```

`78241d39` was the last head of #1256, so this drops the predecessor commits and replays only your
own work onto current `main`.

Worth knowing before you start, because the conflict count looks worse than it is: GitHub currently
reports **nine** conflicting files, but I measured which of them are actually yours, and only three
are:

| conflicting | yours? |
|---|---|
| `CHANGELOG.md`, `package.json`, `server/db.js` | yes |
| `public/reminders.js`, `server/routes/preferences.js`, `server/routes/reminders.js`, `server/services/notifications.js`, `test/test-notifications.js`, `test/test-push.js` | no - #1256 and #1179 territory |

Those six disappear with the restack. They only conflict because the branch still replays the
predecessor, which `main` already has in squashed form.

## Migration: 220 is taken, yours becomes 221

I compared the numbers rather than assuming. Version 219 is identical on both sides, that is #1256's
and it is in. **220 is a genuine collision:** `main` has "Health: configurable fasting reminders and
reminder entity types" from #1179 there, and this branch has "Inventory: recurring tracked dates, a
service log, and a manual odometer". 221 is free.

Where 220 appears on your side:

- `server/db.js` (2 places)
- `server/routes/inventory/item-dates.js` (1)
- `docs/SPEC.md` (1)

Note that #1179 also rebuilt the `reminders` table and extended its CHECK list, so after the restack
please check that your tracked-date entity type is still in the list that survives.

`CHANGELOG.md` and `package.json` are additive on both sides. If you would rather not touch them, say
so and I will resolve those two at merge time, as I did for #1179 and #1180.

## Still open from my review

The restack does not change these, and my request for changes stands on them:

- `server/routes/inventory/service-log.js:194-206` - `updateServiceLogEntry()` never recomputes the
  item's cached odometer, while the DELETE path at `:247-266` does. Clearing or back-dating the
  leading log row strands `inventory_items.odometer` at a value no row carries, and because that
  field is the monotonicity baseline, a later legitimate entry is then rejected permanently,
  including through the completion sheet in the UI.
- `docs/SPEC.md:2823-2825` - the "or gains the flag" half is not true; only the migration writes
  `tracks_odometer`. Three lines down the same paragraph says the opposite, correctly.

No rush on my side, and thanks for the two careful rounds so far.

mclgoerg:
Rebased directly onto `main` now that #1256 merged - the "stacked on" callout is gone, migration is now 221 (main took 220 for the fasting-reminders/entity-type widen).

Should-fix items:
- **`updateServiceLogEntry()` write-path gap.** It only called `maybeAdvanceItemOdometer()` with the new values, so editing the log row that was the current source (clearing its odometer, or moving its date behind another row) left `inventory_items.odometer`/`odometer_on` pointing at data that no longer existed - and a later legitimate reading could then be rejected against that stale cache. Now mirrors the DELETE path's guard: when the edited row was the source, `recomputeItemOdometer()` runs after the write instead of just checking whether the new values individually win. Added both of your measured repros as tests (clearing the only row's odometer; back-dating the leading row so a second row becomes current, then confirming a later legitimate reading is no longer rejected).
- **SPEC contradiction.** Dropped the "or gains the flag" half of the claim - only the migration ever sets `tracks_odometer`, and the file already said as much three lines down. Fixed the same wording in the PR body.

Nice-to-have:
- `inventory_item_service_log.updated_at` is now on the wire (`loadServiceLog`/`loadServiceLogEntry`'s explicit column lists).
- Renamed `isVehicle` to `tracksOdometer` in the completion sheet - the literal restriction it named is gone.

`npm test` is green on the rebased head.

ulsklyc:
Thank you for the restack - it made a real difference. Before it, this branch still replayed seven
commits that `main` already carries as a single squash, and nine files came up as conflicting, six
of which were not yours at all. After it, exactly one file conflicted.

That one file was the `test` chain in `package.json`, and the commit `a2594cc49` on your branch is
mine: I merged `origin/main` in and resolved it, as I offered earlier. Your branch had 354 steps,
`main` has 355 (two suites arrived with #1295 today), and the only step unique to your side was
`npm run test:inventory-service-log` - so 356 in total, nothing dropped, no duplicates. Checked with
`test:suite-chain` (10/10) and `test:changelog` (34/34), and your own suite runs green at 19/19.
Everything else merged cleanly; I touched no other file.

I can see your four review-round commits, including the odometer recompute on the PUT write path.
The `changes requested` on the PR is still formally standing because I have not re-read the diff
yet - that is on me, not on you, and it is next on my list.

ulsklyc (approved):
Re-read the diff on `a2594cc49`. Both findings my `changes requested` stood on are closed, and I
measured them rather than reading your commit message.

## The blocker is closed, and it is genuinely tested

`server/routes/inventory/service-log.js:194-223` now mirrors the DELETE path: it captures whether
this row, with its *old* value, was the source of the item's cached reading, and after the write
either recomputes from all remaining rows or falls back to the plain advance.

I dead-stalled the fix to check the tests actually carry it. Turning
`service-log.js:216` into `if (false && wasSource)`, so the `else` branch reproduces the old
behaviour:

```
test:inventory-service-log   19 tests, 17 pass, 2 fail
✖ PUT raeumt inventory_items.odometer ab, wenn die einzige Log-Zeile ihren Kilometerstand verliert
✖ PUT rechnet inventory_items.odometer neu, wenn die bearbeitete Zeile die Quelle war
```

And red for the right reason, not incidentally: both assertions name the stranded value from my
report, `actual: 520000, expected: null` and `actual: 520000, expected: 50000`.

I did the same for the other half of the machinery, since a guard that was never seen red proves
nothing. Replacing the `odometerBaselineExcluding()` call at
`server/routes/inventory/items.js:674` with `const baseline = item;` turns the self-comparison test
red on its own. So the caller is covered too, not just the function.

`docs/SPEC.md:2827-2836` no longer claims a new or second category gains the flag, and it now agrees
with the sentence three lines down and with `server/db.js:9313`, the only place that writes
`tracks_odometer`.

Also verified against the restack: migration 221, no collision, and `MIGRATIONS` grows only at the
end (`git diff` on `server/db.js` is one hunk with no removed lines). `inventory_tracked_date`
survived #1179's `reminders` rebuild and is in the CHECK list that migration 220 leaves behind, so
there was nothing to fold. CI is green on both Node 22 and 24 on this exact head. I ran 27 targeted
suites locally, all green.

## Blocking

None.

## Should fix

**1. The renumber reached `server/db.js` and nothing else.** The migration is 221, but six places
still say 220:

- `docs/SPEC.md:2769` - "recurrence/service log/odometer in v220"
- `docs/SPEC.md:2791` - the completion endpoint
- `docs/SPEC.md:2809` - the service log table
- `docs/SPEC.md:2819` - the history view
- `docs/SPEC.md:2827` - the odometer
- `server/routes/inventory/item-dates.js:89` - "server/db.js Migration 220"

This is not cosmetic, because 220 is occupied in `main` by #1179's fasting reminders, and
`docs/SPEC.md` attributes it correctly to that migration in two other places
(`docs/SPEC.md:1700`, `docs/SPEC.md:3103`). After the merge the file says two different things about
the same migration number. Inside migration 221 in `server/db.js` itself the number is right, so it
really is only the prose that lagged. Nothing catches it: `test:migrations-append-only`,
`test:schema-mirror` and `test:schema-reconcile` are all green.

**2. `recomputeItemOdometer()` has no `tracks_odometer` gate, so it launders a reading the gate just
refused.** `maybeAdvanceItemOdometer()` checks the category at `service-log.js:173` (the gate you
added in round 3), `recomputeItemOdometer()` at `service-log.js:232-240` does not. I thought that was
unreachable and went to measure it, and it is reachable, because
`server/routes/inventory/categories.js:114` reassigns items with a bare
`UPDATE inventory_items SET category = 'other'` and leaves `odometer`/`odometer_unit`/`odometer_on`
standing. So an item can sit in a non-tracking category with a non-null cached reading, which is
exactly what makes `wasSource` reachable there:

```
1. item in 'vehicles', log row 100000 on 2026-09-10   -> inventory_items.odometer = 100000
2. household deletes the vehicles category            -> category becomes 'other', odometer stays
3. POST a new log row, 105000 on 2026-09-15
   -> odometer = 100000 on 2026-09-10                    the gate holds, correct
4. PUT the source row (110000 on 2026-09-10)
   -> odometer = 105000 on 2026-09-15, category = 'other'
```

Step 4 writes `inventory_items.odometer` on an item whose category does not track it, and it picks up
precisely the 105000 reading that step 3 had just rejected. The DELETE path has the same gap at
`service-log.js:280`. The damage is limited, the next item save clears all three fields again
(`server/routes/inventory/items.js:266-276`), so there is no data loss and no permanent rejection.
But it does undo the round-3 guard on a path that guard cannot see, and the fix looks like one line.

**3. PUT and DELETE can lower or clear the item's reading, and nothing says so.**
`docs/SPEC.md:2815-2818` and `server/openapi/paths/inventory.js:97-98` describe only the advancing
half, "advances `inventory_items.odometer`/`odometer_on` only when its `performed_on` is the newest
reading". Since this round that is half the story: both write paths can also move the cached reading
*down* or to NULL. The POST description at `server/openapi/paths/inventory.js:89` documents its half
properly, which is why the omission on PUT and DELETE stands out. An API consumer editing a log row
has no way to know it mutates a field on a different resource.

## Nice to have

**4. `updated_at` is on the wire but not in the spec.** `docs/SPEC.md:2814` still ends the column
list at "`created_by` (SET NULL) and `created_at`", while the column exists
(`server/db.js:9338`), a trigger maintains it (`server/db.js:9341`) and both load paths return it
(`server/routes/inventory/service-log.js:139` and `:148`). One clause.

**5. A question rather than a request: a backdated edit can leave the log itself non-monotonic.**
`odometerRegressionError()` at `server/routes/inventory/service-log.js:60-67` only compares against
readings at or before `performed_on`, never against a *later* row. Your own new test at
`test/test-inventory-service-log.js:513` therefore ends with the log claiming 520000 km on
2026-08-01 and 50000 km on 2026-09-01. I checked that this causes no lockout, the later 60000 entry
is still accepted in that same test, so it is not the class of bug I blocked on, and the trend chart
is the only place it shows. Since backdating is deliberately exempt from the guard, I suspect this is
a conscious trade and not a defect. If it is, say so and I will drop it. Some manual logs are simply
allowed to be wrong.

## Constraints

Checked against the diff, nothing to report: no `innerHTML`/`outerHTML`, user data through `esc()`
including every interpolation in the SVG chart at `public/pages/inventory.js:843-880` and the
accessible table at `:825-836`, UI text only through `t()` with all 24 locales carrying the new keys,
design values from tokens, `MIGRATIONS` appended only, no `await` before a DB call, and no
`toISOString().slice(0, 10)` day keys. The one truncation that looks like it,
`service-log.js:378`, cuts a stored UTC instant and explains itself in the comment above.

Merge state is clean, no conflict, and `origin/main` is already an ancestor. If `main` moves again
before this lands, please merge it in rather than rebase, so the resolved test chain in
`a2594cc49` survives.

## Recommendation

**APPROVE.** The blocker is closed with a counter-proof I ran myself, and nothing left can lose data
or lock a household out. Four rounds of careful work, and the odometer semantics are in better shape
than when this started. Please land items 1 to 3 before merge, or say the word and I will fold the
migration-number and documentation fixes in at merge time, as I did for `package.json` and
`CHANGELOG.md`.

mclgoerg:
Thanks for the approval and the counter-proof on the blocker - appreciated.

Should-fix items closed:

1. **Stale v220 references.** Fixed all six: the five `docs/SPEC.md` mentions and the `item-dates.js` comment now say 221, matching `server/db.js` itself.
2. **`recomputeItemOdometer()` needed the same `tracks_odometer` gate as `maybeAdvanceItemOdometer()`.** Added it, and added a test reproducing your exact repro: item in `vehicles` gets a reading, the category is deleted (simulated via the same bare `UPDATE ... SET category = 'other'` `categories.js` does), a new log entry is correctly rejected by the gate, and then editing the *source* row no longer launders that rejected value in through `recomputeItemOdometer()`. Good catch - I hadn't thought to check the DELETE-path category reassignment as a way to reach a "tracking category with a stale cached reading" state.
3. **PUT/DELETE can lower or clear the cached reading, not just advance it.** `docs/SPEC.md` and both OpenAPI descriptions now say so explicitly, alongside the "was this row the source" gate that scopes it to the row that mattered.

Nice-to-have:

4. `updated_at` is now in the `docs/SPEC.md` column list.
5. On the backdated-edit question: yes, this is a conscious trade, not a defect. `odometerRegressionError()` only guards forward-in-time competition for *becoming the current reading* - a manually-entered service log is historical record, and the household is trusted to get dates right when backdating (same trust extended to backdating with a correct value, which the guard already allows unconditionally). Catching "this later row now reads lower than an earlier one" in general would mean validating monotonicity across the whole log on every write, not just against the current cached reading, which is a bigger check than this feature set out to do. Since you confirmed it doesn't cause a lockout and only shows up in the trend chart, I'd rather leave it as-is unless you feel strongly otherwise.

`npm test` is green. Per your note, I merged `main` in rather than rebasing this time, so the resolved history from `a2594cc4` carries forward.

ulsklyc (changes requested):
Thanks for folding all six migration-number references in - I checked them at `927aafd7` and all six now say 221, and `docs/SPEC.md:1700`/`:3108` still correctly attribute 220 to the fasting migration. Point 5 is settled: you said it is a conscious trade, and I said I would drop it if you did. Dropped.

I read `527d7af5` on its own, since my approve stood on the merge before it. The gate is the one thing in it that changes behaviour, and I think it is pointed the wrong way.

## `recomputeItemOdometer()` should clear, not return

`server/routes/inventory/service-log.js:240`

```js
function recomputeItemOdometer(itemId) {
  if (!itemCategoryTracksOdometer(itemId)) return;
```

Your comment above it has the goal right: a PUT or DELETE on the source row must not wash a reading back in that the item has long since lost. But `recomputeItemOdometer()` is the clean-up function, and `return` switches off the clean-up along with the wash-in.

Measured, same run shape, only the gate swapped:

| step | gate live (`927aafd7`) | gate dead |
|---|---|---|
| category deleted, item falls back to `other` | 100000 stays | 100000 stays |
| DELETE the only row carrying a reading | **100000 stays, log empty** | NULL, healed |
| category back to `vehicles`, POST 60000 | **400, "must not fall below (100000)"** | 201 |

So the item keeps a reading whose source no longer exists, and that orphan is what the monotonicity validator compares against. It is the same lockout as before, through a different door - only now nothing can clear it, because the one function that used to is the one that returns early.

The fix is the shape you already use in this PR. `server/routes/inventory/items.js:273-276` handles exactly this case on every item save:

```js
if (!categoryTracksOdometer(values.category)) {
  values.odometer = null;
  values.odometer_unit = null;
  values.odometer_on = null;
}
```

with a comment saying why: moving away from an odometer-tracking category "clears a previously set value automatically instead of leaving it there invisibly". Doing the same in the non-tracking branch here stops the wash-in just as well and heals the orphan on the way. The two halves of the PR then agree instead of contradicting each other.

Your new test pins the frozen value (`test/test-inventory-service-log.js:579`, `assert.equal(..., 100000)`), so it moves with the change.

## The docs promise the clearing that the gate removes

`docs/SPEC.md:2816-2818` and `server/openapi/paths/inventory.js:97-98` both say, without qualification, that PUT/DELETE recompute the reading and can "clear it to `NULL` if no row carries a reading anymore". From `527d7af5` on, that is no longer true in the non-tracking case. Code and docs drifted apart inside one commit. If the gate clears instead of returning, both sentences become true again and nothing needs rewording.

## Nice to have

`docs/test-suites.md:128` describes `test:inventory-service-log` without the category gate; the entry did not come along in the delta.

## Everything else

Verified on the delta: `MIGRATIONS` grows only at the end (221 last, `server/db.js:9289`, no `-` line anywhere in the db.js diff), no `await` before a DB call, no `innerHTML`, no `toISOString().slice`, no locales or CHANGELOG touched. Suites all green: `inventory-service-log` (20/20), `inventory-items-routes`, `inventory-categories-routes`, `inventory-tracked-dates`, `inventory-item-reminders`, `inventory-deadlines-feed`, `inventory-category-labels`, `inventory-warranty-status`, `migrations-append-only`, `schema-mirror`, `suite-chain`, `openapi-structure`, `openapi-coverage`.

One thing for whoever merges, me included: main has released v2.68.0 since your merge, so the branch conflicts on `package.json` (one line, the test chain - the union is 356 steps, nothing lost). Worth knowing that an automatic resolution would slide your CHANGELOG entry into the already published `## [2.68.0]` section and leave `## [Unreleased]` empty. `test:changelog` catches it, so it is work rather than a blind spot, but it needs doing deliberately.

mclgoerg:
Good catch - you were right, the gate was pointed the wrong way.

`recomputeItemOdometer()` now clears (`odometer`/`odometer_unit`/`odometer_on` → `NULL`) instead of returning early when the category doesn't track odometer, mirroring `items.js#validateItemFields()`'s existing rule for the same situation. Verified against your exact table: category deleted → reading stays (unchanged, that's `categories.js`'s own gap, out of scope here) → DELETE the only row → now heals to `NULL` instead of freezing → category restored → a fresh legitimate reading is accepted (201), no more permanent lockout.

Added your DELETE-and-restore repro as a new test, and updated the existing PUT-edits-the-source-row test's frozen-value assertion (`100000` → `null`) since it now goes through the same clearing path.

Also fixed:
- `docs/test-suites.md`'s entry now mentions the category gate and this round's fixes.
- Merged `main` again (a release cut to v2.68.0 landed) - and this time actually caught the landmine you warned about: the auto-merge did slide the Inventory CHANGELOG entry into the already-published `## [2.68.0]` section (positionally, since it sat right next to the now-released fasting-reminders entry). Moved it back under `## [Unreleased]` → `### Added`; `test:changelog` is green.

`npm test` is green (327/327 suites).

ulsklyc:
Heads-up on the migration number, so it does not cost you a round: **221 is taken now.** `f90566def` (#1310, rewards quantity) landed on main with it, and a second one of ours is queued behind it for 222.

So when you push the odometer fix, your migration needs the next free number at that moment, not 221. `test:migrations-append-only` requires the numbers to be contiguous, so it will tell you plainly if the branch is behind - merge `main` first and let the guard name the number rather than guessing it.

Sorry about the shuffle. Your PR asked for 221 first; the two that overtook it were ready while this one was in review, and reserving a number with a placeholder migration is the one thing we cannot do - the runner records every number unconditionally, so a placeholder would permanently block the real migration on any installation that ever started that branch.

Nothing else changes about the review: the odometer gate is the only open point.

mclgoerg:
Merged `main` in again per your heads-up - renumbered to 222 (221 was taken by #1310's rewards quantity migration), fixed the five `docs/SPEC.md` v221 mentions and the `item-dates.js` comment to say 222, and updated the PR body. `test:migrations-append-only` and `test:changelog` are both green, and `npm test` is green overall (327/327 suites).

ulsklyc:
I have pushed the conflict resolution to your branch rather than send you another round for it - it was mechanical, and you have already done three.

**The migration number got taken out from under you, and that is on my heads-up, not on you.** I wrote that 221 was gone and that a second one of ours was queued for 222. You renumbered to 222 at 20:05; ours landed on main with 222 at 22:33, three hours later. So the number you were told about was still free when you took it and was not free when main moved. That is the part worth naming: **a migration number cannot be reserved.** `test:migrations-append-only` requires the numbers to be contiguous, so a branch sitting on 223 while main is at 221 goes red - which means the number can only be settled at the moment of the merge, by whoever merges. That is me, so I have taken it on.

What the merge commit does:

- `server/db.js` - main keeps 222 (`6e99a2dcc`, recipe/pantry matches), your inventory migration becomes **223**. The five `v222` mentions in `docs/SPEC.md` and the comment in `server/routes/inventory/item-dates.js` moved with it.
- `package.json` - the test chain as the union of both sides: main's `test:recipe-pantry-match` and your `test:inventory-service-log`.

Measured after the resolution, all exit 0: `test:migrations-append-only`, `test:schema-mirror`, `test:schema-reconcile`, `test:suite-chain`, `test:changelog`, and both sides' own suites - `test:inventory-service-log`, `test:inventory-tracked-dates`, `test:tracked-dates-migration`, `test:item-dates-validation`, `test:inventory-items-routes` and main's `test:recipe-pantry-match`.

Nothing of yours was dropped or rewritten; if you pull, take the merge as is rather than rebasing it away.

Your `e50088ec7` is the answer to the odometer finding and I am going through it now. I will come back on that separately.

ulsklyc:
Full review of the current head. **The odometer finding is properly closed**, and I have taken one blocker off your hands. Three things remain, two of them in code.

## Closed: the orphaned reading

`recomputeItemOdometer()` now clears instead of returning early, with the same rule `items.js#validateItemFields()` uses, and both entry points reach it. I verified it the way the house asks rather than taking the commit title for it: in a scratch tree at your head, 21 tests, 21 green; the same tree with only `service-log.js` reset to `527d7af5`, 21 tests, **2 red** - exactly `recomputeItemOdometer() waescht keine Ablesung ein...` and `DELETE der einzigen Ablesung heilt den verwaisten Stand ab...`. Same test count on both sides, so the tree was intact. That is a fix proven against a red test.

## Fixed by me: the odometer group was visible for every category

I pushed this one rather than send it back, because it is one CSS line and this file already carries the same fix twice.

`.inventory-form-row` sets `display: grid`. That is an author declaration and it beats the user agent's `[hidden] { display: none }` at equal specificity, so `#inv-odometer-group` stood in the form for **every** category - sofa, laptop, bicycle - although the markup sets `hidden`. `odometerGroup.hidden = ...` at `inventory.js:1745` had no effect either. A household could type a reading there, and `saveItem()` dropped it without a word, because `inventory.js:1896` clamps `odometerRaw` to `categoryTracksOdometer(category)`.

Measured in a browser against the real stylesheet, not read off the file:

| | `display` | height | hit-testable |
|---|---|---|---|
| without the rule | `grid` | 22px | yes, with `hidden` set |
| with the rule | `none` | 0px | no |

No guard sees this. `test:frontend-audit`, `test:module-write-access`, `test:detail-view` and your own suite are all green on your head. The repo has now fixed this same cascade five times by hand - `inventory.css` twice, plus `budget.css`, `housekeeping.css`, `category-manager.css` and `calendar.css`. That is our problem to solve, not yours; I am opening a separate issue for a guard.

## Still open - two in code

**1. `service-log.js:392` writes a UTC day into a local-day timeline.** `date: String(doc.created_at).slice(0, 10)`. `documents.created_at` is a UTC instant (`%Y-%m-%dT%H:%M:%SZ`), and the slice gives the UTC day, mixed into a timeline otherwise made of household-local day keys. A document linked at 01:30 on 22 September in Berlin is stored as `2026-09-21T23:30:00Z` and shows in the timeline under **21 September**, sorted there too. West of UTC it goes the other way.

The comment above the line argues for the slice - "rather than letting an evening link fall to the neighbouring day depending on the household zone" - but that is the reverse: truncating a UTC instant is what produces the neighbouring day. `server/utils/timezone.js:86-88` has the conversion already, the same one `todayKey()` uses: `utcToWall(iso, householdTimeZone(db))?.date`.

**2. The service log never writes `odometer_unit`.** `service-log.js:177` and `:252` both set `odometer` and `odometer_on` and leave the unit alone. A vehicle whose readings only ever arrive through "done" or the service log keeps `odometer_unit = NULL`, and `odometerUnitLabel(null)` at `inventory.js:996` answers with the **km** label. A household that counts in miles then sees km in the detail row, the history rows, the chart tooltip and the a11y table - and the completion card has no unit field to correct it with. The non-tracking branch of the same function does clear the unit, so the module is not of one mind about who owns it.

This one is a design question, not a typo, and it is yours: does the unit come from the item's existing value, from the category, or should the completion card ask? I have no opinion I would push on you.

## Still open - one in prose

`docs/SPEC.md:2844` promises `"1,400 km to go, last reading 12 Sep"`. The code renders `t('inventory.trackedDateDistanceHint')` = `"noch {{value}} {{unit}}"` (`inventory.js:729`); there is no "last reading" part anywhere. I can take this line at the merge if you would rather not touch it.

## Nice to have, not gating

- `service-log.js:343` - `do { ... } while (nextDate <= values.performed_on)` runs over user input. `interval_months: 1` with `performed_on: '9999-12-31'` turns roughly 95,000 times before it terminates. It does terminate; a computed jump would just be more honest than a loop over an untrusted bound.
- `inventory.js:1057` - `onDoneTrackedDate` awaits `openItemDetail(refreshed)` inside itself, so every completion nests another await that never settles until the page changes.

## Where it stands

Everything else is clean: no `innerHTML`, no `await` before a DB call, all 21 new keys in all 24 locales, changelog entry with a bold opening line, the suite registered **and** hooked into the `test` chain. The migration is renumbered to 223 across all seven places, and both sides' suites pass on the merge.

Push those two and I will merge. If you would rather I take them, say so and I will - you have done four rounds on this and I do not want a fifth to be about a unit label.

ulsklyc:
**The migration number moved again - 223 is now 224.** I pushed it to your branch rather than send you a fourth round, same as with the conflict yesterday. Nothing else on the branch changed.

Yesterday I wrote that 222 was taken and you renumbered to 223. Since then #1343 (nutrition, stage 1) merged and took 223. That is twice in two days that our own merges have moved your number, and it is a property of the append-only scheme, not of anything you did: a number cannot be reserved, it follows merge order.

**What I changed** - the same seven places as last time:

- the `MIGRATIONS` entry in `server/db.js` (now `version: 224`)
- the comment in `server/routes/inventory/item-dates.js:89`
- five `v223` mentions in `docs/SPEC.md`

No schema-mirror entry to follow - this migration has none, and I checked rather than assumed.

`package.json` conflicted again and I took the union: main added `test:migrate-tolerance` and `test:health-nutrition`, your branch keeps `test:inventory-service-log`.

**One thing worth knowing, because it nearly went wrong quietly.** Both sides now had a line reading `version: 223,` at the same spot, so git treated that single line as *common* and marked only the bodies as conflicting. A straightforward resolution - keep both bodies - therefore produces an inventory migration with **no version line at all**, and `node --check` does not catch it: the object literal stays syntactically valid, it is just missing a key. I noticed it because the version list read 220, 221, 222, 223 with the inventory block sitting unlabelled underneath. I redid it by taking main's file whole and appending your block complete, matched by brace balance.

**Proof, rather than a claim:**

- `test:migrations-append-only`, `test:schema-mirror`, `test:schema-reconcile`, `test:suite-chain`: all exit 0.
- Counter-proof, because a guard that has never been red proves nothing: I set the entry to 225 to open a gap - `test:migrations-append-only` fails with "die Migrationen sind lueckenlos aufsteigend nummeriert", exit 1. Reverted by file copy, exit 0 again.
- Full `npm test` on the merged state before pushing: exit 0 read from a FILE rather than from the pipe, 0 failing lines, 356 suites. The merge is `510cf94a2`.

**Please pull before you carry on** - your local branch and the remote have diverged by this merge commit.

The three findings from my review this morning are still with you and unchanged: the UTC day in the timeline at `service-log.js:392`, `odometer_unit` never being written (a miles household is shown "km"), and the `docs/SPEC.md` line promising output the code does not render.

ulsklyc:
One more merge from `main`, pushed as `08d7defeb` - short this time, because it is mechanical and nothing of yours moved.

Only `package.json` conflicted: `main` gained two suites in the test chain (`test:hidden-cascade` and `test:calendar-sync-target-hint`), and I took the union with `test:inventory-service-log` kept where it was. **Your migration stays 224** - nothing on `main` took a number this time.

One thing worth knowing: `test:hidden-cascade` is new on `main` and is the guard for exactly the trap I found in your odometer group this morning - an element hidden with `hidden` stays visible when its own class sets `display`. It now covers your branch. It is green here, and when I remove the `.inventory-form-row[hidden]` rule it goes red naming `#inv-odometer-group` at `inventory.js:1667` and `:1745`. So that fix is held by a test now, not just by a rule someone could tidy away.

Full `npm test` on the merged state before pushing: exit 0 read from a file, 0 failing lines, 358 suites.

**Please pull before you carry on.** The three points from my review this morning are unchanged and still yours.

mclgoerg:
Thank you for taking the migration renumbering off my plate the last few rounds - and for the CSS cascade fix and the guard issue, that's clearly a systemic thing worth catching once rather than five times.

All three findings from this morning, fixed:

1. **UTC day in the timeline.** `date: String(doc.created_at).slice(0, 10)` is now `utcToWall(doc.created_at, householdTimeZone(db.get()))?.date` - the same conversion `todayKey()` uses. Added a test with your exact numbers (23:30 UTC / 01:30 Berlin next day) pinning the household day, not the UTC day.
2. **`odometer_unit` never written by the log/completion paths.** Went with your first option: the item's existing unit wins if it has one, `'km'` is the default only when it doesn't (same rule as `items.js#validateItemFields()`). Added tests for both the "already `mi`, stays `mi`" and "no unit yet, defaults to `km`" cases.
3. **The stale "last reading" example in `docs/SPEC.md`.** Trimmed to match what `t('inventory.trackedDateDistanceHint')` actually renders.

Both nice-to-haves also done:
- `completeTrackedDate()`'s rollover now computes the needed number of intervals in one step (`monthsBetween()` + a bounded correction loop for day-clamping) instead of a month-by-month loop. Added a test with a 100-year gap to prove it's no longer O(elapsed months) - tried your `9999-12-31` example first and hit an unrelated pre-existing bug in `parseDateKey` (doesn't accept 5-digit years), so I used a smaller but still-illustrative gap instead and left that one alone as out of scope here.
- `onDoneTrackedDate()` no longer awaits its own recursive `openItemDetail()` call - nothing after it depended on the result, and `refocusAfterRender()`'s own comment already says it doesn't need the wait.

`npm test` is green (334/334 suites). Merged `main` in again (no new migration this time, just a locale/permissions batch); pulled your pushes first so nothing of yours got clobbered.

One thing worth naming, since you raised it yourself: this PR has cost both of us a fair number of rounds purely on migration-number churn as other work landed on `main` while this sat in review. Your approach of resolving that directly at merge time (rather than sending it back) is clearly the right shape for it - happy to keep leaning on that going forward rather than me pre-emptively rebasing every time a number shifts. Appreciate the very careful work on this one either way.

ulsklyc:
Thanks for the quick round. The timeline day, the odometer_unit default and the SPEC line all check out: I reverted each fix locally and its test went red, then green again with the fix back.

I pushed two small follow-ups to your branch (plus a merge of current main, where only the `package.json` test chain conflicted; migration 224 is unchanged):

**The closed-form rollover skipped an interval in some cases.** `ceil((elapsed + 1) / interval)` is one step too many when the elapsed months are an exact multiple of the interval and the completion day falls before the due day. Example: a yearly service due 2025-06-10, done on 2026-06-05, rolled to 2027-06-10 instead of 2026-06-10. `rollForwardPast()` now starts at `floor(elapsed / interval)`, which never lands past the completion month, so the correction loop runs at most once. The 100-year test was pinning the off-by-one: it expected 2100-02-15, but the first monthly mark after 2100-01-01 is 2100-01-15. Its `< 1000 ms` check couldn't tell a closed form from the old month-by-month loop (that ran in about 4 ms), so I dropped it and added the yearly case through the route. That one fails with the `ceil` version.

**Deleting the only reading cleared the unit.** `recomputeItemOdometer()` set `odometer_unit` to NULL when no reading was left, so an item with 'mi' from the form lost it, and the next reading came in as 'km'. It now keeps the existing unit in that case. There's a route test for it (mi, log a reading, delete it, unit stays mi).

On `parseDateKey` and five-digit years: I'd leave `parseDateKey` as it is. The only way to get one is to roll a date in year 9999 past 9999-12-31. `dateKey()` then writes `10000-...`, which the `date()` validator never accepts and which sorts before `9999-...` as a string, so accepting it would break key comparisons. Today that case ends in a 500 (the old loop did the same). A 400 in `completeTrackedDate()` would be the tidy answer, but it isn't needed for this PR.

The rest was clean, thank you for the care on this one. Merging now; it goes out with the next release.

Issue #1251, opened 2026-09-17
Overview: a task row does not show whether it has been started


whilenotsatisfied:
### Yuvomi version

2.66

### Deployment

Docker Compose (incl. web/CLI installer)

### Description

To complete a task from the overview page, you must first start the task. Then click it again to complete the task, there is also no indication if a task has been started or not in the overview page.

This is on the mobile view, not sure about the desktop.

### Steps to reproduce

1. Create a task
2. Go to the overview page
3. Click on the task
4. Try to complete the task

### Expected vs. actual behavior

Click the task on the overview, let there be a 'finnished task' button along with the start task button

### Logs

```shell

```

ulsklyc:
Thanks for writing this down - you found a real one, and your diagnosis points at exactly the right place.

The overview does not show a task's status at all. Its rows carry priority, title, due date and the people assigned, and nothing else. Tapping a row opens the task's reading view rather than taking you to the tasks page, and that view had a single status button which moved the task one stage along: an open task could only go to "in progress", and only from there to done.

So completing something from the overview took two rounds - start it, open it again, complete it - and because the row does not carry the status, nothing changed on screen in between. The first tap looked like it had been swallowed. On a phone that route is also the only one, since the list card hides its inline controls on narrow screens.

Fixed in #1273: an open task now offers **Complete** and **Start** side by side, with Complete first. Starting a task is unchanged and still there - "in progress" says something real about a task, it just was never meant to stand in front of completing it.

On the second half of your report - that nothing indicates whether a task has been started - I would rather think that through than bolt it on. The overview row is deliberately sparse and already tight on a phone, so a fourth marker in it is a design decision rather than a bug fix. It stays on the list.

One thing your report led to indirectly, which is worth mentioning because it is the more dangerous half: with two buttons side by side, a double tap on a slow connection could have completed a task and then immediately undone it - taking the awarded points, the recorded completion and, on a repeating task, the next occurrence with it. Either button now locks the other while the server is answering.

ulsklyc:
Narrowing this issue to what is actually left of it.

The first half has shipped in #1273 and is on `main`: an open task now offers **Complete** and **Start** side by side, so finishing something from the overview no longer takes two rounds through a state you did not want.

[...]

PR #1273, opened 2026-09-18
fix(tasks): an open task can be completed in one step, without starting it first

ulsklyc:
Closes nothing on its own - see the last paragraph.

## What was wrong

Reported in #1251 (whilenotsatisfied, mobile): on the overview a task has to be *started* before it can be completed, and nothing shows whether it is running.

The path, at the code:

1. The overview row shows no status. `renderUrgentTasks()` (`public/pages/dashboard.js:990-997`) draws priority, title, due date and the assigned people, and nothing else.
2. Tapping the row opens the task's reading view rather than navigating (`dashboard.js:4189`, `openTaskFromOverview`).
3. That view held a chain. `NEXT_STATUS` (`public/components/task-detail.js:154`) mapped each status to exactly one next step, so `open` led only to `in_progress`, and `done` was reachable only from there.

So completing a task from the overview took two rounds, and because the row does not carry the status, nothing changed on screen in between - the first tap looked swallowed. On a phone this route is also the only one, since the list card hides its inline controls below 640px. The view contradicted its own comment two lines further down: "Der häufigste Grund, eine Aufgabe zu öffnen, ist sie abzuhaken."

## What changed

`NEXT_STATUS` (one status -> one step) became `STATUS_ACTIONS` (one status -> a list of steps). An open task now carries **Complete** and **Start** side by side, Complete first. The intermediate state stays - `in_progress` says something real about a task, it was just never meant to stand in front of completing it.

No new locale keys: `tasks.detailStart`, `tasks.detailFinish` and `tasks.detailReopen` already exist in all 24 locales. No migration.

## Proof

New suite `test:task-detail-finish`. It drives the real `openTaskDetail()` and reads the actions it hands over, instead of reading the source - a text guard cannot see this fault, because the buttons existed, they were just never available at the same time. The loader stub for `/components/detail-view.js` now passes its options to `globalThis.__openDetailView`, the same shape as `__apiStub`.

Against `b592a43fb` **four of the five cases are red**. The fifth - an archived task offers no advancement at all - stays green and is the regression guard. Full chain green.

`test:detail-view` held the old chain as text and went red after the change. It now points at `STATUS_ACTIONS` and defers the behaviour to the new suite, but it is **not narrowed**: the assurance that archived tasks never appear there, and the whole block about `advanceTaskStatus` (route, rollback, error toast), are unchanged. 53 cases green.

## Found in review, and it is the more dangerous half

`btnLoading()` disables only the button it is handed, and while each status had exactly one button that was a complete in-flight guard. With Complete and Start side by side - each with its own listener, the panel open until the response lands - both can be fired on a slow connection. The first call awards points, records the completion and, on a repeating task, spawns the next occurrence; the second reads `prev` fresh, finds `done`, and because `prev.status === 'done' && status !== 'done'` holds it revokes the credit, discards the completion and calls `discardRecurrenceFollowup`, deleting the occurrence the first call just created.

Every status button now disables its siblings for the duration of the call and releases them if it fails. Disabled only, not put into the loading state - the spinner belongs on the button somebody pressed.

Measured with a hanging response, since the state *during* the write is the whole point, and asserting only on the sibling: the pressed button is disabled by `btnLoading()` from the stubbed `modal.js`, so an assertion about it would measure the stub. The failure path is covered as well, because a lock left hanging makes the view unusable without anything having been written. Locking and releasing, stubbed out one at a time rather than deleted, each turn exactly their own case red.

One stub fix came with it: the loader's `btnLoading` returned `undefined` where the real one returns a function. Callers hold it as `stop()` and call it in their error path, so any test driving that path died on a TypeError instead of on the thing it meant to measure.

## Deliberately not in here

The second half of the report - that nothing indicates whether a task has been started - is a question about the overview row, not about this view. That row is deliberately sparse and already tight on a phone; a fourth marker in it is a design decision rather than a bug fix. #1251 therefore stays open, and this PR says `Refs`, not `Closes`.


ulsklyc:
Confirmed at the code and fixed in `b20bc03c` - this one was right, and it is the more dangerous half of the change.

I checked each step rather than taking the report for it: `btnLoading()` in `public/components/modal.js` does only touch the button passed to it, the detail-view footer wires one listener per action with no shared disable, `advanceTaskStatus` keeps the panel open until the response lands, and `server/routes/tasks.js` has no guard for `done -> in_progress` - the `prev.status === 'done' && status !== 'done'` branch runs `discardRecurrenceFollowup` unconditionally. So the sequence described really does end with the completion revoked and the spawned occurrence deleted.

Every status button now disables its siblings for the duration of the call and releases them if it fails. Disabled only, not put into the loading state - the spinner belongs on the button somebody pressed.

The regression cover is in `test:task-detail-finish`, measured with a hanging response so the state *during* the write is visible at all, and asserting only on the sibling: the pressed button is disabled by `btnLoading()`, which the test loader stubs, so an assertion about it would measure the stub rather than the code. Locking and releasing, stubbed out one at a time rather than deleted, each turn exactly their own case red.

Issue #1267, opened 2026-09-18
[Bug] Server migration via backup restore is impossible


BVV2:
### Yuvomi version

from 2.65.1 (TrueNAS app) to 2.67.0 ProxMox script (LXC)

### Deployment

Docker Compose (incl. web/CLI installer)

### Description

The installation via TrueNAS app allows to set up Session Secret and Database Password on container creation (therefore they are controlled). However, ProxMox script installation just pulls the container and all secrets are autogenerated on the first setup.
That means, if there is a need to migrate to a autogenerated instance of Yuvomi, the secrets will not match the old ones and the backup restore will fail.

Accessing /opt/yuvomi/.env in the LXC instance and changing the secrets will simply cause the instance to fail and not load at all. Replacing /opt/yuvomi/data/yuvomi.db with the backup and changing the secrets seems to have same effect of crashing the app on the startup.
I wrote the exact path I went through in the "Steps to reproduce", because there might be something obvious that I am missing.

### Steps to reproduce

1. Download the old instance's backup file.
2. Set up a new instance on another host.
3. Go to Settings -> Administration -> Backup and restore.
4. Upload the .db file into "Restore database backup".
5. Click "Restore backup" and it will give error "is not a file".
6. Access the container via terminal.
7. Replace /opt/yuvomi/data/yuvomi.db with old instance .db file by any means available.
8. Access /opt/yuvomi/.env and replace SESSION_SECRET and DB_ENCRYPTION_KEY values with old instance values.
9. Restart the container.
10. Container is accessible, but the Yuvomi instance is not anymore.

### Expected vs. actual behavior

The expected behavior would be allowing user to migrate from one instance of Yuvomi to another via GUI by simply downloading and uploading backups. That should work only if user knows the old database password.

There might be a solution which will be more user-friendly: in the Admin Settings "Restore Backup" section, the user might be able to provide an old database password, which will be one-time used to decrypt the provided ".db" file, storing the data in memory, applying it as a restored backup and then wiping everything clean to maintain the security.
So, I guess this issue might count as a feature request as well?

### Browser / device / OS

Chromium Version 152.0.7977.82

### Logs

```shell

```

### Additional context

The ProxMox installation script was taken from here: https://community-scripts.org/scripts/yuvomi

ulsklyc:
Thanks for writing down the exact path you took - it is what made this findable.

[...]

PR #1275, opened 2026-09-18
fix(backup): the restore advice no longer locks the instance out

ulsklyc:
The advice we gave on an undecryptable backup was itself the trap. #1272 made the restore dialog
name the encryption key instead of claiming the file was not a database, and it added a sentence
telling the admin what to do: "set DB_ENCRYPTION_KEY to it and restart Yuvomi, then restore again."

The reporter of #1267 followed that sentence. His instance stopped starting altogether - the log he
posted is our own check at `db.js:192`. An instance that has a key of its own has its own database
encrypted with that key, so swapping the key just moves the mismatch from the backup file onto the
live database. `init()` aborts, nothing listens on the port, and the dialog that gave the advice is
out of reach from that moment on. Reproduced here through the real start-up path before writing a
line of this: same check, same message.

Measured, not assumed - four routes on a real instance:

| what was changed | result |
| --- | --- |
| key only (what the message advised) | locked out |
| database file only | locked out |
| file + key, stale `-wal`/`-shm` left in place | **"Wrong encryption key" - with the right key** |
| file + key, `-wal`/`-shm` removed | starts |

The third row is a second finding of its own. A write-ahead log belonging to the *previous* database
is read on open, so it produces the exact same error as a wrong key while the key is in fact
correct. That is what made this take two days to pin down: the message ended the search at the
wrong place.

## What changed

- `unreadableBackupError()`, `DB_KEY` branch: no longer advises the key swap. It says plainly that
  the swap stops Yuvomi from starting, and points at the command-line route, which replaces the
  file and sets the key together with the app stopped.
- **The `!DB_KEY` branch keeps the old advice, deliberately.** There it is correct: an instance
  without a key of its own holds a plaintext database, and `encryptPlaintextDatabase()` encrypts it
  with exactly that key on the next start, after which the key fits both. Measured before the
  distinction went into the code, not reasoned about - a correction that swept both branches would
  have removed the only route that needs no shell.
- `undecryptableDatabaseError()`, new: the start-up abort now names the way back, and names a
  leftover `-wal`/`-shm` **only when one is actually lying next to the database**. A paragraph about
  a file that is not there sends the reader in a direction with nothing to find - the same kind of
  error in the other direction.
- `settings.backupCliForeignKeyHint`, new key in all 24 locales: the CLI restore section described
  swapping the file and never mentioned the key at all. Pointing the message at instructions that
  do not carry would have been the same dead end with an extra step.

No migration. No new dependency.

## Proof

Five cases added to `test:db-encryption`. Against `8e8091287`, **three of them fail** (counter-proof
via file copy, not `git checkout` - the working tree is shared). The two that stay green are the
point of them: "the `!DB_KEY` branch keeps its advice" and "with no journal present the message
invents none" are regression guards, and a correction that overreaches turns them red.

`test:db-encryption` 26 green. `test:i18n` 54, `test:i18n-translated` 92, `test:i18n-plural` 19,
`test:frontend-audit` 398, `test:changelog` 34 - all green. Full chain green.

Refs #1267 - the design question the issue stays open for (entering the source instance's key in the
restore dialog) is not decided here. After this round it is the more convincing shape, because every
alternative ends at a shell prompt; what is left to settle is where that key may live while it is in
use.


ulsklyc:
Both findings were right, and the first one mattered more than its wording suggests.

**The leftover-WAL paragraph.** Confirmed at the code: `journal_mode = WAL` is persistent, and there is no `SIGTERM`/`SIGINT` handler anywhere in `server/` that closes the database, so an ordinary stop leaves a database its own `-wal`. `existsSync()` therefore says almost nothing about a hand-swapped file - it fires on the routine case too, including the plain typo-in-the-key case. And the paragraph then issued an unconditional imperative to delete files that can hold committed transactions not yet in the main file. That is the same shape of error this whole PR exists to remove: advice that damages the instance of the person following it. Thank you for catching it before it shipped.

Rewritten so the message states what the file's presence does and does not prove, says plainly not to delete it in the ordinary case and what would be lost, makes the hand-swap the explicit condition - which only the admin can answer - and, where it does apply, asks for the files to be moved aside rather than deleted.

You were also right that no test covered it. There is one now: it opens an encrypted database, writes, and never closes it - exactly what `docker stop` does - then boots with a wrong key and asserts the message does not tell the admin to delete that log. Counter-proof: red against the previous commit, along with the existing foreign-WAL case, which now also asserts the conditional wording rather than the mere mention.

**The subject line.** Correct, 83 characters. Shortened to 64, which leaves room for the `(#1275)` a squash appends. The PR title is updated to match.

One thing I did not change: the `!DB_KEY` branch still advises setting the key and restarting. That branch is measured, not assumed - an instance with no key of its own holds a plaintext database, `encryptPlaintextDatabase()` encrypts it with that key on the next start, and the key then fits both the database and the backup. A regression test holds that branch in place, because a correction sweeping both would remove the only route that needs no shell.

